PR #32, opened 2018-11-16
Master and Node Versions Default to Latest

trotttrotttrott:
This is as opposed to letting GKE figure out the latest patch suffix. This strategy does not work for specifying node version - only masters. An explicit patch suffix is required for node version.

Latest, "...specifies the highest supported Kubernetes version currently available on GKE in the cluster's zone or region" [1]. I've tested this in `us-west1`, `us-west2`, and `us-central1` and it's worked as suspected - all masters and nodes were launched with version 1.11.2-gke.18.

An alternate approach would be to use the [`google_container_engine_versions` data source](https://www.terraform.io/docs/providers/google/d/google_container_engine_versions.html). This would be useful if we had a need to target a specific minor or patch version as the default. I would argue that it's more appropriate for the projects that use this module to utilize `google_container_engine_versions` to satisfy project-specific needs.

[1]: https://cloud.google.com/kubernetes-engine/versioning-and-upgrades#specifying_cluster_version

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -57,7 +57,7 @@ variable "subnetwork" {
 
 variable "kubernetes_version" {
   description = "The Kubernetes version of the masters. If set to 'latest' it will pull latest available version in the selected region."
-  default     = "1.10.6"
+  default     = "latest"
 }
 
 variable "node_version" {


morgante:
> An alternate approach would be to use the google_container_engine_versions data source.

I think this is a better option, as it will at least make the Terraform version declarative.

trotttrotttrott:
@morgante my mistake, this data source is already in use: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/main.tf#L31-L32. While a specific version is not declared in the module code, the value of the `latest_node_version` attribute is passed in to the GKE API on launch. Unless we'd like to make the default an older version for some reason, `latest` seems to me to be the way to go here.

I did spend some time thinking about how we could implement logic around an explicit version like `1.10.6`. Theoretically, we could search the `valid_node_versions` attribute, but Terraform does not support filtering lists by regex. Nothing particularly elegant comes to mind. I suspect the solution would add more complexity than value.

PR #21, opened 2018-10-23
Private Cluster Configuration

Jberlinsky:
**Note:** This switches the implementation of this module to require the `google-beta` provider to be configured. See discussion in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/13#issuecomment-428627463

This PR has significant overlap with https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/20, and **should not be merged** (or really even reviewed) until https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/20 is.

Addresses https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/8

PR #51, opened 2019-01-03
Private Cluster Configuration

Jberlinsky:
Supersedes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/21. Includes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/48 to allow testing on Concourse.

### DIFF
--- a/cluster_regional_private.tf
+++ b/cluster_regional_private.tf
@@ -82,6 +82,12 @@ resource "google_container_cluster" "primary_private" {
       service_account = "${lookup(var.node_pools[0], "service_account", var.service_account)}"
     }
   }
+
+  private_cluster_config {
+    enable_private_endpoint = "${var.private_enable_private_endpoint}"
+    enable_private_nodes    = "${var.private_enable_private_nodes}"
+    master_ipv4_cidr_block  = "${var.private_master_ipv4_cidr_block}"
+  }
 }
 
 /******************************************
--- a/cluster_regional_private.tf
+++ b/cluster_regional_private.tf
@@ -27,8 +27,8 @@ resource "google_container_cluster" "primary_private" {
   region           = "${var.region}"
   additional_zones = ["${coalescelist(compact(var.zones), sort(random_shuffle.available_zones.result))}"]
 
-  network            = "${data.google_compute_network.gke_network.self_link}"
-  subnetwork         = "${data.google_compute_subnetwork.gke_subnetwork.self_link}"
+  network            = "${replace(data.google_compute_network.gke_network.self_link, "https://www.googleapis.com/compute/v1/", "")}"
+  subnetwork         = "${replace(data.google_compute_subnetwork.gke_subnetwork.self_link, "https://www.googleapis.com/compute/v1/", "")}"
   min_master_version = "${local.kubernetes_version}"
 
   logging_service    = "${var.logging_service}"
@@ -88,6 +88,8 @@ resource "google_container_cluster" "primary_private" {
     enable_private_nodes    = "${var.private_enable_private_nodes}"
     master_ipv4_cidr_block  = "${var.private_master_ipv4_cidr_block}"
   }
+
+  remove_default_node_pool = "${var.remove_default_node_pool}"
 }
 
 /******************************************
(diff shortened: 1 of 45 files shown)


cluster_regional_private.tf
```diff
@@ -0,0 +1,153 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/******************************************
+  Create regional cluster
+ *****************************************/
+resource "google_container_cluster" "primary_private" {
```
morgante: What is the reason for needing a separate cluster config for private clusters? Ideally we should minimize how many different `google_container_cluster` resources we have?

Jberlinsky: Apologies, a commit was missing from this. https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/51/commits/719ee8ba49e9cbc62b5c8454aff8cc01719b9780 shows the actual configuration of the private cluster.

The issue here is that Terraform 0.11.x doesn't support conditionally adding this block. Hopefully Terraform 0.12 will make this less painful, but for now I'm not sure there's a better option than duplicating the resource.

morgante: As far as I can tell, the `private_cluster_config` block itself doesn't force anything. If you set all its arguments to false, you would have a standard public cluster.

it's the `enable_private_nodes` piece which is what is primarily meant by a private cluster usually.

Jberlinsky: @morgante Unfortunately, the provider requires a `master_ipv4_cidr_block` to be [provided as a valid CIDR](https://github.com/terraform-providers/terraform-provider-google/blob/91a513ab832f35673720f5f5902e1f6f34841169/google/resource_container_cluster.go#L567) whenever the `private_cluster_config` block is specified, and overriding it from the default (unspecified) value can have hard-to-reason-about results. If we simply don't provide that attribute, things work, but we fall into the Terraform 0.11.x trap where an empty string is not considered the same as null, and we have no way to represent null when it should be so.

I would suggest that we merge this as-is, understanding that there is additional complexity as a result of this, and file an upstream issue to make the validation on `master_ipv4_cidr_block` less stringent.

morgante: I think we should hold off as this is a change which will cause breakage later on if we walk it back and instead file an upstream provider issue.

aaron-lane: Rather than driving optional features through variable values, can we provide a submodule which can be optionally included in a configuration if a private cluster is desired?

test/integration/simple_regional_private/controls/gcloud.rb
```diff
@@ -0,0 +1,188 @@
...
+          },
+          "networkPolicyConfig" => {
+            "disabled" => true,
+          },
+        })
+      end
+    end
+
+    describe "default node pool" do
+      let(:default_node_pool) { data['nodePools'].select { |p| p['name'] == "default-pool" }.first }
```
aaron-lane: This appears to be obsolete in this context.

test/integration/simple_zonal_private/controls/gcloud.rb
```diff
@@ -0,0 +1,178 @@
...
+        expect(data['status']).to eq 'RUNNING'
+      end
+
+      it "is zonal" do
+        expect(data['location']).to match(/^(.*)[1-9]-[a-z]$/)
+      end
+
+      it "is private" do
+        expect(data['privateClusterConfig']['enablePrivateEndpoint']).to eq true
+        expect(data['privateClusterConfig']['enablePrivateNodes']).to eq true
```
aaron-lane: Ditto about single assertions.

test/integration/simple_regional_private/controls/gcloud.rb
```diff
@@ -0,0 +1,188 @@
...
+        expect(data['status']).to eq 'RUNNING'
+      end
+
+      it "is regional" do
+        expect(data['location']).to match(/^.*[1-9]$/)
+      end
+
+      it "is private" do
+        expect(data['privateClusterConfig']['enablePrivateEndpoint']).to eq true
+        expect(data['privateClusterConfig']['enablePrivateNodes']).to eq true
```
aaron-lane: We should try to limit tests to single assertions in order to provide clarity when tests fail.

cluster_regional_private.tf
morgante: Let's approach this by making a submodule as @aaron-lane suggested.

morgante:
Closing in favor of #69

PR #47, opened 2018-12-19
Add support for setting node pool metadata key/value pairs.

czka:

PR #52, opened 2019-01-03
Node pool metadata key value pairs

Jberlinsky:
Expands on https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/47 with test coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ module "gke" {
     all = {}
 
     default-node-pool = {
-      shutdown-script = "${file(var.kube_shutdown_script)}"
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -52,6 +52,14 @@ module "gke" {
     }
   }
 
+  node_pools_metadata = {
+    all = {}
+
+    default-node-pool = {
+      shutdown-script = "${file(var.kube_shutdown_script)}"
```
morgante: One small suggestion: let's just have a simple metadata annotation here instead of referencing a file + variable.

```suggestion
      node-pool-metadata-custom-value = "my-node-pool"
```

PR #54, opened 2019-01-03
Set horizontal_pod_autoscaling to true by default

### LINKED ISSUE #42: horizontal_pod_autoscaling should be enabled by default
You have `horizontal_pod_autoscaling` disabled by default. In Terraform (https://www.terraform.io/docs/providers/google/r/container_cluster.html) it's **enabled** by default though. Why have the opposite? It seems to me that generic Terraform modules should not change the default behavior of the original resources they are wrappers for, unless there's a good reason.

mkubaczyk:
I just wanted to do a step further with #42 issue reported recently. It might be considered a cosmetic change, but I deeply believe following the standards from the `google_container_cluster` resource is more than appropriate. 
The choose of `false` comes from 21ec8db89dcf0c27fe80ccf2b28703500e234768 which is the very first commit from this module, hence this might be worth updating.

Fixes #42.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Changed
+* Set `horizontal_pod_autoscaling` to `true` by default. #42
 
 ## [v0.4.0] - 2018-12-19
 ### Added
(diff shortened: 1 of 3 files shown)


morgante:
I think this is a reasonable suggestion and change. One ask: could you add a note to the `CHANGELOG.md` in the unreleased section?

mkubaczyk:
@morgante done. :-)

Issue #15, opened 2018-09-26
Remove default node pools and manage them explicitly


pratikmallya:
The default node pools cause trouble with managing the cluster. It seems like a better architecture to manage the node pools explicitly. Does it make sense to use the `remove_default_node_pool` parameter for the `google_container_cluster` and add just manage the node pools explicitly?

ref: https://github.com/terraform-providers/terraform-provider-google/issues/475#issuecomment-383529139

morgante:
We definitely encourage explicitly managing the node pools, as shown in the examples: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples/node_pool

Regarding the `remove_default_node_pool` parameter, I believe we ran into some trouble with bootstrapping using it. @ryanckoch can share more. The approach we have today is, I believe, to create a default node pool with 0 nodes.

pratikmallya:
Oh, would love to hear about what the issues were with using that parameter @ryanckoch :)

AdrienWalkowiak:
I would love to see this fixed too!

miles-:
I'd also like to help get this fixed - what was the issue with the `remove_default_node_pool` parameter? @ryanckoch

morgante:
@ryanckoch Can you chime in here?

aaron-lane:
Fixed by #55.

PR #55, opened 2019-01-03
Add remove_default_node_pool

mkubaczyk:
In the #15 issue there was a request for remove_default_node_pool. I'm interested in introducing this to the module. There's no point in keeping the default node pool the `google_container_cluster` forces us to have on the beginning. Usually, the one wants to manage the pools explicitly, and so do I, hence this PR.

Adding the `remove_default_node_pool` was not enough to make it work perfectly. A change in the manner of providing the `google_container_cluster` with the network and subnetwork parameters had to be changed as well. Without it, terraform produced permadiff with every `plan` and `apply`.  That's the issue with google provider, more on that is described [in this google provider issue](https://github.com/terraform-providers/terraform-provider-google/issues/1566). 

The approach I propose is a workaround which won't cause any backward incompatibilities. The issue I faced might be the problem described from the #15.

If anyone is having the better solution, feel free to propose it!

### DIFF
--- a/cluster_zonal.tf
+++ b/cluster_zonal.tf
@@ -26,8 +26,8 @@ resource "google_container_cluster" "zonal_primary" {
   zone             = "${var.zones[0]}"
   additional_zones = ["${slice(var.zones,1,length(var.zones))}"]
 
-  network            = "projects/${var.project_id}/global/networks/${data.google_compute_network.gke_network.name}"
-  subnetwork         = "projects/${var.project_id}/regions/${var.region}/subnetworks/${data.google_compute_subnetwork.gke_subnetwork.name}"
+  network            = "projects/${data.google_compute_network.gke_network.project}/global/networks/${data.google_compute_network.gke_network.name}"
+  subnetwork         = "projects/${data.google_compute_network.gke_network.project}/regions/${var.region}/subnetworks/${data.google_compute_subnetwork.gke_subnetwork.name}"
   min_master_version = "${local.kubernetes_version}"
 
   logging_service    = "${var.logging_service}"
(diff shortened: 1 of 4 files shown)


cluster_zonal.tf
```diff
@@ -26,8 +26,8 @@ resource "google_container_cluster" "zonal_primary" {
   zone             = "${var.zones[0]}"
   additional_zones = ["${slice(var.zones,1,length(var.zones))}"]
 
-  network            = "${data.google_compute_network.gke_network.self_link}"
-  subnetwork         = "${data.google_compute_subnetwork.gke_subnetwork.self_link}"
+  network            = "projects/${var.project_id}/global/networks/${data.google_compute_network.gke_network.name}"
```
morgante: Isn't that equivalent to the self_link?

The issue with formatting it like this is if the cluster is being launched on a shared VPC, in which case the network is in a *different* project.

mkubaczyk: _self_link_ adds `https://www.googleapis.com/compute/v1` before the `projects/[..]`, hence the permadiff since terraform sees the current path as the one starting from `projects/[..]` and tries to update it.

I'll take a look at the shared VPC case and will come with a solution/an idea hopefully. :-)

mkubaczyk: @morgante I reused the 'data.google_compute_network` to get the project_id, which takes the shared VPC into account right now. Proposed in 8f96807c4a211ed29d2c6d60b343d87ae1c4cfba

morgante: One alternative idea: what if we used the self_link but used `replace` to strip the URL prefix? https://www.terraform.io/docs/configuration/interpolation.html#replace-string-search-replace-

mkubaczyk: Sounds good.
As simple as `network = "${replace(data.google_compute_network.gke_network.self_link, "https://www.googleapis.com/compute/v1/", "")}"` ?

morgante:
Yes I think that should work.

cluster_zonal.tf
mkubaczyk: Added in e997e6604fce1e9d9e77f89c483792f26cc934de. :-)

PR #60, opened 2019-01-09
Add initial_node_count var for node_pool block

mkubaczyk:
Hi,

I wanted to extend the node_pool block parameters a bit and add optional, explicitly set initial_node_count for node pool. It assumes the min_count or 1 when not defined, so it doesn't affect anyone when not set. 
We need this to start new node pool with predefined node count even when autoscaling minimal value is set lower, useful especially when adding new node pool as a reaction to increased traffic etc.

I added no tests. If there's anyone willing to write them since I haven't had a chance yet, please don't hesitate.


Jberlinsky:
Test failures are unrelated, and should be resolved by https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/62

mkubaczyk:
@aaron-lane done

PR #61, opened 2019-01-09
Set up submodule structure

Jberlinsky:
Per conversation in [https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/51#pullrequestreview-190300600](PR #51):

- Moves root Terraform configuration to `./autogen`
- Creates submodule structure in `./modules` and creates a `public-cluster` module


test/helpers/generate_modules/test_generate_modules.py
```diff
@@ -0,0 +1,89 @@
...
+        for output in self.outputs:
+            name = 'output "{output}"'.format(output=output)
+            self.assertIn(name, self.root_outputs_tf)
+
+            value = 'value = "${{module.kubernetes-engine.{0}}}"'.format(
+                output
+            )
+            self.assertIn(value, self.root_outputs_tf)
+
+    def test_has_descriptions(self):
```
aaron-lane: This test seems overly complex just to test for the existence of the description of `'output2'`. In general, we should try to avoid dynamically generated tests as they are harder to reason about.

helpers/generate_modules/generate_modules.py
```diff
@@ -0,0 +1,157 @@
...
+                print("Skipping" + dest_path)
+
+
+def create_submodules():
+    # Read `modules/*` for directories
+    # For every directory, copy all files in `autogen/` there.
+    modules = next(os.walk("./modules"))[1]
+    for module in modules:
+        dest_folder = os.path.join("./modules", module)
+        _copy_recursively_with_symlinks("./autogen", dest_folder)
```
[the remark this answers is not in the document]
Jberlinsky: No -- the idea is that `./autogen` will contain files that are common across all submodules, and the submodule directory itself will contain overrides. So, assuming no further changes to how things are laid out in the repository, there might exist a `./modules/private-cluster/cluster_{regional,zonal}.tf`, which `os.symlink` will not override.

autogen/main.tf
```diff
@@ -0,0 +1,151 @@
+/**
```
morgante: What file does this generate?

main.tf
```diff
@@ -14,138 +14,41 @@
...
-/******************************************
-  Get available container engine versions
- *****************************************/
-data "google_container_engine_versions" "region" {
-  zone    = "${data.google_compute_zones.available.names[0]}"
-  project = "${var.project_id}"
+// File autogenerated by `helpers/generate_modules/generate_module.py`
+
+module "kubernetes-engine" {
+  source = "modules/public-cluster"
```
morgante: We should *not* be refactoring main.tf as that will cause a state migration?

The autogen should fill in the root main.tf appropriately. The root module can be the default - public cluster.

outputs.tf
```diff
@@ -14,94 +14,94 @@
...
-output "name" {
-  description = "Cluster name"
-  value       = "${local.cluster_name}"
-}
+// File autogenerated by `helpers/generate_modules/generate_module.py`
 
-output "type" {
-  description = "Cluster type (regional / zonal)"
-  value       = "${local.cluster_type}"
+output "ca_certificate" {
```
morgante: Why is there a big diff in this file? Shouldn't be changed at all. I'm concerned by the number of unrelated files which are being changed.

test/helpers/generate_modules/test_generate_modules.py
morgante: +1, let's avoid dynamically generated tests.

helpers/generate_modules/generate_modules.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    with open("./test/boilerplate/boilerplate.tf.txt") as fh:
+        buf = fh.read()
+        buf += "\n\n"
+        buf += "// File autogenerated by " + \
+            "`helpers/generate_modules/generate_module.py`\n"
+        return buf
+
+
+def root_main_tf(variables):
```
morgante: I don't think we need this functionality.

Jberlinsky:
@morgante To be clear, the first iteration of this will be template generation from `./autogen` into `/` (relative to the repository root), to avoid changing `main.tf` and triggering a state migration. Future PRs will generate additional module(s) into `./modules/`. Is that your understanding?

morgante:
@Jberlinsky Yes, the addition of the private module can happen in a later PR.

PR #74, opened 2019-01-29
Drop explicit version from simple_zonal example

adrienthebo:
The explicit version in the simple_zonal example is causing failures as
1.11.4-gke.13 has been dropped as a valid master version. This commit
drops the explicit versions in the simple_zonal example; we can just use
the latest version avilable.


aaron-lane:
We should think about a way to test the behaviour of the version variables without the tests becoming brittle.

PR #77, opened 2019-01-31
Node pool create: add an implicit dependency on parent cluster

JordanP:
The explicit `depends_on` is not required as there are other ways
to signal to Terraform that a resource depends on another. Implicit
dependencies are more like TF usually operates.


adrienthebo:
Note for posterity: The integration CI failure appears spurious:

```
       1 error(s) occurred:
       
       * module.example.module.gke.local.cluster_master_auth_list_layer2: local.cluster_master_auth_list_layer2: At column 40, line 1: list "local.cluster_master_auth_list_layer1" does not have any elements so cannot determine type. in:
       
       ${local.cluster_master_auth_list_layer1[0]}
```

PR #78, opened 2019-02-08
Allow oauth_scopes to be overridden

awolde:
oauth_scope in the current master version is always tied to only `cloud-platform`. But a module using this might want to override this. I have a typical case where I want to give pods access to decrypt ciphers from kms.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -211,3 +211,11 @@ variable "service_account" {
   description = "The service account to default running nodes as if not overridden in `node_pools`. Defaults to the compute engine default service account"
   default     = ""
 }
+
+variable "oauth_scopes" {
+  type = "list"
+  description = "The oauth scope the GKE compute nodes are allowed to have"
+  default = [
+      "https://www.googleapis.com/auth/cloud-platform",
+    ]
+}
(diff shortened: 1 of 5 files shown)


variables.tf
```diff
@@ -210,4 +210,12 @@ variable "monitoring_service" {
 variable "service_account" {
   description = "The service account to default running nodes as if not overridden in `node_pools`. Defaults to the compute engine default service account"
   default     = ""
-}
\ No newline at end of file
+}
+
+variable "oauth_scopes" {
+  type = "list"
+  description = "The oauth scope the GKE compute nodes are allowed to have"
```
JordanP: Maybe stick to the same wording here: https://www.terraform.io/docs/providers/google/r/container_cluster.html#oauth_scopes

At least s/scope/scopes/

morgante:
```suggestion
  description = "The oauth scopes the GKE compute nodes are allowed to have"
```

kbroughton:
Hi @awolde,

Just wondering what your use-case is for adding custom scopes.
I know the CIS GKE Benchmark recommends it, but that rec may be stale, at least 
according to the "Best Practices" section at the bottom of the doc below.
https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances

The recommend being liberal with scopes and using only IAM roles to manage permissions.
I'm curious if there is a good use case for scopes though.

Kesten

awolde:
@kbroughton I agree. My use case was solved by IAM rules like you said. I guess we can close this.

PR #85, opened 2019-02-28
Ignore autogen directory when generating terraform documentation

adrienthebo:
This pull request updates the `test/make.sh` script for the latest set of functionality, and ignores the `autogen` directory when generating docs as the docs parser cannot handle the templated code.

For ease of review this pull request does not update the generated READMEs; we can perform that in a followup pull request when this work has been merged.


test/make.sh
```diff
@@ -14,87 +14,146 @@
...
+      helpers/combine_docfiles.py "${path}"/README.md "${tmpfile}"
+    else
+      echo "Skipping ${path} because README.md does not exist."
+    fi
+  done < <(find_files . -name '*.tf' -print0 \
+    | compat_xargs -0 -n1 dirname \
+    | sort -u)
+}
+
+function prepare_test_variables() {
```
Jberlinsky: Hm -- I thought this function was removed a while ago...it's certainly vestigial if it's still around.

test/make.sh
```diff
@@ -14,87 +14,146 @@
...
 
 # This function makes sure that there is no trailing whitespace
 # in any files in the project.
 # There are some exclusions
 function check_trailing_whitespace() {
-  echo "The following lines have trailing whitespace"
-  grep -r '[[:blank:]]$' --exclude-dir=".terraform" --exclude-dir=".kitchen" --exclude="*.png" --exclude="*.pyc" --exclude-dir=".git" .
+  local rc
+  echo "Checking for trailing whitespace"
+  find_files . -print \
```
Jberlinsky: Seems to me that it might make sense for this to scan `autogen` as well -- if a file created by template generation has trailing whitespace and its source template does as well, shouldn't we be flagging the source template?

Jberlinsky: Although I also can see the argument for maintaining consistency and only flagging generated files. Happy to discuss this -- I don't have a particularly strong preference. Perhaps an appropriate middle ground might be checking for a source template and alerting that it might be the underlying source of the issue? If we do that here, though, we might want to also do it in other checks.

thefirstofthe300:
This is fixed by #110.

PR #88, opened 2019-03-05
Change service account default to create a new SA per cluster

Jberlinsky:
Per https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/80


adrienthebo:
Upon closer review I wasn't able to find integration tests verifying the presence of the created service accounts, is this a worthwhile addition?

morgante:
@Jberlinsky This looks fine to me, but `make generate` is showing a diff.

PR #87, opened 2019-03-04
Move CI runtime logic from pipeline to ci_integration.sh

adrienthebo:

PR #90, opened 2019-03-05
Lock google, google-beta provider version to ~> 1.20      

adrienthebo:
The google and google-beta providers are causing some issues with the simple-regional-private example; these issues in turn are causing CI to fail for all recent pull requests. This commit repairs CI by pinning the version of the google and google-beta providers; we can address v2.0 support in the future.


adrienthebo:
CI errored out due to the current overhaul implemented by #87 , I'm bypassing the CI checks to get this landed.

PR #91, opened 2019-03-06
Fix issues with `substr` for Service Accounts

lawliet89:
Fixes the following error:

```
* module.gke.google_service_account.cluster_service_account: substr: 'offset + length' cannot be larger than the length of the string in:

tf-gke-${substr(var.name, 0, 20)}
```

### DIFF
--- a/autogen/sa.tf
+++ b/autogen/sa.tf
@@ -24,7 +24,7 @@ locals {
 resource "google_service_account" "cluster_service_account" {
   count        = "${var.service_account == "create" ? 1 : 0}"
   project      = "${var.project_id}"
-  account_id   = "tf-gke-${substr(var.name, 0, 20)}"
+  account_id   = "tf-gke-${substr(var.name, 0, min(20, length(var.name)))}"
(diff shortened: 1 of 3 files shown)


adrienthebo:
We'll need to merge #87 before CI will pass on this PR; hopefully we can wrap that up by EOD.

adrienthebo:
Verified with the following diff:

```
diff --git i/examples/simple_zonal/main.tf w/examples/simple_zonal/main.tf
index a0dc9ff..c95038a 100644
--- i/examples/simple_zonal/main.tf
+++ w/examples/simple_zonal/main.tf
@@ -27,7 +27,7 @@ provider "google" {
 module "gke" {
   source            = "../../"
   project_id        = "${var.project_id}"
-  name              = "${local.cluster_type}-cluster${var.cluster_name_suffix}"
+  name              = "sz"
   regional          = false
   region            = "${var.region}"
   zones             = "${var.zones}"
```

I'll submit a followup PR to add this code path to our test coverage.

PR #99, opened 2019-03-11
Ensure node-pool example node count < 6

adrienthebo:
When GKE stands up a cluster with more than 6 nodes in the node pool, it will automatically resize the master node. This causes the cluster to enter reconciling, which fouls up our CI configuration.

This commit sidesteps this problem by converting the node-pool example to a zonal pool, removing the default node pool and shrinking the max size of the added node pools. We can add graceful handling of cluster reconciliation in a followup PR.

This pull request depends on #98.

### DIFF
--- a/examples/node_pool/main.tf
+++ b/examples/node_pool/main.tf
@@ -25,19 +25,23 @@ provider "google" {
 }
 
 module "gke" {
-  source            = "../../"
-  project_id        = "${var.project_id}"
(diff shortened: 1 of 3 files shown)


aaron-lane:
@adrienthebo #98 was merged. Please rebase and ready for review.

adrienthebo:
We had a spurious failure on build 187, I've triggered another build to re-test this PR.

PR #103, opened 2019-03-14
Add a `check_generate_docs` target

adrienthebo:
Currently we have no automatic way of determining if a given commit or
pull request has stale documentation; this leads to buildups in changes
that tend to tag along with other fixes.

This commit resolves the issue by adding a check that verifies that a
fresh copy of the repository does not have a diff present after docs
generation.

### DIFF
--- a/test/make.sh
+++ b/test/make.sh
@@ -98,3 +98,27 @@ function generate_docs() {
   done
   rm -f "$TMPFILE"
 }
+
+function check_generate() {
+  make generate
+  git diff --exit-codes
+}
+
+function check_generate_docs() {
+  TMPDIR=$(mktemp -d)
+  git worktree add --detach "$TMPDIR"
+  cd "$TMPDIR" || exit 1
+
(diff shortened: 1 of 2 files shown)


morgante:
I'm not sure this is the best way to do this. What's wrong with just running `generate_docs` and checking if there's a diff?

adrienthebo:
Using a clean worktree avoids issues with crossed results with the `make generate` and `make generate_docs`  tasks. If `make generate` generates a diff and then fails but documentation is current, `make generate_docs` will incorrectly fail because there's a diff present. Is there a cost to using clean worktrees that I'm not seeing?

test/make.sh
```diff
@@ -98,3 +98,27 @@ function generate_docs() {
   done
   rm -f "$TMPFILE"
 }
+
+function check_generate() {
```
adrienthebo: This should be handled in a different PR.

PR #108, opened 2019-03-19
Fix issue with regional cluster roll outs causing version skews

thefirstofthe300:
Regional clusters are created using the newest version of GKE that is
available across all zones in which the masters live. When a GKE version
roll out occurs, the available versions across the zones can become
skewed for zonal clusters with version x.y.z-gke.a being the only
available zonal version in one zone but version x.y[+1].z[+1]-gke.a[+1]
being the only zonal version available in another zone.

The Terraform module only checks for the version available in the first
zone returned by a call to the google_compute_zones data resource.

Consequently, the module will fail to create a regional cluster during a
roll out due to the version available in that zone not being available
across all the zones for regional clusters.


autogen/main.tf
```diff
@@ -31,10 +31,12 @@ resource "random_shuffle" "available_zones" {
 }
 
 locals {
-  kubernetes_version     = "${var.kubernetes_version != "latest" ? var.kubernetes_version : data.google_container_engine_versions.region.latest_node_version}"
-  node_version           = "${var.node_version != "" ? var.node_version : local.kubernetes_version}"
-  custom_kube_dns_config = "${length(keys(var.stub_domains)) > 0 ? true : false}"
-  network_project_id     = "${var.network_project_id != "" ? var.network_project_id : var.project_id}"
+  kubernetes_version_regional = "${var.kubernetes_version != "latest" ? var.kubernetes_version : data.google_container_engine_versions.region.latest_master_version}"
+  kubernetes_version_zonal    = "${var.kubernetes_version != "latest" ? var.kubernetes_version : data.google_container_engine_versions.zone.latest_master_version}"
+  node_version_regional       = "${var.node_version != "" && var.regional == "true" ? var.node_version : local.kubernetes_version_regional}"
```
adrienthebo: Avoid string comparisons against boolean values; this will behave badly if a user uses bare `true` or `false`. Let's use `var.regional` and `!var.regional` so this evaluates in a boolean context.

test/make.sh
```diff
@@ -18,14 +18,12 @@
...
 function check_bash() {
-find . -name "*.sh" | while IFS= read -d '' -r file;
-do
-  if [[ "$file" != *"bash -e"* ]];
-  then
-    echo "$file is missing shebang with -e";
-    exit 1;
-  fi;
-done;
+  find . -name "*.sh" | while IFS= read -d '' -r file; do
```
adrienthebo: Is this related to the change in this PR?

modules/private-cluster/variables.tf
```diff
@@ -116,7 +116,7 @@ variable "ip_range_pods" {
 }
 
 variable "ip_range_services" {
-  description = "The _name_ of the secondary subnet ip range to use for services"
+  description = "The _name_ of the secondary subnet range to use for services"
```
adrienthebo: Was this stale that needed to be regenerated?

autogen/main.tf
thefirstofthe300: Done.

test/make.sh
thefirstofthe300: No. I think this is my `shfmt` extension deciding to clean the file for me.

modules/private-cluster/variables.tf
thefirstofthe300: This was regenerated for this MR, yes.

autogen/main.tf
```diff
@@ -149,7 +151,16 @@ locals {
...
+  provider = "google-beta"
+  region   = "${var.region}"
+  project  = "${var.project_id}"
+}
+
+data "google_container_engine_versions" "zone" {
+  // Work around to prevent a lack of zone declaration from causing regional cluster creation from erroring out due to error
+  //
+  //     data.google_container_engine_versions.zone: Cannot determine zone: set in this resource, or set provider-level zone.
+  //
```
aaron-lane: This should include the provider argument like `data.google_container_engine_versions.region`.

autogen/main.tf
```diff
@@ -149,7 +151,16 @@ locals {
   Get available container engine versions
  *****************************************/
 data "google_container_engine_versions" "region" {
-  provider = "{% if private_cluster %}google-beta{%else %}google{% endif %}"
-  zone     = "${data.google_compute_zones.available.names[0]}"
+  provider = "google-beta"
```
aaron-lane: What is the motivation for dropping the conditional logic?

test/make.sh
```diff
@@ -92,8 +90,8 @@ function generate_docs() {
   echo "Generating markdown docs with terraform-docs"
   TMPFILE=$(mktemp)
   #shellcheck disable=2006,2086
-  for j in `for i in $(find . -type f | grep \.tf$) ; do dirname $i ; done | sort -u` ; do
-    terraform-docs markdown "$j" > "$TMPFILE"
+  for j in $(for i in $(find . -type f -path autogen -prune | grep \.tf$); do dirname $i; done | sort -u); do
```
aaron-lane: This change appears to be unrelated to the branch topic.

autogen/main.tf
thefirstofthe300: I can understand for stylistic reasons, but there is no technical reason we should need to explicitly set the google provider. `data.google_container_engine_versions.region` requires using the `google-beta` provider in all cases because specifying a region instead of a zone is only available in the beta provider.

autogen/main.tf
thefirstofthe300: Same reason as above. In 100% of cases now, we are getting the data for a region which requires the google-beta resource.

test/make.sh
thefirstofthe300: Fixed.

test/make.sh
thefirstofthe300: Reverted.

examples/deploy_service/main.tf
```diff
@@ -24,6 +24,11 @@ provider "google" {
   version     = "~> 1.20"
 }
 
+provider "google-beta" {
+  credentials = "${file(var.credentials_path)}"
```
adrienthebo: Is it appropriate to pin the provider version here?

examples/node_pool/main.tf
```diff
@@ -24,6 +24,11 @@ provider "google" {
   region      = "${var.region}"
 }
 
+provider "google-beta" {
+  credentials = "${file(var.credentials_path)}"
```
adrienthebo: Is it appropriate to pin the provider version here?

examples/simple_regional/main.tf
```diff
@@ -23,6 +23,11 @@ provider "google" {
   region      = "${var.region}"
 }
 
+provider "google-beta" {
+  credentials = "${file(var.credentials_path)}"
```
adrienthebo: Is it appropriate to pin the provider version here?

examples/simple_zonal/main.tf
```diff
@@ -24,6 +24,11 @@ provider "google" {
   region      = "${var.region}"
 }
 
+provider "google-beta" {
+  credentials = "${file(var.credentials_path)}"
+  region      = "${var.region}"
```
adrienthebo: Is it appropriate to pin the provider version here?

PR #110, opened 2019-03-21
[skip ci] Fix documentation generation and update docs

thefirstofthe300:
The generate_docs function would attempt to run terraform_docs against
the autogen/ directory. Since that directory contains templates instead
of actual working Terraform, the terraform-docs command failed, causing
no docs to be properly formatted.

These changes update the function to properly ignore the ./autogen
directory.

Also, updates the documentation.

### DIFF
--- a/autogen/README.md
+++ b/autogen/README.md
@@ -95,70 +95,7 @@ Then perform the following commands on the root folder:
 - `terraform apply` to apply the infrastructure build
 - `terraform destroy` to destroy the built infrastructure
 
-
 [^]: (autogen_docs_start)
-
-
-## Inputs
-
-| Name | Description | Type | Default | Required |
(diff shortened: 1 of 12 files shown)


autogen/README.md
```diff
@@ -95,70 +95,8 @@ Then perform the following commands on the root folder:
 - `terraform apply` to apply the infrastructure build
 - `terraform destroy` to destroy the built infrastructure
 
-
+<!-- Since the inputs and outputs will not be autogenerated in this directory, there is no reason to store them here. -->
```
aaron-lane: I think removing the inputs and outputs from this file is sufficient to make this point. This comment will be copied to the generated READMEs which is confusing when viewing the source. 🙃

thefirstofthe300: Done.

PR #117, opened 2019-03-25
[ci skip] Add upgrade guide and update documentation

thefirstofthe300:


main.tf
```diff
@@ -157,6 +157,8 @@ data "google_container_engine_versions" "region" {
 }
 
 data "google_container_engine_versions" "zone" {
+  provider = "google-beta"
```
aaron-lane: This change is not relevant to the branch topic.

modules/private-cluster/main.tf
```diff
@@ -157,6 +157,8 @@ data "google_container_engine_versions" "region" {
 }
 
 data "google_container_engine_versions" "zone" {
+  provider = "google-beta"
```
aaron-lane: This change is not relevant to the branch topic.

modules/private-cluster/variables.tf
```diff
@@ -126,7 +126,7 @@ variable "remove_default_node_pool" {
 
 variable "disable_legacy_metadata_endpoints" {
   description = "Disable the /0.1/ and /v1beta1/ metadata server endpoints on the node. Changing this value will cause all node pools to be recreated."
-  default     = true
+  default     = "true"
```
aaron-lane: This change is not relevant to the branch topic.

variables.tf
```diff
@@ -126,7 +126,7 @@ variable "remove_default_node_pool" {
 
 variable "disable_legacy_metadata_endpoints" {
   description = "Disable the /0.1/ and /v1beta1/ metadata server endpoints on the node. Changing this value will cause all node pools to be recreated."
-  default     = true
+  default     = "true"
```
aaron-lane: This change is not relevant to the branch topic.

main.tf
thefirstofthe300: This appears to be the result of a code skew between the autogen code and the post autogen'd code introduced by https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/commit/1d73e9d56ee0026aa09ddc4f8e6b2a2c325c1c93

tl;dr: I screwed up in a commit outside of this branch so regenerating the documentation makes a diff in the generated code unavoidable.

That being said, the diff should be for removing the provider definition from the `autogen/main.tf` file, not adding it here.

This should now be fixed.

modules/private-cluster/main.tf
thefirstofthe300: Same as above.

modules/private-cluster/variables.tf
thefirstofthe300: Same as above.

variables.tf
thefirstofthe300: This is also the result of a skew between the templates and autogen'd code in previous PR: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/commit/6813253b5f314a4282b231ce1489f0d65180e1d9

The diff is unavoidable when generating the documentation.

PR #122, opened 2019-03-27
Update documented version of `terraform-docs` to 0.6.0 to comply with the rest of CFT modules

Jberlinsky:


modules/private-cluster/README.md
```diff
@@ -8,9 +8,6 @@ The resources/services/activations/deletions that this module will create/trigge
 - Activate network policy if `network_policy` is true
 - Add `ip-masq-agent` configmap with provided `non_masquerade_cidrs` if `network_policy` is true
 
-
-**Note**: You must run Terraform from a VM on the same VPC as your cluster, otherwise there will be issues connecting to the GKE master.
```
morgante: Looks like a note is getting lost here.

Jberlinsky: @morgante Looks like someone edited this file without putting the change in `autogen/`. Tracking it down...

Jberlinsky: Looks like 1a7a5d46 modifies this file without making a change in `autogen/`. While there's certainly a conversation to be had about whether the READMEs should be auto-generated, for now, can we get this fixed in a separate PR?

Jberlinsky: For the time being, I'll manually keep this line in place.

PR #125, opened 2019-03-29
Add network_policy_provider_enabled variable

bamaralf:
Add  `network_policy_provider_enabled` variable to enable 'CALICO' network policy provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ In either case, upgrading to module version `v1.0.0` will trigger a recreation o
 | name | The name of the cluster (required) | string | n/a | yes |
 | network | The VPC network to host the cluster in (required) | string | n/a | yes |
 | network\_policy | Enable network policy addon | string | `"false"` | no |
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -118,6 +118,7 @@ In either case, upgrading to module version `v1.0.0` will trigger a recreation o
 | name | The name of the cluster (required) | string | n/a | yes |
 | network | The VPC network to host the cluster in (required) | string | n/a | yes |
 | network\_policy | Enable network policy addon | string | `"false"` | no |
+| network\_policy\_provider\_enabled | Enable network policy provider: 'CALICO'. When set to 'true' it requires 'network_policy=true' to work properly. | bool | false | no |
```
morgante: Does it ever make sense to have `network_policy=true` and `network_policy_provider_enabled=false`? If not I suggest we just use the single variable.

bamaralf: I edited and now if network_policy=true, the 'CALICO' network_policy provider will be enabled.

PR #126, opened 2019-03-29
Add network_policy_provider_enabled variable to enable 'CALICO' network policy provider.

bamaralf:


variables.tf
```diff
@@ -106,6 +106,11 @@ variable "network_policy" {
   default     = false
 }
 
+variable "network_policy_provider_enabled" {
```
morgante: Does it ever make sense to have `network_policy=true` and `network_policy_provider_enabled=false`? If not I suggest we just use the single variable.

bamaralf:
This PR was created with the modifications: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/129

PR #127, opened 2019-03-29
Add function to check code autogeneration

thefirstofthe300:
This function adds a make target to test if code autogeneration has been run properly. This new target will be added as part of the CI runs to ensure that code is always properly synced from the templates.

### DIFF
--- a/test/make.sh
+++ b/test/make.sh
@@ -86,6 +86,11 @@ function check_trailing_whitespace() {
   fi
 }
 
+function generate() {
+  pip install --user -r ./helpers/generate_modules/requirements.txt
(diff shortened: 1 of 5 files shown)


test/make.sh
```diff
@@ -97,12 +102,35 @@ function generate_docs() {
   rm -f "$TMPFILE"
 }
 
+function check_generate() {
+  TMPDIR=$(mktemp -d)
+  git worktree add --detach "$TMPDIR" >/dev/null
+  cd "$TMPDIR" || exit 1
+
+  generate >/dev/null
```
morgante: Will this silently swallow errors?

PR #129, opened 2019-03-29
Add 'Calico' network_policy provider.

bamaralf:
When the network_policy variable is set to true, the network_policy provider 'CALICO' will be enabled.


bamaralf:
This PR applies the modifications mentioned in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/126

morgante:
Thanks, looking good. Two asks:
- Update the files in the `autogen/` directory (and use `make generate` to update the others)
- Run `make check` and fix whitespace issues the linter found.

bamaralf:
Updated the autogen/ directory and used `make generate`
Output from `make check`:
```
OK
Checking file headers
The following lines have trailing whitespace
Code was generated properly
Docs were generated properly
```

autogen/cluster_regional.tf
```diff
@@ -56,6 +56,11 @@ resource "google_container_cluster" "primary" {
     }
   }
 
+  network_policy {
+    provider = "CALICO"
+    enabled  = true
```
morgante: This seems to hard code network policy to `true. Shouldn't we only enable it if `network_policy` variable is true?

bamaralf: Sorry, I'll fix it. This was a typo during the edition.

bamaralf: @morgante I uploaded the fix.

autogen/cluster_regional.tf
```diff
@@ -56,6 +56,11 @@ resource "google_container_cluster" "primary" {
     }
   }
 
+  network_policy {
+    provider = "CALICO"
+    enabled  = "${var.network_policy ? 1 : 0}"
```
morgante:
```suggestion
    enabled  = "${var.network_policy ? true : false}"
```

README.md
```diff
@@ -96,69 +96,6 @@ Version 1.0.0 of this module introduces a breaking change: adding the `disable-l
 In either case, upgrading to module version `v1.0.0` will trigger a recreation of all node pools in the cluster.
 
 [^]: (autogen_docs_start)
-
```
morgante: Looks like the docs got lost somehow. Try `make generate_docs`.

bamaralf: > Looks like the docs got lost somehow. Try make generate_docs.
Done!

cluster_regional.tf
```diff
@@ -56,6 +56,11 @@ resource "google_container_cluster" "primary" {
     }
   }
 
+  network_policy {
+    provider = "CALICO"
+    enabled  = "${var.network_policy ? 1 : 0}"
```
morgante: This should also be changed to true/false (and remember to regen).

```suggestion
    enabled  = "${var.network_policy ? true : false}"
```

bamaralf:
@morgante Please review the changes. The tests with `make check` are not returning errors, but the PR didn't pass on the integration-tests.

bamaralf:
Hi @Jberlinsky. That's odd. The invocation of the module doesn't change in this PR. (But maybe we need to update the integration test because of an extra config block necessary). Could you please update the integration test?
Regarding the hardcoded `CALICO`: the reason is that this is the only value supported as documented here: https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy on the `Creating a network policy` section.

bamaralf:
@Jberlinsky I think if I add a `depends_on` to the "node_pool" resource can solve the error on the integration test. 
`google_container_cluster.primary: googleapi: Error 400: The network policy addon must be enabled before updating the nodes., badRequest`

Jberlinsky:
@bamaralf Understood regarding `CALICO` being the only value currently supported. However, I think it might make more sense for that value to be the default value of a variable, with the currently supported values mentioned in the variable's description. That way, we're able to add to the implementation if more network policy support is added to GKE.

Regarding the module invocation in integration tests, I don't believe that we should be making changes to paper over this issue. In fact, I'd rather that we _add_ additional testing to cover the case where the network policy is defined, while still making sure that existing integration tests pass as expected. I'm still seeing the same failure after commit https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/129/commits/140492c0bdf1a142da4e8fb8664a2fb9b7f4bce7.

On a related note, I understand that not being able to see the result of our integration tests can be frustrating in processing a pull request. I've brought this up to the group at large for discussion. In the meantime, you can run the currently failing integration test yourself by following the instructions [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine#integration-test). The relevant test instances that are failing are:

- `deploy-service`
- `node-pool`
- `shared-vpc`
- `simple-regional`
- `simple-regional-private`
- `simple-zonal`
- `simple-zonal-private`

However, they all seem to have the same failure mode -- I would suggest getting one of these test cases working, and the rest should follow.

PR #144, opened 2019-04-24
Adapting for use with latest providers

### LINKED ISSUE #143: Terraform Google provider >= 2.4 throws error: "node_pool": conflicts with remove_default_node_pool
I'm just trying out this module for a PoC and ran into some difficulties following the v2.0.0 release.  I created a cluster OK using with v1.0.1 but just upgraded and am now getting this error when running a plan.  

I've stripped back my config and still see the same issue with the main example config in the 
[README.md](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/README.md#usage)
```
Error: module.gke.google_container_cluster.primary: "node_pool": conflicts with remove_default_node_pool

Error: module.gke.google_container_cluster.primary: "remove_default_node_pool": conflicts with node_pool
```

I'm not setting `remove_default_node_pool` option but have tried explicitly setting both `true` and `false` and get the same error.

### Terraform Version

```
Terraform v0.11.13
+ provider.google v2.5.1
+ provider.google-beta v2.5.1
+ provider.kubernetes v1.6.2
+ provider.null v2.1.1
+ provider.random v2.1.1
```

### Affected Resource(s)

* module.gke.google_container_cluster.primary

### Terraform Configuration Files

```tf
provider "google" {
  project     = "<PROJECT ID>"
  region      = "us-central1"
  zone        = "us-central1-a"
}
provider "google-beta" {
  project     = "<PROJECT ID>"
  region      = "us-central1"
  zone        = "us-central1-a"
}
module "gke" {

  source                     = "terraform-google-modules/kubernetes-engine/google"
  project_id                 = "<PROJECT ID>"
  name                       = "gke-test-1"
  region                     = "us-central1"
  zones                      = ["us-central1-a", "us-central1-b", "us-central1-f"]
  network                    = "vpc-01"
  subnetwork                 = "us-central1-01"
  ip_range_pods              = "us-central1-01-gke-01-pods"
  ip_range_services          = "us-central1-01-gke-01-services"
  http_load_balancing        = false
  horizontal_pod_autoscaling = true
  kubernetes_dashboard       = true
  network_policy             = true
  remove_default_node_pool   = true

  node_pools = [
    {
      name               = "default-node-pool"
      machine_type       = "n1-standard-2"
      min_count          = 1
      max_count          = 100
      disk_size_gb       = 100
      disk_type          = "pd-standard"
      image_type         = "COS"
      auto_repair        = true
      auto_upgrade       = true
      service_account    = "project-service-account@<PROJECT ID>.iam.gserviceaccount.com"
      preemptible        = false
      initial_node_count = 80
    },
  ]

  node_pools_oauth_scopes = {
    all = []

    default-node-pool = [
      "https://www.googleapis.com/auth/cloud-platform",
    ]
  }

  node_pools_labels = {
    all = {}

    default-node-pool = {
      default-node-pool = "true"
    }
  }

  node_pools_metadata = {
    all = {}

    default-node-pool = {
      node-pool-metadata-custom-value = "my-node-pool"
    }
  }

  node_pools_taints = {
    all = []

    default-node-pool = [
      {
        key    = "default-node-pool"
        value  = "true"
        effect = "PREFER_NO_SCHEDULE"
      },
    ]
  }

  node_pools_tags = {
    all = []

    default-node-pool = [
      "default-node-pool",
    ]
  }
}
```

### Expected Behavior

Terraform creates plan successfully.

### Actual Behavior

`terraform plan` errors with
```
Warning: module.gke.data.google_container_engine_versions.region: "region": [DEPRECATED] Use location instead



Warning: module.gke.data.google_container_engine_versions.zone: "zone": [DEPRECATED] Use location instead



Warning: module.gke.google_container_cluster.primary: "region": [DEPRECATED] Use location instead



Warning: module.gke.google_container_node_pool.pools: "region": [DEPRECATED] use location instead



Error: module.gke.google_container_cluster.primary: "node_pool": conflicts with remove_default_node_pool



Error: module.gke.google_container_cluster.primary: "remove_default_node_pool": conflicts with node_pool
```

### Steps to Reproduce

1. `terraform plan`

brandoconnor:
## What's in this change?

1. The core fix was removing the `node_pool` inline block in favor of the now supported `initial_node_count`.
2. resources and data lookups now use `location` instead of `zone` or `region` to get remove deprecation warnings.
3. aligning helper script `combine_docfiles.py` with upstream.

## Why make the change?

The motivation here started with being unable to use the module as is at the v2.0.0 release. The `node-pool` inline block was the source of this error:

```
Error: Error applying plan:

1 error(s) occurred:

* module.gke.google_container_cluster.primary: 1 error(s) occurred:

* google_container_cluster.primary: googleapi: Error 400: Cluster.initial_node_count must be greater than zero., badRequest
```

Replacing `zone` and `region` with `location` was motivated by wanting to resolve deprecation warnings:

```
Warning: module.gke.data.google_container_engine_versions.region: "region": [DEPRECATED] Use location instead
Warning: module.gke.data.google_container_engine_versions.zone: "zone": [DEPRECATED] Use location instead
Warning: module.gke.google_container_cluster.primary: "region": [DEPRECATED] Use location instead
Warning: module.gke.google_container_node_pool.pools: "region": [DEPRECATED] use location instead
```

Note: With the new `location` attribute available, it's possible (though a breaking change) we could also move both regional and zonal clusters and pools to be the same resources rather than spread them across two flavors. I didn't do that here because it's rather impactful but potentially a direction to take this next in simplifying.

Fixes #143


autogen/cluster_regional.tf
```diff
@@ -103,6 +95,7 @@ resource "google_container_cluster" "primary" {
 {% endif %}
 
   remove_default_node_pool = "${var.remove_default_node_pool}"
```
morgante: I'm wondering if we should also start setting remove_default_node_pool to `true` by default?

brandoconnor: I tend to think we should default it to `true`. Especially with how minimal the possible configuration is for the default node pool, I don't know that it makes sense to promote using it.

morgante: Agreed, let's consider changing that in a future PR.

examples/deploy_service/main.tf
```diff
@@ -38,18 +38,16 @@ provider "kubernetes" {
...
 
 module "gke" {
-  source     = "../../"
-  project_id = "${var.project_id}"
-  name       = "${local.cluster_type}-cluster${var.cluster_name_suffix}"
-  region     = "${var.region}"
-  network    = "${var.network}"
-  subnetwork = "${var.subnetwork}"
-
-  kubernetes_version = "1.11.7-gke.12"
```
brandoconnor: Removed as it failed the build - this version is no longer supported for the given region.

brandoconnor:
We're in a conflicting state here and since my current client is not looking to even use this module, it's exceedingly difficult to find the time to follow through. If that decision changes (through pain) or some time clears up, I'll revisit this work.

PR #145, opened 2019-04-25
Simplify examples somewhat

morgante:
Right now our examples and test fixtures are a bit complex and brittle.

This is a simple fix:

- Pin the `2.2.x` version instead of 2.x.x` versions
- Remove the unnecessary `compute_engine_service_account` from the `simple_regional_example`

### DIFF
--- a/examples/stub_domains/main.tf
+++ b/examples/stub_domains/main.tf
@@ -19,12 +19,12 @@ locals {
 }
 
 provider "google" {
-  version = "~> 2.2"
+  version = "~> 2.2.0"
   region  = "${var.region}"
 }
 
 provider "google-beta" {
-  version = "~> 2.2"
+  version = "~> 2.2.0"
   region  = "${var.region}"
 }
 
--- a/examples/stub_domains/main.tf
+++ b/examples/stub_domains/main.tf
@@ -19,12 +19,12 @@ locals {
 }
 
 provider "google" {
-  version = "~> 2.2.0"
+  version = "~> 2.3.0"
   region  = "${var.region}"
 }
 
 provider "google-beta" {
-  version = "~> 2.2.0"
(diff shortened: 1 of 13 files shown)


examples/stub_domains/main.tf
```diff
@@ -19,12 +19,12 @@ locals {
 }
 
 provider "google" {
-  version = "~> 2.2"
+  version = "~> 2.2.0"
```
brandoconnor: I found you need the `2.3.0` provider to have the module work as is. The set of provider versions that works currently is pretty narrow... for the moment it might be best to move to the `=` exact version spec.

test/fixtures/simple_regional_private/example.tf
```diff
@@ -24,5 +24,4 @@ module "example" {
   subnetwork                     = "${google_compute_subnetwork.main.name}"
   ip_range_pods                  = "${google_compute_subnetwork.main.secondary_ip_range.0.range_name}"
   ip_range_services              = "${google_compute_subnetwork.main.secondary_ip_range.1.range_name}"
-  compute_engine_service_account = "${var.compute_engine_service_account}"
```
brandoconnor: Is this causing problems too?

morgante: I'd like to get us into the habit of relying on the default behavior more often in fixtures. Default is an auto-generated service account.

examples/stub_domains/main.tf
morgante: Locked to 2.3.x

test/fixtures/simple_regional_private/example.tf
brandoconnor: Wholly agreed! 👍

aaron-lane:
@morgante would you like to update this?

PR #149, opened 2019-05-02
Add initial_node_count variable for scaling the default node pool

ctrox:
Adds a variable to be able to configure the number of nodes in the default node pool. It defaults to 0 so it should not change existing behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ In either case, upgrading to module version `v1.0.0` will trigger a recreation o
 | disable\_legacy\_metadata\_endpoints | Disable the /0.1/ and /v1beta1/ metadata server endpoints on the node. Changing this value will cause all node pools to be recreated. | string | `"true"` | no |
 | horizontal\_pod\_autoscaling | Enable horizontal pod autoscaling addon | string | `"true"` | no |
 | http\_load\_balancing | Enable httpload balancer addon | string | `"true"` | no |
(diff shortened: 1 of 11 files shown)


morgante:
@ctrox Is your intention to use the default node pool? We encourage explicitly passing the node pools you wish to use (via the node_pools argument on this module) and instead deleting the default node pool.

ctrox:
@morgante I actually want to set the initial node count to implement [this workaround](https://github.com/terraform-providers/terraform-provider-google/issues/3385#issuecomment-488149996). But if someone really wants to use the default node pool we would need to have a lot more configuration options available, so I think this one parameter does not make it likely for someone to actually use it.

morgante:
@ctrox Got it, that use case makes sense. Could we potentially *compute* a realistic initial_node_count based on the node_pool configs provided? Maybe you could share your expected configuration?

ctrox:
@morgante It might be a bit hard to compute it with Terraform 0.11 but essentially it would be just adding the `min_node_count` of all node pools together. With that it will launch as many nodes in the default node pool as the other pools will have afterwards. But it would also change the default behaviour of the module if for example someone did not set `delete_node_pool = true` it would suddenly launch a number of nodes.

aaron-lane:
Hi @ctrox.

Can you please move your implementation to the equivalent files in the `autogen` directory, then run `make generate` and `make generate_docs`?

ctrox:
@aaron-lane Totally missed that, sorry! Updated my branch accordingly.

PR #157, opened 2019-05-31
GKE 101: added guest_accelerator feature and tests

alexkonkin:


alexkonkin:
> 
> 
> The changes to `cluster_regional.tf` and `cluster_zonal.tf` need to be moved to the equivalent files under the `autogen` directory. Then, `make generate` and `make generate_docs` need to be executed.
> 
> There is a test failure which must be addressed:
> 
> https://concourse.infra.cft.tips/builds/4176#L5cf138fb:973

Thank you for your review. Done.

alexkonkin:
I have one suggestion regarding the make.sh file. This file has check_generate / check_generate_docs functions. Both functions use --exit-code to detect if changes are present in the source code.
However the same return code comes to the end of the function and then to the target
in the Makefile. As the result the first of these functions makes the execution of the Makefile to fail.
My modification sets the rc (return code) back to 0 which allows all sub targets in the make target to finish.

aaron-lane:
I have tested this locally; the CI failures are unreleated.

dkozlov:
guest_accelerator forces new resource. 
```      node_config.0.guest_accelerator.#:               "0" => "1" (forces new resource)
      node_config.0.guest_accelerator.0.count:         "" => "0" (forces new resource)
```
@alexkonkin please look at my commit which was made 3 months ago: https://github.com/dkozlov/terraform-google-kubernetes-engine/commit/e6952a4a301063c3ceeef5257c9474a6a0e09e33 it works.

aaron-lane:
@dkozlov thank you for catching that. The feature has not been released yet. Can you please open a pull request with your alternative implementation?

PR #159, opened 2019-06-03
Add network_policy to google_container_cluster

aaron-lane:
This branch adds an explicit `network_policy` block to the `google_container_cluster`, as recommended by https://github.com/terraform-providers/terraform-provider-google/issues/3673. This solves the issue of the partially converged suites which obviates the need for a second converge.

### DIFF
--- a/autogen/variables.tf
+++ b/autogen/variables.tf
@@ -106,6 +106,11 @@ variable "network_policy" {
   default     = false
 }
 
+variable "network_policy_provider" {
+  description = "The network policy provider."
+  default     = "PROVIDER_UNSPECIFIED"
+}
+
 variable "maintenance_start_time" {
   description = "Time window specified for daily maintenance operations in RFC3339 format"
   default     = "05:00"
--- a/autogen/variables.tf
+++ b/autogen/variables.tf
@@ -108,7 +108,7 @@ variable "network_policy" {
 
 variable "network_policy_provider" {
   description = "The network policy provider."
-  default     = "PROVIDER_UNSPECIFIED"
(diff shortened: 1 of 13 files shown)


autogen/variables.tf
```diff
@@ -106,6 +106,11 @@ variable "network_policy" {
   default     = false
 }
 
+variable "network_policy_provider" {
+  description = "The network policy provider."
+  default     = "PROVIDER_UNSPECIFIED"
```
morgante: We should probably default to `CALICO` or something which will actually work (unless this automatically works?).

aaron-lane: `PROVIDER_UNSPECIFIED` is the default value of the underlying resource argument and the tests pass, but we can use `CALICO` if it's a more practical default.

morgante: Can you do some more digging on what `PROVIDER_UNSPECIFIED` actually does?

If it defaults to the GKE-recommended/supported provider and network policies work, let's keep it.

If it doesn't configure a provider and hence network policies break, let's change to calico.

aaron-lane: It appears that [Calico is the only supported provider](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#creating_a_network_policy). I can't find any meaningful documentation on the behaviour of `PROVIDER_UNSPECIFIED`, but converging the stub domains suite with this value yields an enabled network policy with Calico pods on the nodes so it seems to select a default provider.

morgante: Got it. Let's specify CALICO as the default value then, because it's more explicit and clear.

PR #165, opened 2019-06-12
Database encryption

marko7460:
* Adding `database_encryption` variable that enables Application level Secrets Encryption as described here: https://cloud.google.com/kubernetes-engine/docs/how-to/encrypting-secrets


autogen/cluster_regional.tf
```diff
@@ -20,13 +20,13 @@
   Create regional cluster
  *****************************************/
 resource "google_container_cluster" "primary" {
-  provider    = "{% if private_cluster %}google-beta{%else %}google{% endif %}"
+  provider    = "google-beta"
```
Jberlinsky: [This section](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/README.md#terraform-and-plugins) of README.md will have to change as well, since it no longer is as described with this PR.

autogen/variables.tf
```diff
@@ -280,3 +280,15 @@ variable "issue_client_certificate" {
   description = "Issues a client certificate to authenticate to the cluster endpoint. To maximize the security of your cluster, leave this option disabled. Client certificates don't automatically rotate and aren't easily revocable. WARNING: changing this after cluster creation is destructive!"
   default     = "false"
 }
+
+variable "database_encryption" {
```
Jberlinsky: We should probably add an example (and, by extension, a test case) for this functionality

README.md
```diff
@@ -116,6 +116,7 @@ In either case, upgrading to module version `v1.0.0` will trigger a recreation o
 |------|-------------|:----:|:-----:|:-----:|
 | basic\_auth\_password | The password to be used with Basic Authentication. | string | `""` | no |
 | basic\_auth\_username | The username to be used with Basic Authentication. An empty value will disable Basic Authentication, which is the recommended configuration. | string | `""` | no |
+| database\_encryption | Application-layer Secrets Encryption settings. Example:   database_encryption = [{     state = "ENCRYPTED",     key_name = "projects/my-project/locations/global/keyRings/my-ring/cryptoKeys/my-key"   }] | list | `<list>` | no |
```
Jberlinsky: Shouldn't the default here be `[]`, not `<list>`?

examples/simple_regional_private/main.tf
```diff
@@ -18,6 +18,11 @@ locals {
   cluster_type = "simple-regional-private"
 }
 
+provider "google" {
```
Jberlinsky: This doesn't seem relevant to the PR as described -- what's going on here?

examples/simple_zonal_private/main.tf
```diff
@@ -18,6 +18,11 @@ locals {
   cluster_type = "simple-regional-private"
 }
 
+provider "google" {
```
Jberlinsky: Same

modules/private-cluster/variables.tf
```diff
@@ -276,3 +276,16 @@ variable "issue_client_certificate" {
   description = "Issues a client certificate to authenticate to the cluster endpoint. To maximize the security of your cluster, leave this option disabled. Client certificates don't automatically rotate and aren't easily revocable. WARNING: changing this after cluster creation is destructive!"
   default     = "false"
 }
+
+variable "database_encryption" {
```
Jberlinsky: See comment above

test/integration/node_pool/controls/kubectl.rb
```diff
@@ -79,11 +79,6 @@
               key: "all-pools-example",
               value: "true",
             },
-            {
```
Jberlinsky: Why was this removed?

README.md
marko7460: This document is generated using `make generate_docs`. I would prefer not to modify anything that is generated by that task.

examples/simple_regional_private/main.tf
marko7460: This was added in order to support `location` field in `data "google_container_engine_versions" "zone"` and `data "google_container_engine_versions" "region"` in `main.tf`. Failing test was: https://concourse.infra.cft.tips/teams/cft/pipelines/terraform-google-kubernetes-engine/jobs/integration-tests/builds/450

HOWEVER: I will be reverting `location` back to `zone` and `region` and thus these changes as well. I will fix deprication warnings for `zone` and `region` in another PR sometime in the future.

examples/simple_zonal_private/main.tf
marko7460: See my comment from the above

autogen/variables.tf
marko7460: i can do that but that example will have to have key ring and key created like in:
https://cloud.google.com/kubernetes-engine/docs/how-to/encrypting-secrets.

I think that one example will be enough to test this feature. It is the same for public of private cluster.

test/integration/node_pool/controls/kubectl.rb
marko7460: For some reason `node-pool` test started to fail when checking `taints`. See `run-test-node-pool` in https://concourse.infra.cft.tips/teams/cft/pipelines/terraform-google-kubernetes-engine/jobs/integration-tests/builds/453 and https://concourse.infra.cft.tips/teams/cft/pipelines/terraform-google-kubernetes-engine/jobs/integration-tests/builds/452.

After making the change in `kubectl.rb` the test started passing again. See: https://concourse.infra.cft.tips/teams/cft/pipelines/terraform-google-kubernetes-engine/jobs/integration-tests/builds/454

I will revert the change to this test case along with `location` changes and i'll see if the test will still fail.

Jberlinsky: If this fixes CI, it should be a separate PR regardless

examples/simple_regional_private/main.tf
Jberlinsky: :+1:

README.md
Jberlinsky: Can we raise a bug about this? Doesn't seem like expected behavior.

autogen/variables.tf
Jberlinsky: Adding KMS resources to a Terraform example/test case should be pretty straightforward. I'd recommend adding it.

README.md
marko7460: Raise a bug on `terraform-docs` or a bug on `make generate_docs`? You can see `master_authorized_networks_config` has the same behavior. 

If you look at https://github.com/terraform-google-modules/terraform-google-project-factory you will see that `shared_vpc_subnets` has the same format.

test/integration/node_pool/controls/kubectl.rb
marko7460: I'll revert the change first and see what happens. If master branch is passing and my branch is failing than i don't see how separate PR would fix this.

README.md
Jberlinsky: `terraform-docs`. This is ok to merge as-is, but it's worth considering raising a bug/inquiry upstream.

test/integration/node_pool/controls/kubectl.rb
Jberlinsky: If that's the case, we should probably understand why this metadata change is integral to tests passing. Let's see how it goes with the reversion.

autogen/variables.tf
marko7460: example and test added.

test/integration/node_pool/controls/kubectl.rb
marko7460: Change reverted. PR https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/171 opened to fix this issue.

Jberlinsky: Thanks!

autogen/README.md
```diff
@@ -137,11 +137,8 @@ The [project factory](https://github.com/terraform-google-modules/terraform-goog
 - [kubectl](https://github.com/kubernetes/kubernetes/releases) 1.9.x
 #### Terraform and Plugins
 - [Terraform](https://www.terraform.io/downloads.html) 0.11.x
-{% if private_cluster %}
 - [terraform-provider-google-beta](https://github.com/terraform-providers/terraform-provider-google-beta) v2.3, v2.6, v2.7
-{% else %}
 - [terraform-provider-google](https://github.com/terraform-providers/terraform-provider-google) v2.3, v2.6, v2.7
```
Jberlinsky: Will removing this cause a breaking change to existing cluster deployments using the non-beta provider?

examples/database_encryption/variables.tf
```diff
@@ -0,0 +1,54 @@
...
+
+variable "ip_range_pods" {
+  description = "The secondary ip range to use for pods"
+}
+
+variable "ip_range_services" {
+  description = "The secondary ip range to use for pods"
+}
+
+variable "database_encryption" {
```
Jberlinsky: Examples in this repository serve both as tangible demonstrations that users can rely on to show how functionality is utilized, and as test cases -- but primarily as demonstrations. With that in mind, it would probably be worth considering not passing this in as a variable, instead making it immediately obvious what's going on.

test/integration/database_encryption/controls/gcloud.rb
```diff
@@ -0,0 +1,44 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+project_id = attribute('project_id')
+location = attribute('location')
+cluster_name = attribute('cluster_name')
+region  = attribute('region')
+
+control "gcloud" do
+  title "GKE "
```
Jberlinsky: Extra whitespace?

autogen/variables.tf
Jberlinsky: Thanks!

autogen/README.md
marko7460: will run more tests.

autogen/cluster_zonal.tf
```diff
@@ -114,6 +114,7 @@ resource "google_container_cluster" "zonal_primary" {
 {% endif %}
 
   remove_default_node_pool = "${var.remove_default_node_pool}"
+  database_encryption      = "${var.database_encryption}"
```
Jberlinsky: Given that this should be a list, can we wrap it with `[]`? e.g. `database_encryption = ["${var.database_encryption}"]`

autogen/README.md
marko7460: This will not cause breaking changes. 3422f69 fixes setting default values that will not cause recreation of the cluster.

test/integration/database_encryption/controls/gcloud.rb
marko7460: No tests will be added for now.

autogen/cluster_zonal.tf
marko7460: done

examples/database_encryption/variables.tf
marko7460: I can create example but no test cases for now. Would that be sufficient?

Jberlinsky: Adding functionality without adding tests significantly increases the failure domain of the module, and I don't think that's a risk we should be taking.

Jberlinsky: Per conversation out of band, let's just make sure https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/185 is addressed once we have Terraform 0.12 compatibility, and this can be merged.

Issue #101, opened 2019-03-13
Add support for "guest_accelerator" in container_node_pool.node_config


dkozlov:
https://www.terraform.io/docs/providers/google/r/container_node_pool.html#node_config
https://www.terraform.io/docs/providers/google/r/container_cluster.html#guest_accelerator

aaron-lane:
Fixed by #157.

aaron-lane:
Fixed by #197.

PR #171, opened 2019-06-13
Fixing intermittent failing test case

marko7460:


alexkonkin:
The snipped that you've commented out was added because of the feature request:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/101

PR #175, opened 2019-06-17
Quote current directory in Makefile

Jberlinsky:
Avoids issues with non-standard directory names, e.g. those with spaces or special characters in them


morgante:
@Jberlinsky Can you update the module template as well?

Jberlinsky:
Absolutely, @morgante!

Jberlinsky:
@morgante Committed upstream here: https://github.com/terraform-google-modules/terraform-google-module-template/pull/21

PR #177, opened 2019-06-18
Add support for Terraform v0.12

aaron-lane:

PR #174, opened 2019-06-15
Add default_max_pods_per_node and max_pods_per_node

algermissen:
This change adds support for `default_max_pods_per_node` at the cluster level and for `max_pods_per_node` at the pool level.

Merging this will also explicitly set the provider to `google-beta` and drop the private/public cluster conditional in the regional cluster (original code commented).

PR #196, opened 2019-06-27
Adding various beta features and code hygiene

chrislovecnm:
This PR completes some basic code hygiene. We moved the database
encryption block in the same if statement as the other beta components.

With this commit we include various beta features that are supported at this time:

Cluster Configurations

- authenticator_groups_config
- vertical_pod_autoscaling
- workload_identity_config
- enable_intranode_visibility
- default_max_pods_per_node

We have TPU, some nodepool options, and other beta features that I have not put in this PR.
Will revisit in later PRs.

This has been tested using the 2.9.1 google provider.  Upgrading the examples to that
provider will occur in a different PR.

### DIFF
--- a/autogen/variables.tf
+++ b/autogen/variables.tf
@@ -300,6 +300,33 @@ variable "pod_security_policy_config" {
     "enabled" = false
   }]
 }
+
+variable "authenticator_groups_config" {
+  description = "Configuration for the Google Groups for GKE feature.  This block has a field named 'security_group' which lists RBAC security group for use with Google security groups in Kubernetes RBAC."
+  default = []
+}
+
+
+variable "vertical_pod_autoscaling" {
+  description = "Vertical Pod Autoscaling automatically adjusts the resources of pods controlled by it."
+  default     =  false
+}
+
+// TODO document "UNSPECIFIED", "SECURE", "EXPOSE", "GKE_METADATA_SERVER"
+variable "workload_identity_config" {
+  description = "Workload Identity allows Kubernetes service accounts to act as a user-managed Google IAM Service Account."
+  default = "UNSPECIFIED"
+}
+
+variable "enable_intranode_visibility" {
+  description = "Whether Intra-node visibility is enabled for this cluster. This makes same node pod to pod traffic visible for VPC network."
+  default = "false"
+}
+
+variable "default_max_pods_per_node" {
+  description = "The default maximum number of pods per node in this cluster. Note that this does not work on node pools which are 'route-based' - that is, node pools belonging to clusters that do not have IP Aliasing enabled."
+  default = "110"
+}
 {% endif %}
 
 variable "basic_auth_username" {
--- a/autogen/cluster_regional.tf
+++ b/autogen/cluster_regional.tf
@@ -43,9 +43,18 @@ resource "google_container_cluster" "primary" {
   monitoring_service = "${var.monitoring_service}"
 
   {% if beta_cluster %}
-  enable_binary_authorization       = "${var.enable_binary_authorization}"
-  pod_security_policy_config        = "${var.pod_security_policy_config}"
+  database_encryption      = ["${var.database_encryption}"]
+  vertical_pod_autoscaling = "${list(map("enabled", var.vertical_pod_autoscaling))}"
+  workload_identity_config = "${list(map("identity_namespace", var.workload_identity_config))}"
+
+  authenticator_groups_config = "${var.authenticator_groups_config}"
+  default_max_pods_per_node   = "${var.default_max_pods_per_node}"
+  enable_intranode_visibility = "${var.enable_intranode_visibility}"
+  enable_binary_authorization = "${var.enable_binary_authorization}"
+  pod_security_policy_config  = "${var.pod_security_policy_config}"
   {% endif %}
+
+
   master_authorized_networks_config = ["${var.master_authorized_networks_config}"]
 
   master_auth {
@@ -124,9 +133,6 @@ resource "google_container_cluster" "primary" {
 {% endif %}
 
   remove_default_node_pool = "${var.remove_default_node_pool}"
-{% if beta_cluster %}
-  database_encryption      = ["${var.database_encryption}"]
-{% endif %}
 }
 
 /******************************************
@@ -185,6 +191,7 @@ resource "google_container_node_pool" "pools" {
     update = "30m"
     delete = "30m"
   }
+
 }
 
 resource "null_resource" "wait_for_regional_cluster" {
--- a/autogen/cluster_zonal.tf
+++ b/autogen/cluster_zonal.tf
@@ -43,8 +43,15 @@ resource "google_container_cluster" "zonal_primary" {
   monitoring_service = "${var.monitoring_service}"
 
   {% if beta_cluster %}
-  enable_binary_authorization       = "${var.enable_binary_authorization}"
-  pod_security_policy_config        = "${var.pod_security_policy_config}"
+  database_encryption      = ["${var.database_encryption}"]
+  vertical_pod_autoscaling = "${list(map("enabled", var.vertical_pod_autoscaling))}"
+  workload_identity_config = "${list(map("identity_namespace", var.workload_identity_config))}"
+
+  authenticator_groups_config = "${var.authenticator_groups_config}"
+  default_max_pods_per_node   = "${var.default_max_pods_per_node}"
+  enable_intranode_visibility = "${var.enable_intranode_visibility}"
+  enable_binary_authorization = "${var.enable_binary_authorization}"
+  pod_security_policy_config  = "${var.pod_security_policy_config}"
   {% endif %}
 
   master_authorized_networks_config = ["${var.master_authorized_networks_config}"]
@@ -75,7 +82,6 @@ resource "google_container_cluster" "zonal_primary" {
       disabled = "${var.network_policy ? 0 : 1}"
     }
     {% if beta_cluster %}
-
     istio_config {
       disabled = "${var.istio ? 0 : 1}"
     }
@@ -125,9 +131,6 @@ resource "google_container_cluster" "zonal_primary" {
 {% endif %}
 
   remove_default_node_pool = "${var.remove_default_node_pool}"
-{% if beta_cluster %}
-  database_encryption      = ["${var.database_encryption}"]
-{% endif %}
 }
 
 /******************************************
--- a/cluster_regional.tf
+++ b/cluster_regional.tf
@@ -42,6 +42,8 @@ resource "google_container_cluster" "primary" {
   logging_service    = "${var.logging_service}"
   monitoring_service = "${var.monitoring_service}"
 
+
+
   master_authorized_networks_config = ["${var.master_authorized_networks_config}"]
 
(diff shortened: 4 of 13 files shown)


morgante:
@chrislovecnm Please drop a ping when this is ready for review (and remove WIP).

chrislovecnm:
@morgante I am currently completing more testing locally, but I hope that I am coded complete.  PTAL

I am going to drop in a PR next to add beta modules to the testing, if it is not too insane.

chrislovecnm:
CI is failing because of a change in the node pool integration test.  We added a `nvidia-tesla-p4` GPU to the nodepool, and because of that the taints have changed.

chrislovecnm:
@morgante adding [WIP] back as I am having some problems with local testing

chrislovecnm:
@morgante PTAL

chrislovecnm:
@morgante these are all additions so these should not be any breaking changes.  But ...

morgante:
@chrislovecnm In theory, I agree. In practice, there can be important differences in setting an argument to an empty value and leaving it off entirely.

aaron-lane:
This partially overlaps with #174. Can we consolidate the two efforts?

chrislovecnm:
@morgante then what do you recommend?  I have tested this locally, and the integration tests have passed, except for the known bug that we have with the node_pool test.

aaron-lane:
@morgante the beta submodules have not yet been released. Are you concerned about the upgrade path from the existing partially-beta private cluster to the new private beta cluster?

chrislovecnm:
@aaron-lane these are beta features, and IMO should be added to the beta modules that we have now.  No idea how we can combine efforts on #174

chrislovecnm:
@aaron-lane if you want me to remove default_max_pods_per_node I can 🤷‍♂

morgante:
@aaron-lane This seems to be a superset of #174, so I'd say we can merge this and add 174 later if we need to.

@aaron-lane Yes, I'm concerned about people using existing clusters upgrading to the private beta cluster. Just want to confirm there's no state plan changes. It's not a merge blocker though, that can be addressed in release testing/migration notes.

chrislovecnm:
@morgante @aaron-lane PTAL and review.  This hopefully with Marko's fix will pass CI now.

autogen/variables.tf
```diff
@@ -300,6 +300,34 @@ variable "pod_security_policy_config" {
...
   }]
 }
+
+variable "authenticator_groups_config" {
+  description = "Configuration for the Google Groups for GKE feature.  This block has a field named 'security_group' which lists RBAC security group for use with Google security groups in Kubernetes RBAC."
+  default = []
+}
+
+
+variable "vertical_pod_autoscaling" {
```
morgante: We shouldn't force the implementation details of this onto users. I'd prefer to just expose this as `vertical_pod_autoscaling_enabled`.

autogen/variables.tf
```diff
@@ -300,6 +300,34 @@ variable "pod_security_policy_config" {
     "enabled" = false
   }]
 }
+
+variable "authenticator_groups_config" {
```
morgante: Instead of exposing this as a list, can we simply expose `authenticator_groups_security_group` directly?

autogen/variables.tf
```diff
@@ -300,6 +300,34 @@ variable "pod_security_policy_config" {
...
+
+
+variable "vertical_pod_autoscaling" {
+  description = "Vertical Pod Autoscaling automatically adjusts the resources of pods controlled by it."
+  default     = [{
+    "enabled" = false
+  }]
+}
+
+variable "workload_identity_config" {
```
morgante: I can't even find an explanation of the expected value here in the docs.

We should either (a) expose the underlying options as separate variables or (b) provide an example of what the input list should look like.

autogen/variables.tf
chrislovecnm: Then we need to fix psp as well.  It is the exact same way.  The challenge is that sometimes the data model gets changed, for example Isto.  We then have to implement that change in the upstream data model as well.  Our istio config values are already out of date in the beta module.

Let me know exactly what you want me to do, and I will focus on it tomorrow.

morgante: psp can be fixed as a separate PR. Please go ahead and update this PR only for the variables you added.

Yes, the underlying data model sometimes changes but I'd rather have us take on the burden of updating with those changes than forcing every module user to update it themselves.

chrislovecnm: I understand that,  I am not certain we can have an empty or default value for say `workload_identity_config`.

With pod security policies we have this:

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/87bdc01f5325c14163cf0e5ba9e03d87857eafd3/modules/beta-public-cluster/variables.tf#L277-L283

Which I think we can set a default variable easily.

With  `workload_identity_config` we need to have something like this:

```
workload_identity_config = [ 
  {
     identity_namespace = "GKE_METADATA_SERVER"
  }
]
```

The values can be "UNSPECIFIED", "SECURE", "EXPOSE", "GKE_METADATA_SERVER".

We can default a value for them, but again we have to set something, because of the data model the upstream provider has chosen.

I would prefer a bit more complex data model on our end that would not require constant updates, and document it a ton.  I found the values in the provider and can map them to the `--workload-metadata-from-node` flag in gcloud.

I understand your concern, just trying to work on a good design that is lower maintenance, and still usable by our users.

autogen/variables.tf
chrislovecnm: The docs are a bit lacking at times.  From values in the source code I was able to map it back to a flag in gcloud:

```
     --workload-metadata-from-node=WORKLOAD_METADATA_FROM_NODE
        Type of metadata server available to pods running in the nodepool.
        WORKLOAD_METADATA_FROM_NODE must be one of:

         EXPOSED
            Pods running in this nodepool have access to the node's underlying
            Compute Engine Metadata Server.
         GKE_METADATA_SERVER
            Run the Kubernetes Engine Metadata Server on this node. The
            Kubernetes Engine Metadata Server exposes a metadata API to
            workloads that is compatible with the V1 Compute Metadata APIs
            exposed by the Compute Engine and App Engine Metadata Servers. This
            feature can only be enabled if Workload Identity is enabled at the
            cluster level.
         SECURE
            Prevents pods not in hostNetwork from accessing certain VM
            metadata, specifically kube-env, which contains Kubelet
            credentials, and the instance identity token. This is a temporary
            security solution available while the bootstrapping process for
            cluster nodes is being redesigned with significant security
            improvements. This feature is scheduled to be deprecated in the
            future and later removed.
         UNSPECIFIED
            Chooses the default.
```

autogen/variables.tf
morgante: > I understand your concern, just trying to work on a good design that is lower maintenance, and still usable by our users.

Even if we document it, I would much prefer that we take on the maintenance burden rather than customers.

If the field format changes, then it's preferable for us to update it in the module (once) rather than forcing every module user to update (potentially hundreds of times).

Admittedly, there are some cases where setting an appropriate default value is challenging but in those I think we can still use a conditional to pass an empty list.

Please proceed with exposing the higher-level options to users (ex. `workload_identity_namespace`) and transforming them into the correct values within the provider itself.

chrislovecnm: ### Current Status

Because we cannot create default values for `authenticator_groups_config`, I am recommending to have the user set the list by hand. I have not been able to figure out how to use a list and a map with a trinary, without major code complexity.  There are solutions, but they are really, really, really ugly complex code.  With tf 0.12, it becomes a lot easier, but we are not there yet.

You get the following type of error:

```hcl
Error: module.gke.google_container_cluster.primary: 1 error(s) occurred:

* module.gke.google_container_cluster.primary: At column 3, line 1: conditional operator cannot be used with map values in:

${var.authenticator_groups_config != "" ? map("security_group", var.authenticator_groups_config) : map()}
```

If you want to read more https://github.com/hashicorp/terraform/issues/12453 - it is a feature request that they implemented in 0.12.

A single list is not bad: https://github.com/hashicorp/terraform/issues/12453#issuecomment-284273475 with 0.11

Just a map alone is complex https://github.com/hashicorp/terraform/issues/12453#issuecomment-327266951

And we have a list with a map inside of it 🤦‍♂.

### Recommendation

`workload_identity_namespace` and `vertical_pod_autoscaling` are very doable.  I recommend `default authenticator_groups_config` default to an empty list, and then the user has to set it to a 

```hcl
default authenticator_groups_config = [
  {
    security_group = "my-security-group-name"
  }
]
```

while `workload_identity_namespace` or `vertical_pod_autoscaling` would be a simpler value like:

```hcl
workload_identity_namespace = "GKE_METADATA_SERVER"
vertical_pod_autoscaling = true
```

autogen/variables.tf
morgante: Yeah, let's try to include those in the docs.

autogen/variables.tf
morgante: @chrislovecnm In the interest of moving forward, I'd suggest we initially merge only `workload_identity_namespace` and `vertical_pod_autoscaling` in this PR, then do a follow-up PR for the authenticator groups config. Does that seem acceptable?

chrislovecnm: @morgante  - @aaron-lane mentioned that he would prefer the changes in the 0.12 branch.  Has that changed?

morgante: I defer to @aaron-lane but our top priority is 0.12 support.

aaron-lane: #177 is on the cusp of being merged, and released tomorrow. :crossed_fingers:

chrislovecnm:
I'm gonna close this, since we have to update all the code for TF 0.12

morgante:
@chrislovecnm Will you have bandwidth to open a new PR with these features?

If not, can you make sure there are open issues so we can have someone on our team prioritize it?

chrislovecnm:
I have no idea tbh

autogen/variables.tf
Dev25: I'm not sure that is correct @chrislovecnm ?

Those values are intended for setting `nodeMetadata` in [node pool config](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/NodeConfig#workloadmetadataconfig) whereas for the `identityNamespace` for [WorkloadIdentityConfig](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.WorkloadIdentityConfig) it should be in the format of 

https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_workload_identity_on_a_new_cluster
```bash
gcloud beta container clusters create [CLUSTER_NAME] \
  --cluster-version=1.12 \
  --identity-namespace=[PROJECT_NAME].svc.id.goog
```

So for terraform its:
```terraform
  workload_identity_config {
    identity_namespace = "${var.project_id}.svc.id.goog"
  }
```

Dev25:
I can open up a PR for some of the beta addons mentioned in this PR next week. It follows the same principle (e.g. PSP) where its using a default empty list + `dynamic/for_each` where needed.

https://github.com/Dev25/terraform-google-kubernetes-engine/compare/gce-labels...Dev25:beta-addons

PR #198, opened 2019-06-27
Updating google providers to 2.9

marko7460:


examples/deploy_service/main.tf
```diff
@@ -19,12 +19,12 @@ locals {
...
 
 provider "google" {
-  version = "~> 2.7.0"
+  version = "~> 2.9.0"
   region  = "${var.region}"
 }
 
 provider "google-beta" {
-  version = "~> 2.7.0"
+  version = "~> 2.9.0"
```
chrislovecnm: 2.9.1 is out ... any reason we do not use that??

morgante: We don't require 2.9.1, so there's no reason to require it. This matches 2.9.X.

chrislovecnm:
@marko7460 

```
kubernetes nodes pool-02 has the expected taints
```

Your taint test failed.

marko7460:
@chrislovecnm that test has been intermittently failing for a while. we removed it in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/171.

aaron-lane:
Ah, we did not remove the test correctly. The assertion should have been updated to look for a partial match of tags rather than an exact match.

marko7460:
@aaron-lane i fixed the issue in this PR but if you want i can create a new PR just with that fix?

test/integration/node_pool/controls/kubectl.rb
```diff
@@ -73,13 +73,13 @@
         end
 
         it "has the expected taints" do
-          expect(taints).to include(
+          expect(taints).to eq([
```
aaron-lane: This appears to have been rebased incorrectly; #200 changed this to `include`.

marko7460: no idea how that happened. It's fixed now.

PR #187, opened 2019-06-21
Creation of new variable configure_ip_masq controlling ip masq install

chrislovecnm:
Aliased IP addresses do not require ip masquerading anymore.  There are  few use cases where we would need ip masq, but usually it is not recommended to install ip masquerading. This variable allows
for fine gain control on the installation of ip masq as it was always installed via the
network_policy variable previously.

configure_ip_masq defaults to false.

Fixes: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/172

TODO:

- [x] test locally

Added a new field 'configure_ip_masq' to public stub domain example.  Removed the
integration control to check for ip masq in the private cluster as well.

PR #201, opened 2019-06-27
README updates

chrislovecnm:
- updated the section about ip masquerading since it was inaccurate since #187 
- made some other notes about the private and beta modules.

Have a comment in a section, because I don't quite understand what we are trying to communicate to the user.


README.md
```diff
@@ -6,7 +6,9 @@ The resources/services/activations/deletions that this module will create/trigge
 - Create GKE Node Pool(s) with provided configuration and attach to cluster
 - Replace the default kube-dns configmap if `stub_domains` are provided
 - Activate network policy if `network_policy` is true
-- Add `ip-masq-agent` configmap with provided `non_masquerade_cidrs` if `network_policy` is true
+- Add `ip-masq-agent` configmap with provided `non_masquerade_cidrs` if `configure_ip_masq` is true
+
+Sub modules are provided from creating private clusters, beta private clusters, and beta public clusters as well.  Beta sub modules allow for the use of various GKE beta features. See the modules directory for the various sub modules.
```
morgante:
```suggestion
Su modules are provided from creating private clusters, beta private clusters, and beta public clusters as well.  Beta sub modules allow for the use of various GKE beta features. See the [modules directory](./modules) for the various submodules.
```

README.md
```diff
@@ -90,6 +92,7 @@ module "gke" {
 }
 ```
 
+<!-- do not understand what this is about -->
```
morgante: These are just the steps to run Terraform. In theory you could copy & paste the above config into a Terraform file then follow these steps to deploy the GKE cluster.

README.md
morgante: We should probably exclude this block in submodules.

README.md
chrislovecnm: I will add what you wrote and it will make more sense to me.

README.md
aaron-lane: I think this was intended to be `Submodules` rather than `Su modules`.

PR #166, opened 2019-06-13
Restrict node access to cluster metadata service

dmyerscough:
I was recently had my private GKE cluster pen test and one of the issues that were reported by this test was POD's had access the metadata service which exposes sensitive data:-

```bash
# curl -H 'Metadata-Flavor: Google' 'http://metadata.google.internal/computeMetadata/v1/instance/attributes/kube-env'
ALLOCATE_NODE_CIDRS: "true"                                                                                                                                            API_SERVER_TEST_LOG_LEVEL: --v=3
...
...
KUBE_PROXY_TOKEN: h9gjoyXnyaB2MDnBeOPmJG3OEJ0zS7GKXtU-9xMJRTw=
KUBELET_CERT: XXXXXXXXXX
KUBELET_KEY: XXXXXXXXX
KUBERNETES_MASTER: "false"
KUBERNETES_MASTER_NAME: 10.17.5.2
LOGGING_DESTINATION: gcp
...
...
```

By default, GKE clusters expose "UNSPECIFIED". However, you can switch this off by setting `node_metadata` to "SECURE".

```bash
$ curl -H 'Metadata-Flavor: Google' 'http://metadata.google.internal/computeMetadata/v1/instance/attributes/kube-env'
This metadata endpoint is concealed.

$ curl -H 'Metadata-Flavor: Google' 'http://metadata.google.internal/computeMetadata/v1/instance/attributes/disable-legacy-endpoints'
false
```

This PR will allow the clusters to move along as normal but will give anyone the ability to shut off metadata service by specifying "SECURE" instead of the default "UNSPECIFIED".

I have tested this on our internal zonal cluster and it worked as expected.

PR #203, opened 2019-07-02
Restrict node access to cluster metadata service

marko7460:
This PR is implementation of https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/166 based on @aaron-lane's last comment.

See https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/166 for more details.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -104,3 +104,10 @@ suites:
       systems:
         - name: stub_domains_private
           backend: local
+  - name: "workload_metadata_config"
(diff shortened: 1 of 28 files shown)


marko7460:
@aaron-lane you are correct that `node_metadata` will cause node-pools to be recreated. I will implement your solution.

aaron-lane:
Thanks @marko7460 and @dmyerscough!

PR #217, opened 2019-07-25
Add Authenticator Groups

Dev25:
Follow up PR after #216 gets merged, will rebase then.

Related PR: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/196

### DIFF
--- a/autogen/main.tf
+++ b/autogen/main.tf
@@ -71,7 +71,7 @@ locals {
     node_metadata = var.node_metadata
   }]
 
-  cluster_authenticator_security_group = var.authenticator_security_group == "" ? [] : [{
+  cluster_authenticator_security_group = var.authenticator_security_group == null ? [] : [{
     security_group = var.authenticator_security_group
   }]
 
--- a/autogen/main.tf
+++ b/autogen/main.tf
@@ -71,6 +71,10 @@ locals {
     node_metadata = var.node_metadata
   }]
 
+  cluster_authenticator_security_group = var.authenticator_security_group == "" ? [] : [{
+    security_group = var.authenticator_security_group
+  }]
+
 {% endif %}
 
   cluster_output_name           = google_container_cluster.primary.name
--- a/autogen/variables.tf
+++ b/autogen/variables.tf
@@ -386,7 +386,7 @@ variable "identity_namespace" {
 variable "authenticator_security_group" {
   type        = string
   description = "The name of the RBAC security group for use with Google security groups in Kubernetes RBAC. Group name must be in format gke-security-groups@yourdomain.com"
-  default     = ""
+  default     = null
 }
 
 {% endif %}
--- a/autogen/variables.tf
+++ b/autogen/variables.tf
@@ -383,5 +383,10 @@ variable "identity_namespace" {
   default     = ""
 }
 
+variable "authenticator_security_group" {
+  type        = string
+  description = "The name of the RBAC security group for use with Google security groups in Kubernetes RBAC. Group name must be in format gke-security-groups@yourdomain.com"
+  default     = ""
+}
 
 {% endif %}
--- a/modules/beta-public-cluster/variables.tf
+++ b/modules/beta-public-cluster/variables.tf
@@ -359,6 +359,6 @@ variable "identity_namespace" {
 variable "authenticator_security_group" {
   type        = string
   description = "The name of the RBAC security group for use with Google security groups in Kubernetes RBAC. Group name must be in format gke-security-groups@yourdomain.com"
-  default     = ""
+  default     = null
 }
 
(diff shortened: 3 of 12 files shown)


autogen/main.tf
```diff
@@ -261,6 +261,17 @@ locals {
     regional = element(concat(google_container_cluster.primary.*.vertical_pod_autoscaling.0.enabled, [""]), 0)
     zonal    = element(concat(google_container_cluster.zonal_primary.*.vertical_pod_autoscaling.0.enabled, [""]), 0)
   }
+
+  cluster_type_workload_identity_namespace = {
+    regional = element(concat(google_container_cluster.primary.*.workload_identity_config.0.identity_namespace, [""]), 0)
+    zonal    = element(concat(google_container_cluster.zonal_primary.*.workload_identity_config.0.identity_namespace, [""]), 0)
```
Dev25: On doing some additional testing, this might need to be rethinked a bit more. Going from a WI enabled cluster to empty list (on another note you can't disable WI on a cluster via terraform without forcing a cluster replacement, https://github.com/terraform-providers/terraform-provider-google/issues/4041) causes terraform errors:

```
Error: Invalid index

  on .terraform/modules/gke/modules/beta-private-cluster/main.tf line 249, in locals:
 249:     regional = element(concat(google_container_cluster.primary.*.workload_identity_config.0.identity_namespace, [""]), 0)

The given key does not identify an element in this collection value.
```

morgante: Yeah, we might consider not having an output for now if you want a quick fix.

Dev25: Removed from output/local vars @morgante

modules/beta-public-cluster/variables.tf
```diff
@@ -343,3 +343,14 @@ variable "enable_vertical_pod_autoscaling" {
...
   default     = false
 }
+
+variable "workload_identity_config" {
+  description = "Workload Identity allows Kubernetes service accounts to act as a user-managed Google IAM Service Account."
+  default     = []
+}
+
+variable "authenticator_groups_config" {
+  type        = list(map(string))
```
morgante: Instead of embedding the map, could we directly accept a list of security groups?

```suggestion
  type        = list(string)
```

Dev25: I've changed it to `list(object({ security_group = string }))` which is a more accurate depiction of the usage and made it clearer in the docs on the object format (similar to 

The initial reasoning in using a map/object was being consistent with how other variables are done when using dynamic blocks e.g. pod security. However on second thought it should be simple to make these variables a string/bool to the user and internally have a local variable/conditonal that converts it into a list (the list format being required for dynamic blocks)

I take a crack at doing the latter in this PR if that sounds reasonable @morgante

morgante: In this case, I'd rather go straight to `list(string)` so we don't introduce an unnecessary breaking change in the future. Updating the others can be a separate PR though.

Does that work?

Dev25: @morgante 

Moving to a single item string list works, but you still get the awkward API for users by forcing a list even though its going to be a single item at max. It also assumes the underlying API structure won't change. Although if that ever does change then i guess a breaking change in this module is fair game. 

These are the options we can go with, personally my preference as a end user would be `string` over any `list()`. If we go a `list()` based type then i don't think it makes sense to move to `string` in the future since that would be a breaking change.

```terraform
# list(object) attempting to mimic the actual provider/gke api structure
authenticator_groups_config = [{
  security_group = "group@yourdomain.com"
}]
workload_identity_config = [{
  identity_namespace = "{your-project-id}.svc.id.goog"
}]

# Moving to list(string)
authenticator_groups_config = ["group@yourdomain.com"]
workload_identity_namespace = ["your-project-id}.svc.id.goog"]

# Simplified, convert to list inside the module using conditional. Default is ""
authenticator_groups_config = "group@yourdomain.com"
workload_identity_namespace = "{your-project-id}.svc.id.goog"
```

morgante: Agreed, I'm in favor of moving straight to a `string`.

Dev25: PR updated to just include these changes for authenticator groups since Workload Identity has already been merged to master, ready for review @morgante

autogen/variables.tf
```diff
@@ -383,4 +383,10 @@ variable "identity_namespace" {
   default     = ""
 }
 
+variable "authenticator_security_group" {
+  type        = string
+  description = "The name of the RBAC security group for use with Google security groups in Kubernetes RBAC. Group name must be in format gke-security-groups@yourdomain.com"
+  default     = ""
```
morgante: Can we use `null` as the default?

Dev25: Yep, done.

Issue #170, opened 2019-06-13
Support for default_max_nodes_per_host


algermissen:
Hi,
I would like to add default_max_nodes_per_host parameter. I have created a branch for this: https://github.com/algermissen/terraform-google-kubernetes-engine/tree/add-default-max-pods-per-node

However, I had to hard-code the beta-provisioner and this prevents me from raising a PR.

Can you give any advice how you would like to have it integrated? Currently it is dependent on another condition and adding to that will make everything less clear.

Please provide direction and I'll adjust and raise PR.

aaron-lane:
Hi @algermissen. Thank you for your interest in contributing! 

We are considering a redesign of this module in the near future which may require use of the beta provider for both public and private clusters, so this is not an unprecedented change. Please propose your change and we will try to include it when the new version is released.

algermissen:
Hi @aaron-lane PR is here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/174

morgante:
Fixed by #220.

Issue #209, opened 2019-07-19
Incomplete documentation for terraform_docs.sh dependancy


Dev25:
As part of the migration to Terraform 0.12 #177, https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/177/commits/eb0a0e77bcfd84d3581d76b36b338ecc1ff0a96e#diff-e9cfabf9b162cebf5922498a77b8f766R103 introduced a depedancy on a file from https://github.com/antonbabenko/pre-commit-terraform which is not mentioned in the README. 

Trying to run `make generate_docs` on master without the script in your $PATH will lead to errors, `terraform_docs.sh` is required because upstream terraform-docs does not support Terraform 0.12.

morgante:
FYI we're planning to standardize on using Docker images to run all tests, which should solve this.

aaron-lane:
This will be fixed by #251.

aaron-lane:
Fixed by #262.

PR #220, opened 2019-07-30
Add default_max_pods_per_node and max_pods_per_node after beta_cluster templating 

richardmcsong:
Relates to #174 , #170. I've transplanted the changes made by @algermissen. However, it seems like the `make generate_docs` is failing with:
```
Working in ./modules/private-cluster ...
test/make.sh: line 140: terraform_docs.sh: command not found
Warning! Exit code: 127
```

Am I misconfiguring something? I have terraform-docs 0.6.0 installed.


Dev25:
@richardmcsong you need to install https://github.com/antonbabenko/pre-commit-terraform or at least have that terraform_docs script in your $PATH. 

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/209

autogen/cluster_regional.tf
```diff
@@ -59,6 +59,7 @@ resource "google_container_cluster" "primary" {
 {% if beta_cluster %}
   enable_binary_authorization = var.enable_binary_authorization
   enable_intranode_visibility = var.enable_intranode_visibility
+  default_max_pods_per_node = var.default_max_pods_per_node
```
morgante: Formatting needs to be flush with line above.

autogen/cluster_zonal.tf
```diff
@@ -204,7 +205,9 @@ resource "google_container_node_pool" "zonal_pools" {
     "initial_node_count",
     lookup(var.node_pools[count.index], "min_count", 1),
   )
-
+  {% if beta_cluster %}
+  max_pods_per_node  = lookup(var.node_pools[count.index], "max_pods_per_node", 110)
```
morgante: Should this reference var. default_max_pods_per_node instead of harcoding 110?

richardmcsong: great point.

var.default_max_pods_per_node is a terraform attribute for google_container_cluster though. maybe we should be looking to only set max_pods_per_node if the node pool has max_pods_per_node set in the var.node_pools map?

morgante: Agreed. We can likely use the Terraform 0.12 `nil` value for this use case.

richardmcsong: Great shout out, I didn't even know about that functionality! Added null as the default value for the lookup

PR #230, opened 2019-08-13
Added check for beta feature field istio_enabled

### LINKED ISSUE #226: Certain fields may be missing from google's API call while refreshing state
In particular, addonsConfig.istioConfig, PodSecurityPolicyConfig, VerticalPodAutoscaling seem to sometimes be missing. I've raised a Google support ticket, and they've confirmed that it seems to be expected behaviour for now. This behaviour will raise errors when this module checks for those fields.

richardmcsong:
Some beta feature fields may not be returned by the Google API call that terraform uses.

According to Google Support (Case 20165122), when asked why certain Google API calls don't return `istio_config` or `pod_security_policy_config`: 
```
Actually the fields are not optional. If you have those features enabled in your cluster it will show up in the description. These fields default to disabled and when they are disabled they are not present in the rest resource which is why these fields are missing in the cluster description. The API is configured to register them as disabled if left blank.
```

Since the terraform provider upstream operates by transpiling the API call into the correct resources, I thought it might be most appropriate to code defensively here and check for the field before trying to access it. Fixes #226 

I also added resource_usage_export_config support in this PR (https://www.terraform.io/docs/providers/google/r/container_cluster.html#resource_usage_export_config).

### DIFF
--- a/autogen/main.tf
+++ b/autogen/main.tf
@@ -104,7 +104,7 @@ locals {
 
 {% if beta_cluster %}
   # BETA features
-  cluster_output_istio_enabled                    = google_container_cluster.primary.addons_config.0.istio_config.0.disabled
+  cluster_output_istio_disabled                    = google_container_cluster.primary.addons_config.0.istio_config != null && length(google_container_cluster.primary.addons_config.0.istio_config) == 1 ? google_container_cluster.primary.addons_config.0.istio_config.0.disabled : false
   cluster_output_pod_security_policy_enabled      = google_container_cluster.primary.pod_security_policy_config != null && length(google_container_cluster.primary.pod_security_policy_config) == 1 ? google_container_cluster.primary.pod_security_policy_config.0.enabled : false
   cluster_output_intranode_visbility_enabled      = google_container_cluster.primary.enable_intranode_visibility
   cluster_output_vertical_pod_autoscaling_enabled = google_container_cluster.primary.vertical_pod_autoscaling != null && length(google_container_cluster.primary.vertical_pod_autoscaling) == 1 ? google_container_cluster.primary.vertical_pod_autoscaling.0.enabled : false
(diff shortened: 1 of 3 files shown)


autogen/variables.tf
```diff
@@ -354,6 +354,14 @@ variable "pod_security_policy_config" {
   }]
 }
 
+variable "resource_usage_export_config" {
+  description = "Enables network egress metering for this cluster. If enabled, a daemonset will be created in the cluster to meter network egress traffic."
+  default     = [{
```
morgante: Is there a reason this needs to be a list? Seems like it should be an object: https://www.terraform.io/docs/configuration/types.html#object-

richardmcsong: I actually just copied the pod_security_policy_config variable, as I reused the same pattern for that implementation. My bad for blindly copying code without understanding it. Should we change it, or maybe both, to pure objects instead?

morgante: We should change this one to be an object.

The other one could be changed in a future PR, but it's potentially a breaking change so I'd rather keep it disconnected.

morgante:
FYI we have consolidated the zonal and regional resources into a single resource so this PR will need to be rebased. Thanks!

richardmcsong:
image

I'm sorry if this is a really stupid question, but why is this happening? it doesn't error out if i swap over to `google_container_cluster.primary.addons_config.0.istio_config == []`, but i thought when the expression evals to false that it goes to the second result after the colon, not to the first... Am I going nuts?

paulpalamarchuk:
> image
> 
> I'm sorry if this is a really stupid question, but why is this happening? it doesn't error out if i swap over to `google_container_cluster.primary.addons_config.0.istio_config == []`, but i thought when the expression evals to false that it goes to the second result after the colon, not to the first... Am I going nuts?

@richardmcsong `google_container_cluster.primary.addons_config.0.istio_config != []` not the same as `google_container_cluster.primary.addons_config.0.istio_config != [{}]` (empty list is not equal to list with one element that is empty object).

paulpalamarchuk:
@richardmcsong Define, please, version of provider `google-beta` "~> 2.10.0", to use [resource_usage_export_config](https://github.com/terraform-providers/terraform-provider-google-beta/pull/825) and remove older version mentions. Thank you

richardmcsong:
Finally got a chance to rebase. Noticed the other beta features were already done, just missing the istio one. Also fixed some naming, because it's actually detecting for the disabled flag, not the enabled flag.

autogen/variables.tf
richardmcsong: Fixed in #238

PR #234, opened 2019-08-14
Add workload identity support

sylvioneto:
Workload Identity is already available in beta and it works well.
This PR adds it to beta-private-cluster.

PR #237, opened 2019-08-17
Make workload_identity_config dynamic

sylvioneto:
This PR makes workload_identity_config dynamic in the cluster configuration in order to solve the problem reported by @richardmcsong  in this PR https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/234#pullrequestreview-276211508, when a cluster was created without that config and should not prompt any change in the terraform plan.


modules/beta-public-cluster/variables.tf
```diff
@@ -350,9 +350,10 @@ variable "enable_vertical_pod_autoscaling" {
...
 }
 
-variable "identity_namespace" {
-  type        = string
-  description = "Workload Identity namespace"
-  default     = ""
+variable "workload_identity_config" {
+  description = "Workload Identity configuration"
+  type        = list(object({ identity_namespace = string }))
+  default     = []
```
morgante: We shouldn't force users to specify this as a list. I'd like to keep the API the same.

We should be able to build a list within the module based on the value provided.

sylvioneto: Done, The module builds the list now and uses former API .
I ran it here without identity_namespace and the cluster has been created successfully.

README.md
```diff
@@ -125,78 +125,6 @@ Version 1.0.0 of this module introduces a breaking change: adding the `disable-l
 In either case, upgrading to module version `v1.0.0` will trigger a recreation of all node pools in the cluster.
 
 <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
```
kopachevsky: please revert documentation changes, each `make genereate` will erase autogenerated documentation, and should be followed by `make generate_docs`

kopachevsky:
@sylvioneto  can you please update provider version for ./examples in this PR as well?

README.md
sylvioneto: done

sylvioneto:
Closed, it requires more tests.

sylvioneto:
Reopening the PR because tests worked successfully. Solution was creating a local `local.cluster_workload_identity_config` to use in the dynamic block. If the var identity_namespace is empty, this local var is a empty list so that a empty workload_identity_config block is not added to the cluster.

sylvioneto:
@kopachevsky, @morgante, all changes done.

PR #241, opened 2019-08-21
Add flag to enable GKE Sandbox

### LINKED ISSUE #240: Add sandbox flag for node-pools
It would be nice if support for the [sandbox](https://cloud.google.com/kubernetes-engine/sandbox/) was able to be enabled. [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/sandbox-pods)

paulpalamarchuk:
* Add `sandbox_enabled` variable to use Sandbox (Fixes #240)


kopachevsky:
Do we handle this ? DOn't sure it's on our side, but still

https://prnt.sc/ow3qpx

kopachevsky:
Also, according to this https://cloud.google.com/kubernetes-engine/docs/how-to/sandbox-pods
it's not compatible with istio  do we need to track istio/sandbox combination?

paulpalamarchuk:
@kopachevsky thank you for the notice. I will update PR

PR #243, opened 2019-08-23
Fix linter

anantasty:
Linter in check_generate and check_generate docs was raising an issue due to differences in
the existing README files and what was auto generated. The auto generated README's remove large portions of the README. 
This needs to be a separate issue to fix the auto generate issues in the docs. The current PR raises no linter issues but is not ready to be merged.


aaron-lane:
Fixed by #292

PR #247, opened 2019-08-27
Changed `region` variable

### LINKED ISSUE #235: Region variable should be optional for zonal clusters
Currently, the 'region' variable is always required: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/variables.tf#L41

We should only require this be specified for regional clusters (not for zonal clusters).

paulpalamarchuk:
* `region` variable became optional for zonal clusters (Fixes #235)


autogen/main.tf
```diff
@@ -27,7 +27,7 @@ data "google_compute_zones" "available" {
   {% endif %}
 
   project = var.project_id
-  region  = var.region
+  region  = var.region ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
kopachevsky: should it be 
`var.region ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))`
or 
`var.regional ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))`

kopachevsky: in first case you check if var region is passed in second if cluster regional

paulpalamarchuk: In that case I want to make [`region` variable optional for zonal clusters](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/235)

kopachevsky: I meant in this case if user pass var.regional=true and will not pass var.region, it will have random region right?

kopachevsky: sorry, not random, but region from zones

paulpalamarchuk: @kopachevsky in that case we'll got the region name based on the zones, if the `zones` var set. If it's not set - run fails

aaron-lane: Let's capture this logic as a `local` instead of duplicating between `main.tf` and `networks.tf`.

paulpalamarchuk: Logic moved to `local` section.

autogen/main.tf
```diff
@@ -38,6 +38,7 @@ resource "random_shuffle" "available_zones" {
 locals {
   // location
   location = var.regional ? var.region : var.zones[0]
+  region   = var.region ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
aaron-lane: Condition values can not be `null` so checking `var.region` directly is invalid.

```suggestion
  region   = var.region != null ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```

paulpalamarchuk: fixed

autogen/main.tf
```diff
@@ -38,6 +38,7 @@ resource "random_shuffle" "available_zones" {
 locals {
   // location
   location = var.regional ? var.region : var.zones[0]
+  region   = var.region != null ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
Jberlinsky: This would be more clear if inverted -- e.g. `region = var.region == null ? join("-", slice(split("-", var.zones[0]), 0, 2)) : var.region`. As written, it's hard to tell exactly what's going on.

main.tf
```diff
@@ -34,6 +34,7 @@ resource "random_shuffle" "available_zones" {
 locals {
   // location
   location = var.regional ? var.region : var.zones[0]
+  region   = var.region != null ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
Jberlinsky: See above -- this should be taken care of by making the change above and re-generating.

modules/beta-private-cluster/main.tf
```diff
@@ -34,6 +34,7 @@ resource "random_shuffle" "available_zones" {
 locals {
   // location
   location = var.regional ? var.region : var.zones[0]
+  region   = var.region != null ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
Jberlinsky: See above

modules/beta-public-cluster/main.tf
```diff
@@ -34,6 +34,7 @@ resource "random_shuffle" "available_zones" {
 locals {
   // location
   location = var.regional ? var.region : var.zones[0]
+  region   = var.region != null ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
Jberlinsky: See above

modules/private-cluster/main.tf
```diff
@@ -34,6 +34,7 @@ resource "random_shuffle" "available_zones" {
 locals {
   // location
   location = var.regional ? var.region : var.zones[0]
+  region   = var.region != null ? var.region : join("-", slice(split("-", var.zones[0]), 0, 2))
```
Jberlinsky: See above

autogen/main.tf
paulpalamarchuk: Reworked

main.tf
paulpalamarchuk: updated

modules/beta-private-cluster/main.tf
paulpalamarchuk: updated

modules/private-cluster/main.tf
paulpalamarchuk: updated

aaron-lane:
Please fix any Cloud Build issues. Terraform lint issues should be addressed in the autogen directory so that the generated files do not require changes.

kopachevsky:
Approved when cloud build is green

Issue #139, opened 2019-04-17
Suggest enabling metadata-concealment by default


bgeesaman:
Strongly suggest that the metadata-concealment proxy be enabled to protect cluster privilege escalation attacks (without this control in place, any pod has the ability to use the instance metadata API to obtain the kubelet's credentials which provides a path to gain access to all cluster secrets).

https://www.terraform.io/docs/providers/google/r/container_cluster.html#node_metadata
e.g.
```
workload_metadata_config {
  node_metadata = "SECURE"
}
```
See: https://cloud.google.com/kubernetes-engine/docs/how-to/protecting-cluster-metadata#concealment https://www.4armed.com/blog/kubeletmein-kubelet-hacking-tool/ and https://www.qwiklabs.com/focuses/5158?parent=catalog for more background info on why this control is so important.

morgante:
We should expose this as a variable (with a `SECURE` default).

morgante:
We have exposed this (as `node_metadata`) but should make the default SECURE.

PR #248, opened 2019-08-27
Enabling metadata-concealment by default

ingwarr:
Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/139


autogen/variables.tf
```diff
@@ -368,7 +368,7 @@ variable "pod_security_policy_config" {
 
 variable "node_metadata" {
```
paulpalamarchuk: Add please variable type

ingwarr: Thank you!

PR #255, opened 2019-08-29
Set cluster initial node count from var

julianvmodesto:
Currently, the `var.initial_node_count` is not specified from the variable.

This forces a cluster replacement:

```
      - initial_node_count          = 1 -> null # forces replacement
```

An alternative is adding to `ignore_lifecycle`, like what we do with the `node_pool.*.initial_node_count`


julianvmodesto:
Done, PTAL @morgante

PR #256, opened 2019-09-04
Allow a node pool to be created before it is destroyed

asproul:
This change is inspired by https://github.com/terraform-providers/terraform-provider-google/issues/1054#issuecomment-363971251 . 

Currently, if a node pool has to be recreated for any number of reasons, the node pool is deleted then, created. This can be a problem if it is the only node pool in the gke cluster and the new node pool cannot be provisioned. In this scenario, pods could not be scheduled.

This change allows a node pool to be created before it is deleted so that any issues with node pool creation and/or provisioning are discovered before the node pool is removed. This feature is controlled by the variable `node_pools_create_before_destroy`. In order to avoid node pool name collisions, a 4 character alphanumeric is added as a suffix to the name.


autogen/cluster.tf
```diff
@@ -219,14 +219,31 @@ resource "google_container_cluster" "primary" {
 /******************************************
   Create Container Cluster node pools
  *****************************************/
+resource "random_id" "name" {
+  # if any node_pool definition has a create_before_destroy key, then create random_id names
```
morgante: Generating the random_id is a safe operation, so I think we could simply drop this `count` and simplify the code.

autogen/cluster.tf
```diff
@@ -219,14 +219,31 @@ resource "google_container_cluster" "primary" {
...
+  count = length(compact([for node_pool in var.node_pools : lookup(node_pool, "create_before_destroy", "")])) > 0 ? length(var.node_pools) : 0
+
+  byte_length = 2
+  prefix      = format("%s-", lookup(var.node_pools[count.index], "name"))
+
+  keepers = {
```
morgante: How are we tracking which fields force a recreate and which can be updated in-place?

autogen/cluster.tf
asproul: Sounds good. I will update `count`. 

My original thinking was that if one were to see the `random_id.name` changing and there are no node pools with `create_before_destroy` defined, it would be confusing to the user of this module because they might think something is changing when in fact nothing is changing for them or their node pools.

autogen/cluster.tf
asproul: I am not exactly sure how to track which resources should be added to keepers other than looking at the node pool schema and seeing which ones have "ForceNew" set. What do you think about adding any resources specified with `ForceNew: true` in the [schemaNodeConfig](https://github.com/terraform-providers/terraform-provider-google/blob/master/google/node_config.go#L22) variable to keepers? 

Then adding a comment above keepers to signify which keepers how to determine which resources to add.

asproul: Following the `ForceNew: true` method, one way to signify to track the keepers is through a local variable like `resources_which_force_node_pool_recreation` and iterating through the list to lookup those resources in the node_pool. 

```hcl
locals {
  resources_which_force_node_pool_recreation = [
    "disk_size_gb",
    "disk_type",
    "accelerator_count",
    "accelerator_type",
    "local_ssd_count",
    "machine_type",
    "preemptible",
    "service_account",
  ]
}

resource "random_id" "name" {
  count       = length(var.node_pools)
  byte_length = 2
  prefix      = format("%s-", lookup(var.node_pools[count.index], "name"))
  keepers     = zipmap(
      local.resources_which_force_node_pool_recreation,
      [ for keeper in local.resources_which_force_node_pool_recreation : lookup(var.node_pools[count.index], keeper, "") ]
  )
}
```

The problem that I am running into is the list of map resources such as `labels`, `metadata`, `oauth_scopes`, and `tags` which do not conform to the same pattern as these resources are a result from merging the `all` tag with the node_pool config. 

What do you think about retaining a small list of keepers with a local such as `resources_which_force_node_pool_recreation` ?

asproul: The latest commit tracks all resources including `labels`, `metadata`, oauth_scopes`, and `tags` by converting each map or list into a string and creating a map for keepers.

asproul:
PTAL @morgante

autogen/cluster.tf
morgante: Perfect, thanks. Could we add a comment that this keepers list should be based on the `ForceNew` params?

autogen/cluster.tf
```diff
@@ -219,14 +219,75 @@ resource "google_container_cluster" "primary" {
...
+  count       = var.node_pools_create_before_destroy ? length(var.node_pools) : 0
+  byte_length = 2
+  prefix      = format("%s-", lookup(var.node_pools[count.index], "name"))
+  keepers = merge(
+    zipmap(
+      local.force_node_pool_recreation_resources,
+      [for keeper in local.force_node_pool_recreation_resources : lookup(var.node_pools[count.index], keeper, "")]
+    ),
+    {
+      labels_all = join(",", keys(var.node_pools_labels["all"]), values(var.node_pools_labels["all"]))
```
morgante: I think we need to combine `labels_all` and `labels_node_pool` into a single keeper. Otherwise moving a label from `node_pool` to `all` could incorrectly force a recreation. (Similar for metadata, oauth, etc.)

autogen/cluster.tf
```diff
@@ -219,14 +219,75 @@ resource "google_container_cluster" "primary" {
...
+    {
+      tags_all = join(",", var.node_pools_tags["all"])
+    },
+    {
+      tags_node_pool = join(",", var.node_pools_tags[var.node_pools[count.index]["name"]])
+    }
+  )
+}
+
+{% for create_before_destroy_value in ["false", "true"] %}
```
morgante: Is there a reason these need to be two different node pool resources? I'd like to avoid that complexity as much as possible.

asproul: Unfortunately, the `lifecycle` block does not support interpolations [according to the docs](https://www.terraform.io/docs/configuration/resources.html#lifecycle-lifecycle-customizations):

> The `lifecycle` settings all effect how Terraform constructs and traverses the dependency graph. As a result, only literal values can be used because the processing happens too early for arbitrary expression evaluation.

Additionally, discussed further in https://github.com/hashicorp/terraform/issues/3116

As I see it, the value has to be a literal in the lifecycle block hence the templating of two different node pools. 

If you want to avoid the two different node pool resources, it sounds like there are a couple of options until the lifecycle block can use interpolations:
1 - switch to `create_before_destroy = true` for all node pools
2 - no support for `create_before_destroy = true`
3 - ???

Maybe there is another option?

autogen/cluster.tf
asproul: Makes sense. Will update

asproul: updated.

autogen/cluster.tf
morgante: I see. In that case, I think we might want to switch to an additional (templated) submodule?

asproul: To be clear, your suggestion is to add another `module` to the `MODULES` variable here: 
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/helpers/generate_modules/generate_modules.py#L44

Which contains another variable such as `'lifecycle_variant': True` and add some templating to `autogen/cluster.tf` based on this variable, correct?

autogen/cluster.tf
asproul: added.

autogen/cluster.tf
asproul: Created a new templated submodule, `private-cluster-lifecycle-variant`, (I'm open to other names. Not sure what to call this module(s)) which adds the `create_before_destroy = true` to the lifecycle block.

asproul: @morgante - is `private-cluster-lifecycle-variant` along the lines of what you were suggesting?

morgante: Yes, that's exactly what I was thinking. We might try for a more descriptive name though? `update-variant`?

autogen/cluster.tf
```diff
@@ -233,6 +233,8 @@ locals {
   ]
 }
 
+# This keepers list is based on the terraform google provider schemaNodeConfig
+# resources where "ForceNew" is "true"
```
morgante: It'd be good to have a direct link to the code as well.

asproul: added a link to the source.

autogen/cluster.tf
asproul: done. replaced `lifecycle-variant` with `update-variant`.

autogen/cluster.tf
```diff
@@ -219,14 +219,76 @@ resource "google_container_cluster" "primary" {
...
+  count       = length(var.node_pools)
+  byte_length = 2
+  prefix      = format("%s-", lookup(var.node_pools[count.index], "name"))
+  keepers = merge(
+    zipmap(
+      local.force_node_pool_recreation_resources,
+      [for keeper in local.force_node_pool_recreation_resources : lookup(var.node_pools[count.index], keeper, "")]
+    ),
+    {
+      labels = join(",",
```
morgante: We should probably sort these so reordering has no effect?

asproul: updated.

asproul:
PTAL @morgante

asproul:
PTAL @morgante

morgante:
Thanks for the contribution!

PR #257, opened 2019-09-05
Move to using for_each for node pools

### LINKED ISSUE #128: Change node pool resources to use for_each
In the circumstance that a node pool must be replaced and workload transitioned with zero downtime, having all node pools defined in a list and launched with a single [`google_container_node_pool`](https://www.terraform.io/docs/providers/google/r/container_node_pool.html) resource makes it difficult to carry out.

For example, if two node pools are defined and node pool 0 must be replaced. It seems like the only safe approach is to temporarily create a third node pool, transition the workload from node pool 0 to the new node pool, change or replace node pool 0, transition the workload back to node pool 0, and finally destroy the temporary node pool so you're back to two. Which is fine, but it's probably more work than necessary.

Another consideration is if there are many node pools defined and you need to destroy node pool 0 completely. There is no way to do this without affecting node pool 1 and any other subsequently defined node pools.

This adaptability seems to be a limitation of this module. But also perhaps the use of `count` in resources in general. It might be better if users had control over their node pools as independent resources. However, leveraging `count` and a list of node pools is likely the only way to make any GKE module flexible enough for broad adoption given the current limitations of Terraform.

I'm opening this issue to see if there is a better way. Or if we can come up with a way to improve conditions.

Dev25:
Use for_each instead of count/index for managing node pools. This is going to be a breaking change for users and they will need to manually edit the state to migrate node pools over.

State Migration:
- Remove existing state: `terraform state rm module.<name>.google_container_node_pool.pools`
- Import each node pool: `terraform import 'module.<name>.google_container_node_pool.pools["name"]' {region}/{cluster}{name}`
  - Example: `terraform import 'module.gke.google_container_node_pool.pools["spot-pool"]' europe-west1/dev-cluster/spot-pool`

Fixes #128


autogen/main.tf
```diff
@@ -112,8 +112,8 @@ locals {
   # /BETA features
   {% endif %}
 
-  cluster_output_node_pools_names    = concat(google_container_node_pool.pools.*.name, [""])
-  cluster_output_node_pools_versions = concat(google_container_node_pool.pools.*.version, [""])
+  cluster_output_node_pools_names    = concat([for np in google_container_node_pool.pools : np.name], [""])
+  cluster_output_node_pools_versions = concat([for np in google_container_node_pool.pools : np.version], [""])
```
Dev25: FYI splat no longer works on resources created by `for_each`
https://github.com/hashicorp/terraform/issues/22562

autogen/variables.tf
```diff
@@ -148,14 +148,9 @@ variable "disable_legacy_metadata_endpoints" {
...
-  description = "List of maps containing node pools"
-
-  default = [
-    {
-      name = "default-node-pool"
-    },
-  ]
+  type        = map
+  description = "map containing node pools"
+  default = {}
```
morgante: Needs to be formatted.

autogen/variables.tf
```diff
@@ -148,14 +148,9 @@ variable "disable_legacy_metadata_endpoints" {
...
-  type        = list(map(string))
-  description = "List of maps containing node pools"
-
-  default = [
-    {
-      name = "default-node-pool"
-    },
-  ]
+  type        = map
+  description = "map containing node pools"
```
morgante: I don't think we need to make this into a map? I'd like to keep the external interface the same.

autogen/cluster.tf
```diff
@@ -157,7 +157,7 @@ resource "google_container_cluster" "primary" {
     initial_node_count = var.initial_node_count
 
     node_config {
-      service_account = lookup(var.node_pools[0], "service_account", local.service_account)
+      service_account = local.service_account
```
morgante: I think this is an unrelated change?

autogen/cluster.tf
```diff
@@ -225,62 +225,63 @@ resource "google_container_node_pool" "pools" {
   {% else %}
   provider = google
   {% endif %}
-  count    = length(var.node_pools)
-  name     = var.node_pools[count.index]["name"]
+  for_each = var.node_pools
```
morgante: Let's keep the external interface as list but build the map internally within the module.

morgante:
@Dev25 Just checking if you think you'll be able to make the requested changes (and rebase)?

Dev25:
Yep, hopefully within the next week or 2 @morgante

autogen/variables.tf
Dev25: Will keep the existing interface via 
```tf
node_pool_names = [for np in toset(var.node_pools) : np.name]
node_pools = zipmap(local.node_pool_names, tolist(toset(var.node_pools)))
```

autogen/cluster.tf
Dev25: Will keep the existing interface via having this in locals:
```tf
node_pool_names = [for np in toset(var.node_pools) : np.name]
node_pools = zipmap(local.node_pool_names, tolist(toset(var.node_pools)))
```

autogen/cluster.tf
Dev25: Reverted.

Dev25:
Rebased and kept the same external interface @morgante 

Will be porting over my existing clusters to this rebased version this week which should further test it, PTAL

Dev25:
@morgante Can you take another look now? CI is passing after the recent master fixes.

PR #273, opened 2019-10-01
Add a parameter 'registry_project_id' to the GKE module templates

mmontan:
The PR allows configuring the project holding the GCR registry when
used in connection with 'create_service_account'=true and
grant_registry_access=true.

Holding the GCR is a project with other resources increases the risk of
exposing sensitive data to the service account running the nodes, as the
required permissions of role roles/storage.objectViewer provide access
to all storage objects in the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,22 @@ Then perform the following commands on the root folder:
 - `terraform apply` to apply the infrastructure build
 - `terraform destroy` to destroy the built infrastructure
 
+## Upgrade to v3.0.0
+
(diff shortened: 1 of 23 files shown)


mmontan:
The Cloud Build check pass now. I can't access logs from the concourse builds, so it's hard to debug them. How do I find the logs?

morgante:
>  I can't access logs from the concourse builds, so it's hard to debug them. 

You can ignore Concourse tests. They're legacy and no longer needed.

PR #282, opened 2019-10-16
Make default metadata, labels and tags optional in beta private cluster

kevang:
**SUMMARY**
A number of default tags, metadata and labels are added to the nodepools by the different modules. This is fine when creating a new nodepool, but can be problematic when importing an existing nodepool that does not have those, as a new plan will lead to nodepool recreation, for example:
```
~ labels            = { # forces replacement
     + "cluster_name" = "cluster-somename-9965"
     + "node_pool"    = "basic-pool"
}
```

**PROPOSED CHANGE**
The proposed change makes it possible to use the modules with existing gke clusters/nodepools. It's not introducing any breaking changes and was tested with the beta private cluster. I'm happy to update the other modules, if you think the change makes sense.

### DIFF
--- a/autogen/cluster.tf
+++ b/autogen/cluster.tf
@@ -341,22 +341,14 @@ resource "google_container_node_pool" "pools" {
     image_type   = lookup(var.node_pools[count.index], "image_type", "COS")
     machine_type = lookup(var.node_pools[count.index], "machine_type", "n1-standard-2")
     labels = merge(
-      {
-        "cluster_name" = var.name
-      },
-      {
-        "node_pool" = var.node_pools[count.index]["name"]
-      },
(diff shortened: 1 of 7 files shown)


kevang:
Thanks @morgante, I updated it.

kevang:
I reformatted and the linting tests all pass, yet I still get `Error: The following tests have failed: check_documentation`. I tried `make docker_generate_docs` but it doesn't seem to make any changes, not sure if there's anything else I'm missing

morgante:
@kevang Looks like the linters are passing now, just waiting for integration tests to pass before merging.

kevang:
@morgante do you think that it's ok to merge now that the checks have passed? 🙂

PR #283, opened 2019-10-17
Release Channels

chrissng:


morgante:
@chrissng Please make sure tests pass.

chrissng:
@morgante I've made the necessary changes to pass the tests. Please review my PR.

autogen/variables.tf
```diff
@@ -422,4 +422,15 @@ variable "authenticator_security_group" {
   default     = null
 }
 
+variable "enable_release_channel" {
```
morgante: Do we need a separate `ENABLE` variable here? I'm wondering if the default for `release_channel` could be to simply not enable it at all?

chrissng: Yes, you are right. Since release channel is defaulted to `UNSPECIFIED`, then it is assumed that release channels are not used.

Removed `enable_release_channel`

morgante: I think we should have a `null` default and omit the entire block in that case.

chrissng: Done

PR #284, opened 2019-10-17
[wait-for-cluster.sh] Set CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE.

omazin:
This is due to the fact that GCP client libraries and gcloud CLI automatically check for different environment variables - `GOOGLE_APPLICATION_CREDENTIALS` and `CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE` respectively.

PR #285, opened 2019-10-18
[wait-for-cluster.sh] Update the script in all submodules.

omazin:
Follows up #284


paulpalamarchuk:
@omazin please update sh-script in autogen directory (`autogen/scripts/wait-for-cluster.sh`)

omazin:
@paulpalamarchuk updated

omazin:
Updated, without the fix it would throw the following error when `terraform create / destroy`:
```
module.gke.null_resource.wait_for_cluster (local-exec): ERROR: (gcloud.container.clusters.list) ResponseError: code=403, message=Required "container.clusters.list" permission(s) for "projects/your-project-id".
```

aaron-lane:
Merging since the test timeout will be addressed with #292 and this solves a regression in the generation diff.

PR #286, opened 2019-10-18
Add test for sandbox

### LINKED ISSUE #240: Add sandbox flag for node-pools

### LINKED ISSUE #252: Automatically update labels for sandboxed node pools
This is a follow up to #240.

Based on https://github.com/terraform-providers/terraform-provider-google/issues/4210, if sandboxing is enabled then the control plain will add this label to the cluster automatically: `sandbox.gke.io/runtime" = "gvisor`.

This causes a permadiff. We should fix this by automatically injecting that label into the config if the sandbox flag is `true` on the module.

paulpalamarchuk:
* Moved `sanbox_config` to `node_pool` resouse (Fix #240)
 * Created test for sandbox (Fix #252)

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -131,3 +131,10 @@ suites:
       systems:
         - name: workload_metadata_config
           backend: local
+  - name: "sandbox_enabled"
+    driver:
+      root_module_directory: test/fixtures/sandbox_enabled
+    verifier:
+      systems:
+        - name: sandbox_enabled
+          backend: local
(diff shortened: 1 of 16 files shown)


.kitchen.yml
```diff
@@ -131,3 +131,10 @@ suites:
       systems:
         - name: workload_metadata_config
           backend: local
+  - name: "sandbox_enabled"
```
morgante: Please add this test to the [cloud build config](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/build/int.cloudbuild.yaml)?

paulpalamarchuk: Test added

PR #292, opened 2019-10-22
Parallelize the builds

aaron-lane:
Running all Kitchen instances in serial is taking over 3 hours to complete which is an extreme amount of time to wait for validation of changes.


aaron-lane:
Looks like the build was reduced to 33 minutes!

morgante:
@aaron-lane Looks like tests are failing though.

bharathkkb:
@aaron-lane is it possible to template int.cloudbuild.yaml and generate it automatically based on whats in test/ci? I thought that was the use of test/ci maybe I am missing something.

morgante:
@bharathkkb We could probably write a script to do that, but it's not a high priority right now.

`test/ci` stores the actual test definitions and config, the Cloud Build config is just calling them.

PR #301, opened 2019-10-31
Tying identity namespace output to successful cluster creation.

kevensen:
When configuring workload identity, there is the possibility that when creating an IAM binding, Terraform attempts to create the binding, before the workload identity is configured by GKE.  This PR creates a tighter dependency between the creation of a GKE cluster and the availability of the workload identity.

This is the first step in addressing https://github.com/forseti-security/terraform-google-forseti/issues/298.


kevensen:
@morgante Thanks for the review!

PR #303, opened 2019-11-01
Add node_locations for node pools

### LINKED ISSUE #290: Allow setting node locations on individual node pools
As added [upstream](https://github.com/GoogleCloudPlatform/magic-modules/pull/2320), we need to update this module to support specifying `node_locations` on individual node pools (that potentially differ from the cluster's overall locations).

This will unblock #263.

bharathkkb:
fixes #290 

Add support for node_locations to set locations for individual node pools.


DanielWozniak:
Don't mean to rush, but was wondering what is the status on this? When could we expect this to be merged?

morgante:
@DanielWozniak I think we're waiting for tests to pass. @bharathkkb Could you take a look?

bharathkkb:
@morgante I'll try re running the tests. It was failing destroy stage of a different example.

DanielWozniak:
Looks like the checks are passing, could one of the approvers please take a look?

DanielWozniak:
Thanks guys, appreciate you putting this in.

Issue #308, opened 2019-11-06
Provide example of configuring Private Google Access


morgante:
We should add an equivalent of the [public networking example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples/simple_regional_with_networking) with a private networking example which includes:
1. Setting up the required VPC subnets
2. Setting up private Google Access
3. Creating the actual GKE cluster
4. Any required firewall rules

This would help to address issues like [this one](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/305#issuecomment-550150921).

ideasculptor:
I believe this is still a bug, not an enhancement.  I have had private google access enabled through all of my tests (which explains why I failed to note the issue, since I had certainly read every word of the page you linked numerous times before I started documenting the issue), and node pools have consistently failed to come up.  When a node pool is not private, private google access has no effect (according to the google docs), and the nodes, with the default configuration, are unable to download necessary resources from storage.googleapis.com over port 443.  It is possible that the failure is something different when I run a completely private node pool, but since I can't get that configuration to come up at all, I have limited ability to diagnose the problem.  When I create publicly accessible nodes, they are unable to fetch the url until I explicitly add rules and routes to enable access via the default internet gateway.  

I can well believe that the nodes need to use a different hostname in order to resolve to the private google access endpoint, but there is no mention of that in the documentation for the module, nor even an obvious way to override it.

morgante:
> I believe this is still a bug, not an enhancement. 

It's not a bug. Nothing in our module is meant to configure Private Google Access for you, and we explicitly *don't* attempt to control networking from the module.

The fact that your chosen network config didn't have a workable path to GCP APIs isn't the scope of this module.

I think you're overfocusing on this module. Private Google Access isn't specific to GKE and if you'd attempted to use GCE on that network it'd also have similar issues with accessing any GCP samples.

I know you spent a lot of time troubleshooting this, but we really can't cover everything. If you had attempted to manually create the same GKE cluster through the console, you'd have had the same issue.

Anyways, the private access docs to cover the requirements for setting it up: https://cloud.google.com/vpc/docs/configure-private-google-access#requirements

We can add examples but to claim this is a bug when it's an issue with your underlying network config is inaccurate. Reminder: most customers we work with explicitly separate network creation/management from the GKE clusters themselves.

ideasculptor:
[...]

Issue #309, opened 2019-11-06
module region output can be blank, breaking tests


ideasculptor:
Looks like the `region` output of the beta_private_cluster module (computed from a local.cluster_output_region, assigned as `cluster_output_region = google_container_cluster.primary.region`) can be null.  This potentially breaks tests.  

The case that I exercised was as follows:

```
regional = false
region = 'us-east4'
zones = ['us-east4-a']
```

According to the documentation of the google_container_cluster resource, the region attribute is deprecated, and will only have a value if the cluster is regional instead of zonal.  The module should be computing the region from the location (zone), in a zonal cluster, if it is going to output a region - since tests for the with_network examples require a region.  Location will contain the region in a regional cluster.

I fixed the private_zonal_with_network example by using the incoming variable as the output value instead of retrieving it from the resource attribute.

PR #310, opened 2019-11-06
Private zonal example

ideasculptor:
Creates an example that demonstrates creating a network and a private, zonal GKE cluster. 

Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/308

Exposed  https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/309

### DIFF
--- a/examples/private_zonal_with_networking/main.tf
+++ b/examples/private_zonal_with_networking/main.tf
@@ -22,9 +22,9 @@ module "gcp-network" {
 
   subnets = [
     {
-      subnet_name   = var.subnetwork
-      subnet_ip     = "10.0.0.0/17"
-      subnet_region = var.region
+      subnet_name           = var.subnetwork
+      subnet_ip             = "10.0.0.0/17"
+      subnet_region         = var.region
       subnet_private_access = "true"
     },
   ]
@@ -44,24 +44,24 @@ module "gcp-network" {
 }
 
 data "google_compute_subnetwork" "subnetwork" {
-  name    = var.subnetwork
-  project = var.project_id
-  region  = var.region
+  name       = var.subnetwork
+  project    = var.project_id
+  region     = var.region
   depends_on = [module.gcp-network]
 }
 
 module "gke" {
-  source                  = "../../modules/beta-private-cluster/"
-  project_id              = var.project_id
-  name                    = var.cluster_name
-  regional                = false
-  region                  = var.region
-  zones                   = slice(var.zones, 0, 1)
+  source     = "../../modules/beta-private-cluster/"
+  project_id = var.project_id
+  name       = var.cluster_name
+  regional   = false
+  region     = var.region
+  zones      = slice(var.zones, 0, 1)
 
   // This craziness gets a plain network name from the reference link which is the
   // only way to force cluster creation to wait on network creation without a
   // depends_on link.  Tests use terraform 0.12.6, which does not have regex or regexall
-  network                 = reverse(split("/", data.google_compute_subnetwork.subnetwork.network))[0]
+  network = reverse(split("/", data.google_compute_subnetwork.subnetwork.network))[0]
 
   subnetwork              = data.google_compute_subnetwork.subnetwork.name
   ip_range_pods           = var.ip_range_pods_name
--- a/examples/private_zonal_with_networking/main.tf
+++ b/examples/private_zonal_with_networking/main.tf
@@ -44,7 +44,7 @@ module "gcp-network" {
 }
 
 data "google_compute_subnetwork" "subnetwork" {
-  name    = module.gcp-network.subnets_names[0]
+  name    = var.subnetwork
   project = var.project_id
   region  = var.region
   depends_on = [module.gcp-network]
@@ -57,7 +57,12 @@ module "gke" {
   regional                = false
   region                  = var.region
   zones                   = slice(var.zones, 0, 1)
-  network                 = data.google_compute_subnetwork.subnetwork.network
+
+  // This craziness gets a plain network name from the reference link which is the
+  // only way to force cluster creation to wait on network creation without a
+  // depends_on link.  Tests use terraform 0.12.6, which does not have regex or regexall
+  network                 = reverse(split("/", data.google_compute_subnetwork.subnetwork.network))[0]
+
   subnetwork              = data.google_compute_subnetwork.subnetwork.name
   ip_range_pods           = var.ip_range_pods_name
   ip_range_services       = var.ip_range_services_name
--- a/examples/private_zonal_with_networking/main.tf
+++ b/examples/private_zonal_with_networking/main.tf
@@ -1,0 +1,82 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module "gcp-network" {
+  source       = "terraform-google-modules/network/google"
+  version      = "~> 1.4.0"
+  project_id   = var.project_id
+  network_name = var.network
+
+  subnets = [
+    {
+      subnet_name   = var.subnetwork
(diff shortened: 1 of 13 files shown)


ideasculptor:
When running the new test manually, it became clear that the destroy process is not corectly reversing the dependency chain between the network module and the gke module, resulting in the subnet deletes happening before the cluster is fully torn down.  This causes destroy to fail.  I don't have handy example code for setting up a manual delay via a provisioner or some other mechanism to correctly reverse those dependencies during destroy, so this will probably fail running tests, but it is otherwise correct when I run it locally.

examples/private_zonal_with_networking/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  regional   = false
+  region     = var.region
+  zones      = slice(var.zones, 0, 1)
+
+  // This craziness gets a plain network name from the reference link which is the
+  // only way to force cluster creation to wait on network creation without a
+  // depends_on link.  Tests use terraform 0.12.6, which does not have regex or regexall
+  network = reverse(split("/", data.google_compute_subnetwork.subnetwork.network))[0]
+
+  subnetwork              = data.google_compute_subnetwork.subnetwork.name
```
morgante: Can we make this a reference to the `gcp-network` outputs? That might help with the dependency graph issue.

ideasculptor: I don't think so.  When I used a direct dependency, it would always fail the first time, because the cluster would come up before the network.  Those lines were written that way specifically to eliminate that problem.  I had to try a number of ways to string those dependencies together to get it to stop erroring out when creating both the network and the cluster in the same terraform operation - the problem doesn't exist when they are run separately.

I can't remember which variations my 3am brain attempted - every failed attempt is a really long test cycle, so it's entirely possible that I was addled enough from lack of sleep to have missed something obvious.  If memory serves, if I used the constant cidr range as a string and just injected it into ipv4_master_networks_config, that caused dependencies to fire before the network was ready.  I'm reasonably certain my first solution would have been to grab the first value of subnets_ranges from the gcp-network module outputs, but maybe I went straight from constant value to complex hack without first trying all the obvious middle steps, though I certainly tried a few.  

Because the network name is passed to the gcp-network outputs directly from an input value, dependencies in the gke module were allowed to fire with the network name before the network existed, since terraform's dependency graph wouldn't wait for other module to complete.  You can't use the depends_on thing with modules, so I had no way to make the dependency work correctly without modifying one or both modules to manually provide a depends_on link between resources in each module and I've just spent far too much time and energy on this already.  

My quick solution was to use the data source as an intermediary, but since the gke module can't accept the network name as a fully qualified link, I had to parse the network name out of that data source's attributes.  That corrected the dependency ordering on creation, but breaks the destroy ordering, since the data source doesn't act as a barrier to the network module being torn down, so it gets torn down before the GKE cluster has gone away.

morgante: Network name should come from the network outputs though, not directly passing through the input: https://github.com/terraform-google-modules/terraform-google-network/blob/master/main.tf#L19

ideasculptor: This speaks to the lack of real-world integration examples for the modules.  On the one hand, real-world production usage will almost never bring up the network and the GKE cluster in the same terraform plan, but on the other, integration examples always will, so if those examples existed, the modules would already have handled the outputs being directly linked to inputs and failing to result in a valid dependency graph, and someone other than a user would have spotted the need for some form of computed pass-through or provided a null_resource (and corresponding consumer in other modules), which would serve as a block.

The pattern I've used in the past is for every module to include a null_resource.completion resource with an output that depends on it and a set of triggers that includes all computed resources in the dependency chain. Additionally, every module can receive an input which is used only as a trigger for a different null_resource.block resource, with every other resource in the module including a `depends_on = [null_resource.block]`. Feeding the null_resource.completion-based output  of the first module to the trigger of the null_resource.block resource in the dependent module guarantees that nothing in the dependent module can execute until everything in the dependency has completed.  If no value is passed through, no blockage happens.

First module:
```
resource "null_resource" "continuation" {
  depends_on = [google_compute_resource1.name, google_compute_resource2.name, ...]
}

output "block_until_ready" {
  value = "${null_resource.continuation.id}"
}
```

Dependent Module:
```
resource "null_resource" "block_until_ready" {
  triggers {
    # Explicitly wait on the passed in block_until_ready variable
    block_until_ready = "${var.block_until_ready}"
  }
}

resource "google_some_resource" "my_dependent_resource" {
  some_val = var.some_val
...
  depends_on = ["null_resource.block_until_ready"]
}
```

morgante: I'm still not following the issue though. The linked code shows that the network output isn't directly linked to the input.

ideasculptor: Your constant assumption that I'm an idiot who doesn't know what he is doing, rather than someone with a great deal of experience with both terraform and infrastructure automation, encountering actual real-world problems that your simple examples and mimimal documentation just don't account for, has burned through pretty much all of the goodwill that was allowing me to spend so much of my **spare** time improving your product for free.  Honestly, is there ANYTHING in any of my significant contributions so far that makes you think I don't read documentation before asking for assistance? Maybe don't assume that to be the case every time I raise an issue that is more significant than a typo.

I don't have the big picture of where you are going with those modules, but that's only because it hasn't been laid out anywhere public where I can see it, so far as I am aware.  If you put modules out there, publicly acknowledging that they are built in-house by Google, you have to assume people are going to try to make use of them, as they surely represent something akin to common best practices.  If you don't provide meaningful guidance about how to do use them as such, and also refuse to respond helpfully when people encounter problems, you are going to burn down whatever community your open source efforts are trying to build.  

Honestly, in the immediate wake of a nearly 72 hour unscheduled service outage, maybe consider the kind of frustrations I might be experiencing when I still can't get things to work, and take a few minutes from your obviously very busy schedule to give an assist to someone who has shown no hesitation to devote significant time and energy into giving your team a helping hand?...

ideasculptor: An hour or less of a single engineer's time doesn't seem like a huge sacrifice to make in exchange for the 100 or so hours that I have devoted just to correcting problems in these modules - the vast majority of which had nothing whatsoever to do with any kind of non-standard or mistaken use.  

And maybe consider that if someone who is clearly pretty darn fluent in this stack is having difficulty intuiting your intent, someone with minimal terraform or GCP experience, looking to get a leg up on their own automation efforts, is going to have a **much** more difficult time of it, and that maybe some documentation or assistance is in order, rather than public shaming.

morgante: > Your constant assumption that I'm an idiot who doesn't know what he is doing, rather than someone with a great deal of experience with both terraform and infrastructure automation, encountering actual real-world problems that your simple examples and mimimal documentation just don't account for, has burned through pretty much all of the goodwill that was allowing me to spend so much of my spare time improving your product for free. Honestly, is there ANYTHING in any of my significant contributions so far that makes you think I don't read documentation before asking for assistance? Maybe don't assume that to be the case every time I raise an issue that is more significant than a typo.

I'm sorry you feel that way — I certainly think you're very skilled and appreciate all the work you've been doing. It's hugely helpful!

> I don't have the big picture of where you are going with those modules, but that's only because it hasn't been laid out anywhere public where I can see it, so far as I am aware. If you put modules out there, publicly acknowledging that they are built in-house by Google, you have to assume people are going to try to make use of them, as they surely represent something akin to common best practices. If you don't provide meaningful guidance about how to do them as such, and also refuse to respond helpfully when people encounter problems, you are going to burn down whatever community your open source efforts are trying to build.

Agreed, I know we're not doing a good enough job here. In terms of our goals, [this page](https://cloud.google.com/foundation-toolkit/) is probably the best we have.

We absolutely do need to provide more and better end-to-end examples, and it's on our radar for what to build next. Sorry if my responses are too brusque thus far, I definitely recognize that need.

> Honestly, in the immediate wake of a nearly 72 hour unscheduled service outage, maybe consider the kind of frustrations I might be experiencing when I still can't get things to work, and take a few minutes from your obviously very busy schedule to give an assist to someone who has shown no hesitation to devote significant time and energy into giving your team a helping hand?...

You're right: as you can imagine, there have been a lot of fires lately but we should still appreciate our contributors.

ideasculptor: I can promise you, when I used a direct link between the one module and the other, it errored out on the first run (and succeeded on the 2nd, if no destroy happened first) every single time. Feel free to try it.  If I'm wrong, and it was certainly hour 17 or 18 of my work day at that point, it's a simple enough change to make.  I've contributed my last PR for a good long while, at this point.

morgante: Ok, I'll give it a shot! It's just surprising that this doesn't work consider how it's coded but I certainly believe you've tried it.

PR #322, opened 2019-11-14
New private cluster example with multiple node pools and IAP bastion host access pattern

wripley:
Created from https://github.com/GoogleCloudPlatform/gke-terraform-generator/blob/master/examples/example.yaml

Enables recommended security features and demonstrates a realistic private IP access pattern for a private cluster.


ideasculptor:
Thanks!  This is a super helpful example.

examples/zonal_private_iap_bastion/outputs.tf
```diff
@@ -0,0 +1,132 @@
...
+}
+
+output "get_credentials" {
+  description = "Gcloud get-credentials command"
+  value       = format("gcloud container clusters get-credentials --project %s --zone %s --internal-ip %s", var.project_id, module.gke.location, var.cluster_name)
+}
+
+output "bastion_ssh" {
+  description = "Gcloud compute ssh to the bastion host command"
+  value       = format("gcloud beta compute ssh %s --tunnel-through-iap --project %s --zone %s -- -L8888:127.0.0.1:8888", module.bastion.hostname, var.project_id, local.bastion_zone)
```
ideasculptor: Tunnel shouldn't be to 127.0.0.1:8888, which would connect to the bastion host.  It should be to the private endpoint of the gke cluster

-L8888:${module.gke.endpoint}:8888 or -L8888:%s:8888 and add module.gke.endpoint to the format parameter list

That seems to fix it, but I have to fix my firewall to allow connections on port 8888, first, I think.  Mine is just hanging when trying to connect across that tunnel - but it used to error out with the tunnel to 127.0.0.1:8888

wripley: Bastion is running TinyProxy, so the IAP tunnel is allowing us to proxy 8888 from localhost through the bastion and finally to the cluster master, which allows us to run kubectl commands through. What was the error when it was set to 127.0.0.1:8888?

ideasculptor: Failure to connect. 

But I’ll check it again when I get back to work on it later. I have tiny proxy installed on mine, too, but i didn’t dig into the logs and make sure the startup script ran correctly and the proxy started up. I didn’t grok that tinyproxy was intended to be receiving that connection, so I didn’t even try to debug the failed connection. I just switched it to the endpoint address.

I’m curious, what’s the purpose of using tiny proxy instead of setting up the tunnel to connect directly to the endpoint? Just the convenience of being able to connect to any internal address and port from that one tunnel, instead of one tunnel for each service being connected to? Or is there more to it?

ideasculptor: Confirmed - tinyproxy didn’t install due to lack of outbound internet access. User error

wripley: The intent was to lock down access to the cluster to be all private. The firewall is only configured to allow certain connections in from the bastion host, and only resources inside the VPC are configured in the master authorized network list for the cluster.

ideasculptor: I understand that, but if the bastion host can get to the k8s cluster, then `-L8888:${private_endpoint}:8888` works in the exact same way, just without the extra hop through a proxy server on the bastion.  In either case, port 8888 on the ssh client host connects to the bastion and then connects to the specified destination address on the destination port from there.  Any DNS lookup, if a hostname is used instead of ip address, also occurs on the bastion, not on the ssh client host, so you can use internal DNS names there.

wripley: I ran into a similar issue when I was migrating the system over to use the bastion host module. TinyProxy did not install due to the image being set to centos instead of debian, and I did not realize the startup script did not run. I agree that documentation is a good idea, especially for those who cannot pull down tiny proxy. I will add it and resolve this conversation once I make the updates.

ideasculptor: Success! I finally got everything to work.  I have the fabric firewall module in play (with defaults, so traffic is fairly well locked down), and I had to muck around with firewall rules and routes a bit, but everything is now functional.  

This example was super helpful, too.  It probably would have taken me several more days to get it all put together.

Changes I made to make it work on a network with default firewall rules:
- I had to make sure the subnet my bastion is in was added to the authorized networks, which also sets up the route
- I added a firewall rule which allows the bastion ip range to target all ports and protocols in the gke service account, but I suspect I can also make it use bastion source tag and gke target tag - I just haven't tested it yet.  The second form is more flexible.

These are the changes I made to get the bastion to startup with the proxy correctly, but a network configured differently would probably get access to the public internet differently:
- I had to add tags to the bastion host (PR already merged to support that), in order to create a route for the bastion out to the public internet, which fixed the proxy install (in combination with a firewall rule that allows the traffic).
- I added a firewall rule to allow outbound https traffic to 0.0.0.0/0 from a bastion tag (for fetching packages)

I think everything else just worked as-is when copied from your example.

morgante: @wripley I wonder if it would be helpful if we included explicit firewall rules in this example? Even though we're spinning up a new network (which will by default allow all egress), it might be helpful to show what you need to allow in a more locked down network.

wripley: We could show an egress firewall rule being allowed for the bastion host.

ideasculptor: I haven’t terraformed the routes and rules I created manually to get it all working yet, but I will do so in the next 24 hours. I’ll paste something here once I’ve got them - save you most of the work.

examples/zonal_private_iap_bastion/outputs.tf
```diff
@@ -0,0 +1,132 @@
...
+}
+
+output "bastion_kubectl" {
+  description = "kubectl command using the local proxy once the bastion_ssh command is running"
+  value       = "HTTPS_PROXY=localhost:8888 kubectl get pods --all-namespaces"
```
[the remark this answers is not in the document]
ideasculptor: https://github.com/GoogleContainerTools/skaffold/issues/3254

That is pretty much the problem.  I want to use skaffold to deploy stuff, but it runs both gcloud and kubectl commands, and one needs http proxy and the other must not use it.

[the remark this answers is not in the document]
ideasculptor: In truth, the generic http proxy method for accessing a private cluster through a bastion doesn't seem to really work all that well.  At least on OS X, the system-wide proxy settings also have only a no-proxy setting, rather than an only-proxy setting, so it is going to be an exercise in constantly fighting against that proxy configuration just for the endpoint addreess until I do something drastic like run a configurable proxy locally, which only forwards certain addresses to the ssh tunnel, or I figure out how to get an actual VPN connection to my VPC without having to buy a VPN-capable router.  There has to be some kind of software (preferably open-source software) solution to the problem of connecting to a VPC network from a single host in a coffee shop or airport.

Th other alternative is to set up an ssh tunnel directly to the private endpoint, and then configure kubectl to access the server at 127.0.0.1, and let the ssh tunnel translate that to the correct host.  But that can break things, too, if kubectl ever returns values that include the endpoint address without using the value received in the Host header.

examples/zonal_private_iap_bastion/outputs.tf
ideasculptor: Perhaps consider adding `-A` to the ssh args, which enables ssh agent forwarding, in order to be able to bounce through the bastion to node_pool instances.

examples/zonal_private_iap_bastion/network.tf
```diff
@@ -0,0 +1,51 @@
...
+}
+
+module "cloud-nat" {
+  source        = "terraform-google-modules/cloud-nat/google"
+  project_id    = var.project_id
+  region        = var.region
+  create_router = "true"
+  router        = "${var.cluster_name}-cloud-nat"
+  network       = module.gke-network.network_name
+}
```
ideasculptor: I found I had to add an explicit route for destination 0.0.0.0/0 in order to take advantage of the NAT instance. I didn't completely tear down the host project, however, so maybe the original firewall configuration had left some configuration changes behind?  I did destroy all the subnets and recreate them, this time skipping the creation of any firewall rules, and the bastion still couldn't access the public internet.  Nor could it get to nodes in my gke subnet.  The routes existed, but it needed a firewall rule granting ingress permission.  Examples provided below.

```
resource "google_compute_route" "bastion-egress" {
  project          = var.project_id
  network          = module.gke-network.network_name
  name             = "bastion-egress"
  description      = "Allow bastion to talk to the internet"

  # latest bastion module allows setting of tags.  Not released, but merged.  
  # Otherwise use service_account?
  tags             = var.tags

  dest_range       = "0.0.0.0/0"
  next_hop_gateway = "default-internet-gateway"
}

resource "google_compute_firewall" "bastion-ingress" {
  # required for 'enable-logging'
  provider       = "google-beta"
  name           = "ingress-bastion"
  description    = "Allow inbound traffic from bastion host"
  network        = module.gke-network.network_name
  project        = var.project_id
  direction      = "INGRESS"
  enable_logging = true

  # use service account if you don't have tags
  source_tags    = ["bastion"]

  dynamic "allow" {
    for_each = ["tcp", "udp", "icmp"]
    content {
      protocol = allow.value
    }
  }
}
```

ideasculptor: Another data point.  When I added a bunch of services to the cluster, requiring vertical auto_scaling, the node pool size was bumped up to 4 (from 2), and the new instances launched, but until I explicitly added a route to the public internet for nodes in the cluster, the new nodes were not detected by the kubernetes cluster and pods remained unschedulable.  As soon as I added a route to 0.0.0.0./0 for gke nodes, the additional nodes were detected and pods were scheduled. 

I had such a route when launching the cluster, but eventually swapped it for one which only allowed the bastion to go out to 0.0.0.0/0.  Then I tried to scale the cluster up, which didn't work until I added the route back.  I suspect that this implies that private_access requires more configuration than simply setting subnet_private_access to true in the network module, though I don't know what else is required.  The docs sure seem to imply that it should just work unless you've overwritten the default route (https://cloud.google.com/vpc/docs/configure-private-google-access) so maybe this is another artifact of my prior use of the firewall module without destroying the host project and recreating it, but I did destroy all subnets and recreate them.

My guess would be that it wants me to configure DNS in the VPC with a cname from *.googleapis.com to restricted.googleapis.com in order to make this work without explicitly routing to 0.0.0.0/0 for any gke nodes (in order to go through NAT, but that implies it isn't using the private access without such a cname).

wripley: It looks like I forgot to add the subnets from the original example. Will add and test.

bharathkkb:
Hi @wripley 
Would you be interested in finishing up this PR possibly with the `safer-cluster` module or I can help in completing this? This would be a great addition :)

bharathkkb:
Replaced by #544

PR #323, opened 2019-11-14
Add more detailed documentation for private cluster connectivity

morgante:
This makes it less ambiguous how you should connect to private clusters for Terraform.

We also make sure to explicitly set the cluster endpoint output to the `public_endpoint` if `deploy_using_private_endpoint` is false.

### DIFF
--- a/autogen/README.md
+++ b/autogen/README.md
@@ -12,9 +12,18 @@ The resources/services/activations/deletions that this module will create/trigge
 Sub modules are provided from creating private clusters, beta private clusters, and beta public clusters as well.  Beta sub modules allow for the use of various GKE beta features. See the modules directory for the various sub modules.
 
 {% if private_cluster %}
-**Note**: You must run Terraform from a VM on the same VPC as your cluster, otherwise there will be issues connecting to the GKE master.
+## Private Cluster Endpoints
+When creating a [private cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters), nodes are provisioned with private IPs.
+The Kubernetes master endpoint is also [locked down](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#access_to_the_cluster_endpoints), which affects these module features:
+- `configure_ip_masq`
+- `stub_domains`
 
-  {% endif %}
+If you are *not* using these features, then the module will function normally for private clusters and no special configuration is needed.
+If you are using these features with a private cluster, you will need to either:
+1. Run Terraform from a VM on the same VPC as your cluster (allowing it to connect to the private endpoint) and set `deploy_using_private_endpoint` to `true`.
+2. Include the external IP of your Terraform deployer in the `master_authorized_networks_config`.
+
+{% endif %}
 
 ## Compatibility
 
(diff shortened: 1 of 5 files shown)


autogen/README.md
```diff
@@ -12,9 +12,18 @@ The resources/services/activations/deletions that this module will create/trigge
...
+When creating a [private cluster](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters), nodes are provisioned with private IPs.
+The Kubernetes master endpoint is also [locked down](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#access_to_the_cluster_endpoints), which affects these module features:
+- `configure_ip_masq`
+- `stub_domains`
 
-  {% endif %}
+If you are *not* using these features, then the module will function normally for private clusters and no special configuration is needed.
+If you are using these features with a private cluster, you will need to either:
+1. Run Terraform from a VM on the same VPC as your cluster (allowing it to connect to the private endpoint) and set `deploy_using_private_endpoint` to `true`.
+2. Include the external IP of your Terraform deployer in the `master_authorized_networks_config`.
```
ideasculptor: Having tried this, it is probably worth pointing out that you can't just stick the ip address of your local host with a /32 netmask in master_authorized_networks_config.  It must be a 'reserved range' (I believe that was the terminology used in the error message).  Perhaps it is possible (I haven't looked into it) to add an arbitrary address/subnet as a reserved range, but you will have to do that in the network configuration (I assume) before it can be added to master_authorized_networks_config.

It is definitely helpful to call out which features do require connecting to the cluster as part of deployment.  

Given that the new private cluster with network example PR (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/322 ) includes commands for connecting kubectl via a bastion tunnel in the outputs, perhaps a provisioner could be included which makes that connection and then enables those features that way, rather than forcing an entry in master_authorized_networks_config and/or running the deployment from a host on the VPC?  No need to block merging this on that request (I'd like that fix for the public endpoint output sooner rather than later), but maybe update the documentation and then consider it for a future enhancement?

bharathkkb: In the case of #322 it is still a VM on the same VPC as the cluster and master_authorized_networks_config set up to allow access from the bastion right?

ideasculptor: a `kubectl_https_proxy = "https://localhost:8888"` variable, and mods to the kubect calls which would cause kubectl to proxy through the bastion tunnel, would get the job done.  There is already an example (in a PR) which sets up a bastion and provides the ssh command necessary to establish the tunnel.  Simply adding one of these features to that example, and ensuring the example executes correctly via the tunnel would keep the doucmentation requirements to a minimum, since you could just make brief mention of it, along with a pointer to the example in question.  But just having that example to play with, even without the need to run kubectl during setup, should make it fairly clear how to get kubectl to run against a private cluster without being on the VPC network.  But it would be good if the various modules were built to accommodate that setup by making the calls correctly, with HTTPS_PROXY in the environment.

morgante: I'm a little confused here: you *should* be able to add an arbitrary external IP range to the master authorized networks config, was that not working somehow?

I agree that the bastion pattern is another alternative, but it seems that's best addressed through the forthcoming examples.

ideasculptor: Try it and see.  You definitely cannot do so.  You probably can list an outside address in a config somewhere, making it a qualified range in the VPC, ths allowing it to be added to the naster authorized networks config, but when I just stuck the public ip address of my home NAT in there, it rejects it as not being a valid range.

morgante: You are very right, and sorry for the confusion. This is honestly very surprising to me and not covered at all in our public documentation, adding a caveat.

PR #346, opened 2019-11-22
Add variable to allow a pre-generated SSH key to be passed to the ACM module

### LINKED ISSUE #329: Add var to allow a pre-generated SSH key to be passed to the ACM module
The current implementation of the ACM module only allows you to generate a new SSH key or not to create the git-secret at all. Ideally it should be possible to pass in an `ssh_key` variable as well, so you have the ability to pass in a pre-generated key. 

Something like this would be nice functionality

* Was `ssh_key` provided?
   * Yes - > Use it
   * No -> Is `create_ssh_key` set to `true`?
      * Yes -> Generate one
      * No -> Do nothing

paulpalamarchuk:
* Fix #329

### DIFF
--- a/modules/acm/main.tf
+++ b/modules/acm/main.tf
@@ -18,7 +18,7 @@ locals {
   cluster_endpoint       = "https://${var.cluster_endpoint}"
   token                  = data.google_client_config.default.access_token
   cluster_ca_certificate = data.google_container_cluster.primary.master_auth.0.cluster_ca_certificate
-  private_key            = var.create_ssh_key ? tls_private_key.git_creds[0].private_key_pem : ""
+  private_key            = var.create_ssh_key && var.ssh_auth_key == null ? tls_private_key.git_creds[0].private_key_pem : var.ssh_auth_key
   download_operator      = var.operator_path == null ? true : false
   operator_path          = local.download_operator ? "${path.module}/config-management-operator.yaml" : var.operator_path
 }
--- a/modules/acm/variables.tf
+++ b/modules/acm/variables.tf
@@ -62,6 +62,12 @@ variable "create_ssh_key" {
   default     = true
 }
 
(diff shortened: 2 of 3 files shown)


modules/acm/variables.tf
```diff
@@ -62,6 +62,12 @@ variable "create_ssh_key" {
   default     = true
 }
 
+variable "ssh_auth_key" {
+  description = "Key for Git authentication. Overrides 'create_ssh_key' variable."
+  type        = string
+  default     = ""
```
aaron-lane: We should use `null` for empty defaults when possible.

```suggestion
  default     = null
```

modules/acm/main.tf
```diff
@@ -18,7 +18,7 @@ locals {
   cluster_endpoint       = "https://${var.cluster_endpoint}"
   token                  = data.google_client_config.default.access_token
   cluster_ca_certificate = data.google_container_cluster.primary.master_auth.0.cluster_ca_certificate
-  private_key            = var.create_ssh_key ? tls_private_key.git_creds[0].private_key_pem : ""
+  private_key            = var.create_ssh_key && var.ssh_auth_key == "" ? tls_private_key.git_creds[0].private_key_pem : var.ssh_auth_key
```
aaron-lane:
```suggestion
  private_key            = var.create_ssh_key && var.ssh_auth_key == null ? tls_private_key.git_creds[0].private_key_pem : var.ssh_auth_key
```

modules/acm/variables.tf
paulpalamarchuk: Set to `""` because it's used in ["local-exec" command](https://github.com/paulpalamarchuk/terraform-google-kubernetes-engine/blob/add_var_to_pass_ssh-key_in_acm/modules/acm/main.tf#L75)

PR #362, opened 2019-12-02
Compute region output for zonal clusters

### LINKED ISSUE #309: module region output can be blank, breaking tests

morgante:
Fixes #309

### DIFF
--- a/autogen/main.tf.tmpl
+++ b/autogen/main.tf.tmpl
@@ -96,8 +96,6 @@ locals {
 {% endif %}
 
   cluster_output_name           = google_container_cluster.primary.name
-  cluster_output_location       = google_container_cluster.primary.location
-  cluster_output_region         = google_container_cluster.primary.region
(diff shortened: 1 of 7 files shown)


morgante:
@aaron-lane PTAL

Issue #294, opened 2019-10-23
Import cluster/nodepools fails after #247


AkhilGNair:
After the change made in #247 for region to become optional, import now fails on regional clusters.

```
terraform import -config=/some/backend module.gke.google_container_node_pool.pools[0] path/to/pool
module.gke.google_container_node_pool.pools[0]: Importing from ID "path/to/pool"...
module.gke.google_container_node_pool.pools[0]: Import prepared!
  Prepared google_container_node_pool for import
module.gke.google_container_node_pool.pools[0]: Refreshing state... [id=path/to/pool]

Error: Invalid index

  on .terraform/modules/gke/modules/beta-private-cluster/main.tf line 37, in locals:
  37:   region   = var.region == null ? join("-", slice(split("-", var.zones[0]), 0, 2)) : var.region
    |----------------
    | var.zones is empty list of string

The given key does not identify an element in this collection value.
```

In the resource definition, `region` is being passed in as below

```
module "gke" {
  ...
  source = "git@github.com:terraform-google-modules/terraform-google-kubernetes-engine.git//modules/beta-private-cluster?ref=master"
  ...
  region = var.region
  ...
}
```

@paulpalamarchuk FYI :)

julianvmodesto:
Even when providing `zones`, this issue persists.

Seems like this may be the upstream issue:
https://github.com/hashicorp/terraform/issues/22721#issuecomment-535914976

omazin:
Hi @AkhilGNair, can you try overriding the `region` variable as follows:
- `import -var 'region=us-east-2' ... `
- create a `terraform.tfvars` or `*.auto.tfvars` with the following contents `region = "us-east-2"` and run your import command
- create a file named `region.tfvars` with the same contents and run `import -var-file="region.tfvars"`

If in all cases it still fails then it's indeed a terraform's bug.

omazin:
Was able to reproduce the error using the `simple_regional_cluster` example. The error log's below:
```
module.gke.google_container_cluster.primary: Importing from ID "flask-app-254610/europe-west3/simple-regional-private-clustertest123"...
module.gke.google_container_cluster.primary: Import prepared!
  Prepared google_container_cluster for import
module.gke.google_container_cluster.primary: Refreshing state... [id=simple-regional-private-clustertest123]

Error: Invalid index

  on ../../modules/private-cluster/main.tf line 37, in locals:
[...]

PR #376, opened 2019-12-09
Make setting 'region' more clear.

omazin:
Workaround for #294
But the real issue's with terraform not able to override a default value when importing resources is there.


omazin:
> Does this actually fix anything? It seems to be almost the exact same logic.

Yes it does make importing described in #294 possible due to the fact that `regional` variable is set to true by default. But of course it doesn't fix the underlying issue with terraform https://github.com/hashicorp/terraform/issues/22721.

morgante:
Got it, thanks.

PR #394, opened 2020-01-06
Enable Surge Upgrades (Beta)

Dev25:
Enable Surge upgrades for all beta node-pools, using the current default values set by GKE.

imo it should be a best practice/enforced by this module, but i can also add a disable flag if we want the option to disable it?

[GKE Doc](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#surge)
[GKE API Doc](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/UpgradeSettings)


Dev25:
That's reasonable, i suggest we go for 2 rather then adding a flag now and then defaulting that to yes in a future breaking change anyway. 

I assume we can get the for_each PR #257  merged soon and since that requires a state migration it would be a breaking change?

morgante:
Yes, #257 would also be a breaking change.

morgante:
@Dev25 Could you resolve conflicts? I'm hoping to release a new version this week (with a few breaking changes).

Dev25:
Awesome, done + updated the README node pool var section too @morgante

Issue #398, opened 2020-01-13
Workload Identity config per node pool


tuananhnguyen-ct:
I'm using beta-private-cluster and unable to disable workload identity per node pool, is there any related documentation or example on how to configure that? I can only use `node_metadata              = "GKE_METADATA_SERVER"` at cluster level to configure/enable for all node pools at once.

Dev25:
Existing issue https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/265

Since the for_each PR still hasn't been merged yet, i'll create another PR just with those the changes.

Issue #265, opened 2019-09-19
Allow overriding the node_metadata in node pools


Dev25:
It is not currently possible to do a selective rollout of the (beta) Workload Identity feature. If you set `var.node_metadata` then all node-pools will use that value which also forces a node pool recreation.

As a end user i want to set node_metadata with `GKE_METADATA_SERVER` on certain node-pools only.

The fix is fairly simple: https://github.com/Dev25/terraform-google-kubernetes-engine/commit/6f61744b50145412c7c027eda4f82d27be9fb123 but that is based on the for_each PR that has not yet been merged in #257

PR #400, opened 2020-01-13
Allow overriding node_metadata in node pools

Dev25:
Related Issues: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/265 https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/398

Allow users to create node pools with different node_metadata config.


Dev25:
Hey @morgante , looks like its already there as a required field for the beta module READMEs.

Although it is set to required, i'm going to set that to Optional

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/commit/60363d1c5ed8698a832bcce208a3abdc7aad40e3#diff-34a5f0b5ab84cf5f33367f7912231babR172


node_metadata | Options to expose the node metadata to the workload running on the node |   | Required
-- | -- | -- | --
```

PR #404, opened 2020-01-16
refactor: use gcloud module for scripts, closes #401

### LINKED ISSUE #401: gcloud executable requirement, removal of jq
Hello,

I'd like to discuss the possibility of using the [gcloud](https://github.com/terraform-google-modules/terraform-google-gcloud) module as a replacement for the `wait-for-cluster.sh` script. Perhaps a solution similar to this PR: https://github.com/terraform-google-modules/terraform-google-project-factory/pull/343.

I am a TFE customer and would like to avoid opting out of the "wait for cluster" feature via `skip_provisioners`. Thoughts? I'm thinking you could avoid requiring `jq` by doing something like this:

```
gcloud container clusters list --filter=name:<CLUSTER_NAME> --format="value(status)"
```

If this of interest I would be more than happy to put a PR together.

Thanks.

marshallford:
Uses gcloud module for wait-for-cluster and delete-default-resource scripts.

Fixes #401

In other words: Removes dependency on locally installed `glcloud`, `kubectl`, and `jq` binaries.

### DIFF
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -25,9 +25,8 @@ module "gcloud_delete_default_kube_dns_configmap" {
   enabled               = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners
   additional_components = ["kubectl"]
 
-  upgrade            = var.gcloud_upgrade
-  gcloud_sdk_version = var.gcloud_sdk_version
-  skip_download      = var.gcloud_skip_download
+  upgrade       = var.gcloud_upgrade
+  skip_download = var.gcloud_skip_download
 
   create_cmd_entrypoint = "${path.module}/scripts/kubectl_wrapper.sh"
   create_cmd_body       = "https://${local.cluster_endpoint} ${data.google_client_config.default.access_token} ${local.cluster_ca_certificate} ${path.module}/scripts/delete-default-resource.sh kube-system configmap kube-dns"
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -25,6 +25,10 @@ module "gcloud_delete_default_kube_dns_configmap" {
   enabled               = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners
   additional_components = ["kubectl"]
 
+  upgrade            = var.gcloud_upgrade
+  gcloud_sdk_version = var.gcloud_sdk_version
+  skip_download      = var.gcloud_skip_download
+
   create_cmd_entrypoint = "${path.module}/scripts/kubectl_wrapper.sh"
   create_cmd_body       = "https://${local.cluster_endpoint} ${data.google_client_config.default.access_token} ${local.cluster_ca_certificate} ${path.module}/scripts/delete-default-resource.sh kube-system configmap kube-dns"
 
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -21,7 +21,7 @@
  *****************************************/
 module "gcloud_delete_default_kube_dns_configmap" {
   source                = "terraform-google-modules/gcloud/google"
-  version               = "~> 0.3"
+  version               = "~> 0.4"
   enabled               = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners
   additional_components = ["kubectl"]
 
@@ -51,7 +51,7 @@ EOF
   }
 
   depends_on = [
-    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
+    module.gcloud_delete_default_kube_dns_configmap.wait,
     data.google_client_config.default,
     google_container_cluster.primary,
     google_container_node_pool.pools,
@@ -78,7 +78,7 @@ EOF
   }
 
   depends_on = [
-    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
+    module.gcloud_delete_default_kube_dns_configmap.wait,
     data.google_client_config.default,
     google_container_cluster.primary,
     google_container_node_pool.pools,
@@ -108,7 +108,7 @@ EOF
   }
 
   depends_on = [
-    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
+    module.gcloud_delete_default_kube_dns_configmap.wait,
     data.google_client_config.default,
     google_container_cluster.primary,
     google_container_node_pool.pools,
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -21,7 +21,7 @@
  *****************************************/
 module "gcloud_delete_default_kube_dns_configmap" {
   source                = "terraform-google-modules/gcloud/google"
-  version               = "~> 0.5"
+  version               = "~> 1.0.1"
   enabled               = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners
   additional_components = ["kubectl"]
 
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -21,7 +21,7 @@
  *****************************************/
 module "gcloud_delete_default_kube_dns_configmap" {
   source                = "terraform-google-modules/gcloud/google"
-  version               = "~> 0.4"
+  version               = "~> 0.5"
   enabled               = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners
   additional_components = ["kubectl"]
 
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -51,7 +51,7 @@ EOF
   }
 
   depends_on = [
-    module.gcloud_delete_default_kube_dns_configmap,
+    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
     data.google_client_config.default,
     google_container_cluster.primary,
     google_container_node_pool.pools,
@@ -78,7 +78,7 @@ EOF
   }
 
   depends_on = [
-    module.gcloud_delete_default_kube_dns_configmap,
+    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
     data.google_client_config.default,
     google_container_cluster.primary,
     google_container_node_pool.pools,
@@ -108,7 +108,7 @@ EOF
   }
 
   depends_on = [
-    module.gcloud_delete_default_kube_dns_configmap,
+    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
     data.google_client_config.default,
     google_container_cluster.primary,
     google_container_node_pool.pools,
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -27,6 +27,12 @@ module "gcloud_delete_default_kube_dns_configmap" {
 
   create_cmd_entrypoint = "${path.module}/scripts/kubectl_wrapper.sh"
   create_cmd_body       = "https://${local.cluster_endpoint} ${data.google_client_config.default.access_token} ${local.cluster_ca_certificate} ${path.module}/scripts/delete-default-resource.sh kube-system configmap kube-dns"
+
+  module_depends_on = concat(
+    [data.google_client_config.default.access_token],
+    [google_container_cluster.primary.master_version],
+    [for pool in google_container_node_pool.pools : pool.name]
+  )
 }
 
 /******************************************
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -19,18 +19,14 @@
 /******************************************
   Delete default kube-dns configmap
  *****************************************/
-resource "null_resource" "delete_default_kube_dns_configmap" {
-  count = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners ? 1 : 0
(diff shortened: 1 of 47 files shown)


modules/beta-private-cluster-update-variant/dns.tf
```diff
@@ -55,7 +51,7 @@ EOF
   }
 
   depends_on = [
-    null_resource.delete_default_kube_dns_configmap,
+    module.gcloud_delete_default_kube_dns_configmap,
```
morgante: Unfortunately it looks like tests are failing and I think the cause is Terraform not correctly traversing the dependency graph for modules. Instead I think we need to depend on the `wait` output.

```suggestion
    module.gcloud_delete_default_kube_dns_configmap.outputs.wait,
```

marshallford: Fixed.

marshallford: I ran `make docker_test_lint` locally and I don't think that works. Any ideas? I wish I could see the Cloud Build outputs...

```
Error: Invalid depends_on reference

  on ../../modules/beta-public-cluster/dns.tf line 111, in resource "kubernetes_config_map" "kube-dns-upstream-nameservers-and-stub-domains":
 111:     module.gcloud_delete_default_kube_dns_configmap.outputs.wait,

References in depends_on must be to a whole object (resource, etc), not to an
attribute of an object.
```

morgante: Hmm, you might have to incorporate it into an implicit dependency then. FYI Cloud Build is just running `make docker_test_integration` so you can verify locally.

morgante: `module.gcloud_delete_default_kube_dns_configmap.wait` might also work.

marshallford: Still need to brainstorm how to add an implicit dependency, but it doesn't look like depending on `.wait` will work.

```
Error: Reference to undeclared output value

  on ../../dns.tf line 81, in resource "kubernetes_config_map" "kube-dns-upstream-namservers":
  81:     module.gcloud_delete_default_kube_dns_configmap.wait,

An output value with the name "wait" has not been declared in
module.gke.module.gcloud_delete_default_kube_dns_configmap.
```

morgante: This example from the gcloud module might help - https://github.com/terraform-google-modules/terraform-google-gcloud/tree/master/examples/dependency_example

marshallford: Based on a little digging it appears that the `module_depends_on` "workaround" is only for when a module is depending on another module, not a resource depending on a module.

morgante: Yeah, I'm wondering if we could add an implicit dependency? Maybe via a bogus label on the resource?

marshallford: Would you accept a PR that does this:

```
labels = {
  maintained_by = "terraform"
  terraform_depends_on = module.gcloud_delete_default_kube_dns_configmap.outputs.wait
}
```

morgante: I think that's a reasonable direction, though it might cause noisy diffs. How about `length(module.gcloud_delete_default_kube_dns_configmap.outputs.wait) > 1 ? "done" : "undone"`

marshallford: Doesn't `wait >= 1` mean, yes wait some more?

morgante: No, the `wait` output is kind of just a dummy list of the inputs: https://github.com/terraform-google-modules/terraform-google-gcloud/blob/master/main.tf#L35

marshallford: Ok sure, should it be greater than zero though?

morgante: Yeah that might work too.

marshallford: Part of the issue might be that I have been testing this PR with version `0.3` of the gcloud module.

Could you please take a look at this issue: https://github.com/terraform-google-modules/terraform-google-gcloud/issues/19

marshallford: I believe I cleared up the dependency issues in both `cluster.tf` and `dns.tf`. I also resolved the merge conflict.

marshallford: The command `make docker_generate && make docker_generate_docs && make docker_test_lint` ran locally just fine.

marshallford:
I'm new to the `module_depends_on` concept but both elements in that list are of type "resource". Is that not specific enough? Should I instead depend on two string values returned by the resources instead?

morgante:
> I'm new to the module_depends_on concept but both elements in that list are of type "resource". Is that not specific enough? Should I instead depend on two string values returned by the resources?

Yeah I think they need to be the exact same type. Let's try making them two string values.

marshallford:
How does the usage of `for_each` in the resource `google_container_node_pool.pools` impact how I reference a returned value?

```
  module_depends_on = [
    google_container_cluster.primary.master_version,
    google_container_node_pool.pools.name, // not just one pool name?
  ]
```

morgante:
> How does the usage of for_each in the resource google_container_node_pool.pools impact how I reference a returned value?

You'll probably need to use splat syntax:
```
  module_depends_on = concat([google_container_cluster.primary.master_version], google_container_node_pool.pools.*.name)
```

morgante:
I'm getting this error now:
```
Step #35 - "converge stub-domains-local":        Error: Unsupported attribute
Step #35 - "converge stub-domains-local":        
Step #35 - "converge stub-domains-local":          on ../../../cluster.tf line 238, in module "gcloud_wait_for_cluster":
Step #35 - "converge stub-domains-local":         238:     join("-", google_container_node_pool.pools[*].name),
Step #35 - "converge stub-domains-local":        
Step #35 - "converge stub-domains-local":        This object does not have an attribute named "name".
```

marshallford:
`google_container_node_pool` does not export any values to depend on.

marshallford:
I'd like to keep this PR moving. @morgante, what's the current error? I'd love to wrap this up.

morgante:
Rerunning tests.

morgante:
Here's the error I'm seeing in CI, which makes me thing the dependency order isn't working properly:

```
Step #43 - "converge stub-domains-upstream-nameservers-local":        Error: configmaps "kube-dns" already exists
Step #43 - "converge stub-domains-upstream-nameservers-local":        
Step #43 - "converge stub-domains-upstream-nameservers-local":          on ../../../dns.tf line 88, in resource "kubernetes_config_map" "kube-dns-upstream-nameservers-and-stub-domains":
Step #43 - "converge stub-domains-upstream-nameservers-local":          88: resource "kubernetes_config_map" "kube-dns-upstream-nameservers-and-stub-domains" {
```

marshallford:
I recant my earlier statement, this recent PR clarifies that `google_container_node_pool` does export  attributes: https://github.com/terraform-providers/terraform-provider-google/pull/5477

morgante:
@marshallford FYI this is still failing. This is the error:
```
       Error: Invalid value for module argument
       
         on ../../../cluster.tf line 237, in module "gcloud_wait_for_cluster":
        237:   module_depends_on = [
        238:     google_container_cluster.primary.master_version,
        239:     google_container_node_pool.pools[*].name
        240:   ]
       
       The given value is not suitable for child module variable "module_depends_on"
       defined at
       .terraform/modules/example.gke.gcloud_wait_for_cluster/terraform-google-modules-terraform-google-gcloud-84e31b8/variables.tf:35,1-29:
       all list elements must have the same type.
       
       
       Error: Unsupported attribute
       
         on ../../../cluster.tf line 239, in module "gcloud_wait_for_cluster":
        239:     google_container_node_pool.pools[*].name
```

bharathkkb:
@marshallford it would be great to get this in. Anything we can do to help complete this?

marshallford:
@bharathkkb As you can see I fought quite a bit with TF trying to establish a dependency between `gcloud_wait_for_cluster` and `google_container_node_pool`. If you have any suggestions on using the helper input `module_depends_on` please let me know.

bharathkkb:
@marshallford I have a couple of ideas. Let me know if you have additional questions
i) I think the way to do the implicit dependency is 
```tf
module_depends_on = concat(
    [google_container_cluster.primary.master_version],
    [for pool in google_container_node_pool.pools: pool.name]
  )
``` 
This similar to what @morgante suggested but I don't believe you can use splat with for_each resources as it will output a map. So I have used the usual for expansion here.
ii) A few other things are we should expose `skip_download`  and possibly `upgrade` from the gcloud module so that users have that additional flexibility and imho `skip_download` should default to true. @morgante any thoughts on this?

bharathkkb:
@marshallford it looks like a bunch of unrelated `verify` steps are failing. I see a couple of `converge` steps passed, so it looks promising. Could you try a rebase so we have the latest tests from master?

bharathkkb:
@marshallford this is the test that failed and it was fixed here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/454
```
×  Command: `gcloud beta --project=ci-gke-10f1 container clusters --zone=us-central1 describe simple-regional-beta-cluster-1hne --format=json` cluster has the expected nodeMetadata conseal config
     
     expected: {"nodeMetadata"=>"EXPOSE"}
          got: {"mode"=>"GCE_METADATA", "nodeMetadata"=>"EXPOSE"}
     
     (compared using ==)
     
     Diff:
     @@ -1,2 +1,3 @@
     +"mode" => "GCE_METADATA",
      "nodeMetadata" => "EXPOSE",

     ✔  Command: `gcloud beta --project=ci-gke-10f1 container clusters --zone=us-central1 describe simple-regional-beta-cluster-1hne --format=json` cluster has the expected podSecurityPolicyConfig config
```

marshallford:
I hope I rebased correctly, dealing with the generated files wasn't fun. Had to step through a couple of manual fixes and `git rebase --continue`.

bharathkkb:
@marshallford I am seeing some submodule generation errors, could you run `make build` and commit? That should take care of that error. 
```
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=.git' /workspace/modules/beta-public-cluster-update-variant/cluster.tf /tmp/tmp.HIfNPcb4Hd/workspace/modules/beta-public-cluster-update-variant/cluster.tf
471,472c471,474
< resource "null_resource" "wait_for_cluster" {
<   count = var.skip_provisioners ? 0 : 1
---
> module "gcloud_wait_for_cluster" {
>   source  = "terraform-google-modules/gcloud/google"
>   version = "~> 1.0.1"
>   enabled = var.skip_provisioners
474,477c476,479
<   triggers = {
<     project_id = var.project_id
<     name       = var.name
<   }
---
>   create_cmd_entrypoint  = "${path.module}/scripts/wait-for-cluster.sh"
>   create_cmd_body        = "${var.project_id} ${var.name}"
>   destroy_cmd_entrypoint = "${path.module}/scripts/wait-for-cluster.sh"
>   destroy_cmd_body       = "${var.project_id} ${var.name}"
.
.
```

autogen/main/dns.tf.tmpl
```diff
@@ -19,18 +19,14 @@
...
-  depends_on = [
-    data.google_client_config.default,
-    google_container_cluster.primary,
-    google_container_node_pool.pools,
-  ]
+module "gcloud_delete_default_kube_dns_configmap" {
+  source                = "terraform-google-modules/gcloud/google"
+  version               = "~> 1.0.1"
+  enabled               = (local.custom_kube_dns_config || local.upstream_nameservers_config) && ! var.skip_provisioners
+  additional_components = ["kubectl"]
```
bharathkkb, line 26: @marshallford wouldn't we also need some dependency here?
maybe something like 
```tf
module_depends_on = concat(
    [data.google_client_config.default.access_token],
    [google_container_cluster.primary.master_version],
    [for pool in google_container_node_pool.pools : pool.name]
)
```

marshallford: What a "Duh" moment, not sure how I missed that.

bharathkkb:
@marshallford looks like tests are green 😄 

> A few other things are we should expose skip_download and possibly upgrade from the gcloud module so that users have that additional flexibility and imho skip_download should default to true.

I would still like to provide this functionality if possible in this PR

marshallford:
@bharathkkb Agreed. Given that I've been at odds with the module dependency issue I figured I'd focus on that first.

Can you expand on what you mean by

> possibly upgrade from the gcloud module

bharathkkb:
> Can you expand on what you mean by
> 
> > possibly upgrade from the gcloud module

@marshallford sure. From the gcloud module I think we should surface the following  `upgrade`, `gcloud_sdk_version`, `skip_download` as optional variables for the GKE modules. 

imho the recommended defaults should be `upgrade:false`, `gcloud_sdk_version:296.0.0` and `skip_download:true`. I believe these defaults will  keep parity with the way we currently do it with local-exec. @morgante any thoughts on these defaults?

morgante:
> imho the recommended defaults should be upgrade:false, gcloud_sdk_version:296.0.0 and skip_download:true. I believe these defaults will keep parity with the way we currently do it with local-exec. @morgante any thoughts on these defaults?

I don't think we need to expose sdk_version, but otherwise agreed.

bharathkkb:
@marshallford the lint tests are failing due to a timeout issue, should be safe to ignore.

modules/beta-private-cluster-update-variant/dns.tf
marshallford: Fixed.

bharathkkb:
@morgante do we need anything else before we can merge this?

PR #408, opened 2020-01-27
Fixes pod_security_policy_config type issue

kri5:
When attempting to use the variable pod_security_policy_config, as it has no specified type,
terraform interpret it has string, leading to errors such as 

```
Error: Invalid dynamic for_each value

  on cluster.tf line 80, in resource "google_container_cluster" "primary":
  80:     for_each = var.pod_security_policy_config
    |----------------
    | var.pod_security_policy_config is "[{\"enabled\":true}]"
```

As it expects a list of map, i specified the type accordingly.
I used any for the value type inside the map for future-proof reason.

Feel free to comment!

### DIFF
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -406,6 +406,7 @@ variable "enable_binary_authorization" {
 }
 
 variable "pod_security_policy_config" {
+  type = list(object({ enabled = bool }))
   description = "enabled - Enable the PodSecurityPolicy controller for this cluster. If enabled, pods must be valid under a PodSecurityPolicy to be created."
 
   default = [{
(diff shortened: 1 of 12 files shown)


autogen/main/variables.tf.tmpl
```diff
@@ -406,6 +406,7 @@ variable "enable_binary_authorization" {
 }
 
 variable "pod_security_policy_config" {
+  type = list(map(string))
```
morgante: I think an object would be better actually.

```suggestion
  type = list(object({ enabled = bool }))
```

kri5: I just updated the PR accordingly, wasn't sure about how strong the constraint should be :)

PR #412, opened 2020-01-29
If release_channel is active on the module, set min_master_version to…

### LINKED ISSUE #383: min_master_version and release_channel conflicting
I want to use the `release_channel` feature in 6.1.1, specifically in the `beta-private-cluster` module, however, skimming the code I'm pretty sure this applies to any of the `beta_*` modules.  In this specific case, I'm using the `release_channel` and specifying a `region` with 3 `zones`, resulting in a `master_version` (as of last Thursday) of 

```
master_version              = "1.14.8-gke.12"
```

The cluster deployed just fine. However now when I trying to apply an unrelated change, I'm seeing

```
        master_version              = "1.14.8-gke.12"
      ~ min_master_version          = "1.14.8-gke.17" -> "1.15.4-gke.22"
```

Trying to apply the change to the cluster results in the following error:

```
Error: googleapi: Error 400: Current master version cannot be updated when releaseChannel REGULAR is set., badRequest
```

I initially tried to use `ignore_changes` in my module invocation to ignore `min_master_version`, and immediately ran into the fact this can't currently be done.

kunalkg11:
Fixes #383 

Changes in the PR:

1. If release_channel is active on the module, set min_master_version to null
2. Modified beta-cluster example, removed zonal test and added regional test
3. Added release_channel to simple-regional-beta example

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -56,7 +56,7 @@ resource "google_container_cluster" "primary" {
 {% endif %}
 
   subnetwork         = data.google_compute_subnetwork.gke_subnetwork.self_link
-  min_master_version = local.master_version
+  min_master_version = var.release_channel != null ? null : local.master_version
 
   logging_service    = var.logging_service
   monitoring_service = var.monitoring_service
@@ -351,7 +351,7 @@ resource "google_container_node_pool" "pools" {
 
   cluster = google_container_cluster.primary.name
 
-  version = lookup(each.value, "auto_upgrade", false) ? "" : lookup(
+  version = lookup(each.value, "auto_upgrade", true) ? "" : lookup(
     each.value,
     "version",
     local.node_version,
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -55,8 +55,13 @@ resource "google_container_cluster" "primary" {
   }
 {% endif %}
 
-  subnetwork         = data.google_compute_subnetwork.gke_subnetwork.self_link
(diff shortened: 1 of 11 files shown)


autogen/main/cluster.tf.tmpl
```diff
@@ -351,7 +351,7 @@ resource "google_container_node_pool" "pools" {
 
   cluster = google_container_cluster.primary.name
 
-  version = lookup(each.value, "auto_upgrade", false) ? "" : lookup(
+  version = lookup(each.value, "auto_upgrade", true) ? "" : lookup(
```
bharathkkb: I think this was unintended

autogen/main/cluster.tf.tmpl
```diff
@@ -56,7 +56,7 @@ resource "google_container_cluster" "primary" {
 {% endif %}
 
   subnetwork         = data.google_compute_subnetwork.gke_subnetwork.self_link
-  min_master_version = local.master_version
+  min_master_version = var.release_channel != null ? null : local.master_version
```
bharathkkb: release_channel is beta so we would need to modify it such that it is not introduced in non beta modules

morgante: Correct. @kunalkg11 You'll need to make this conditional.

autogen/main/cluster.tf.tmpl
kunalkg11: Fixed.

autogen/main/cluster.tf.tmpl
kunalkg11: Fixed.

PR #418, opened 2020-02-04
Add comment for `master_ipv4_cidr_block`

ryuheechul:
just a helpful hint for how that value will be used in some cases


ryuheechul:
I figured this out by looking at https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/4c7b3997d4e9ef38ef7c7fd629b7a1ff5ca0418e/modules/beta-private-cluster/cluster.tf#L195-L199

ryuheechul:
I think it was premature PR since I realised my understanding might be not very correct. So I will close this for now.

Issue #405, opened 2020-01-21
External references from destroy provisioners are deprecated


s1mark:
*module version* - **v6.2.0**
*terraform version* - **0.12.19** *(also affected version 0.12.18)*

```
Warning: External references from destroy provisioners are deprecated
  on .terraform/modules/gke/modules/beta-private-cluster/cluster.tf line 365, in resource "null_resource" "wait_for_cluster":
 365:     command = "${path.module}/scripts/wait-for-cluster.sh ${var.project_id} ${var.name}"
Destroy-time provisioners and their connection configurations may only
reference attributes of the related resource, via 'self', 'count.index', or
'each.key'.
References to other resources during the destroy phase can cause dependency
cycles and interact poorly with create_before_destroy.
(and 2 more similar warnings elsewhere)
```

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/95e169e69f6633ce0e4c62704745536d21cac50b/modules/beta-private-cluster/cluster.tf#L363-L366

Probably other sub-modules also affected

Discussion thread at HashiCorp:
https://github.com/hashicorp/terraform/issues/23679

PR #420, opened 2020-02-05
Fix the warning about deprecated external references from destroy provisioners

qlikcoe:
Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/405

I've used the solution proposed by @apparentlymart in 
https://discuss.hashicorp.com/t/how-to-rewrite-null-resource-with-local-exec-provisioner-when-destroy-to-prepare-for-deprecation-after-0-12-8/4580


morgante:
Test failure is unrelated.

PR #439, opened 2020-02-19
Rollout default_max_pods_per_node to all

### LINKED ISSUE #435: private-cluster does not allow changing default_max_pods_per_node
This is more of a question than an issue. Is there any reason for limiting the private-cluster default_max_pods_per_node to 110, which is the default?

ChandranshuRao14:
Fixes #435 
Beta flag removed to allow changes to `default_max_pods_per_node`
Added test for expected max pods


test/fixtures/shared/outputs.tf
```diff
@@ -82,3 +82,7 @@ output "service_account" {
 output "registry_project_id" {
   value = var.registry_project_id
 }
+
+output "default_max_pods" {
+  value = var.registry_project_id
```
morgante: This looks like a typo? We probably don't need this output at all.

ChandranshuRao14: Fixed.

PR #442, opened 2020-02-21
Use gcloud module in ACM

### LINKED ISSUE #454: GKE Integration tests failing
GKE Integration tests are failing for the beta-cluster. Looks like a recent API change has added a field `mode` in the json response under` ['nodeConfig']['workloadMetadataConfig']`. 

We were testing for object equality in the has `the expected nodeMetadata conseal config` config test for beta-cluster. Although the expected ` "nodeMetadata" => 'EXPOSE'` is present, due to the additional unexpected mode field, it is failing.

bharathkkb:
Refactor ACM module utilizing the gcloud module
Fixes #454 for beta-cluster failing test by changing checking that json`['nodeConfig']['workloadMetadataConfig']` includes `"nodeMetadata" => 'EXPOSE'` instead of eq.


bharathkkb:
@morgante sounds good

PR #451, opened 2020-03-10
Fix kubernetes_version and node_version variables in safer-cluster su…

skinlayers:
…bmodules.

### DIFF
--- a/autogen/safer-cluster/main.tf.tmpl
+++ b/autogen/safer-cluster/main.tf.tmpl
@@ -33,14 +33,14 @@ module "gke" {
 
   // We need to enforce a minimum Kubernetes Version to ensure
   // that the necessary security features are enabled.
-  kubernetes_version = "latest"
+  kubernetes_version = var.kubernetes_version
 
   // Nodes are created with a default version. The nodepool enables
   // auto_upgrade so that the node versions can be kept up to date with
   // the master upgrades.
   //
   // https://cloud.google.com/kubernetes-engine/versioning-and-upgrades
-  node_version = ""
+  node_version = var.node_version
 
   master_authorized_networks = var.master_authorized_networks
 
--- a/modules/safer-cluster-update-variant/main.tf
+++ b/modules/safer-cluster-update-variant/main.tf
@@ -29,14 +29,14 @@ module "gke" {
 
   // We need to enforce a minimum Kubernetes Version to ensure
   // that the necessary security features are enabled.
-  kubernetes_version = "latest"
+  kubernetes_version = var.kubernetes_version
 
   // Nodes are created with a default version. The nodepool enables
   // auto_upgrade so that the node versions can be kept up to date with
   // the master upgrades.
   //
   // https://cloud.google.com/kubernetes-engine/versioning-and-upgrades
-  node_version = ""
+  node_version = var.node_version
 
   master_authorized_networks = var.master_authorized_networks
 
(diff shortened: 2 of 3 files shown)


skinlayers:
@morgante I traced this out last week, and found that it does not appear to conform to the GKE hardening recommendations.

https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#upgrade_your_infrastructure_in_a_timely_fashion_default_2019-11-11
```
CIS GKE Benchmark Recommendation: 6.5.3. Ensure Node Auto-Upgrade is enabled for GKE nodes
[...]
In Google Kubernetes Engine, the masters are patched and upgraded for you automatically. Node auto-upgrade also automatically upgrades nodes in your cluster.
```
https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-upgrades
```
Node pools with auto-upgrades enabled are scheduled for upgrades when they meet the selection criteria (announced in the release notes). Rollouts are phased across multiple weeks to ensure cluster and fleet stability.
[...]
Note: Enabling auto-upgrades does not cause your nodes to upgrade immediately. For more information, see Cluster and node upgrades.
```
https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#upgrading_automatically
```
When you create a cluster using the Google Cloud Console, auto-upgrade is enabled on the cluster and its node pools by default, and Google upgrades your clusters when a new GKE version is selected for auto-upgrade.

When you create a cluster using the gcloud command or the GKE API, node auto-upgrade is currently enabled by default.
```
https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#auto-upgrade-version-selection
```
New GKE versions are released regularly, but a version is not selected for auto-upgrade right away. When a GKE version has accumulated enough cluster usage to prove stability over time, Google selects it as an auto-upgrade target for clusters running a subset of older versions.

New auto-upgrade targets are announced in the release notes. Until an available version is selected for auto-upgrade, you can upgrade to it manually. Occasionally, a version is selected for cluster auto-upgrade and node auto-upgrade during different weeks.
```
https://cloud.google.com/kubernetes-engine/docs/release-notes-regular
https://cloud.google.com/kubernetes-engine/docs/release-notes-stable

https://www.terraform.io/docs/providers/google/r/container_node_pool.html#version
```
version - (Optional) The Kubernetes version for the nodes in this pool. Note that if this field and auto_upgrade are both specified, they will fight each other for what the node version should be, so setting both is highly discouraged. While a fuzzy version can be specified, it's recommended that you specify explicit versions as Terraform will see spurious diffs when fuzzy versions are used. See the google_container_engine_versions data source's version_prefix field to approximate fuzzy versions in a Terraform-compatible way.
```

Putting all of this together, there are a few problems that stand out:
1. For many production environments, pinning the k8s version (and dependency versions in general) is a requirement to avoid always being forced onto the latest version without testing.
2. The terraform docs say a node pool version should not be specified if auto-upgrade is also enabled. If the node pool version is set to "latest", then auto_upgrade is disabled (or conflicting with the specified version), and we're not conforming to the GKE hardening recommendations.
3. Looking at the current output of `gcloud container get-server-config --region us-central1`, and comparing it to the GKE release notes reveals the currently:
  * us-central1 defaultClusterVersion: 1.14.10-gke.17
  * us-central1 "latest" validNodeVersions: 1.15.9-gke.12
  * stable channel: 1.14.10-gke.17
  * regular channel: 1.15.9-gke.8

If we want something newer than the default "stable" channel release, a better option would be to enabled the "regular" channel with auto_upgrade, rather than explicitly setting the version.

Issue #429, opened 2020-02-10
Latest kubernetes provider breaks the auth.tf


qlikcoe:
`kubernetes` provider v1.11 was released 3 hours ago. Right away it breaks kubernetes provider configuration in `auth.tf`: 

```
Error: Failed to initialize config: invalid configuration: no configuration has been provided
  on .terraform/modules/gke/terraform-google-modules-terraform-google-kubernetes-engine-064f308/modules/private-cluster/auth.tf line 29, in provider "kubernetes":
  29: provider "kubernetes" {
```

Pinning it to v1.10 is a current workaround: 

```
provider "kubernetes" {
  version = "= 1.10"
}
```

morgante:
Upstream issue: https://github.com/terraform-providers/terraform-provider-kubernetes/issues/759

lawliet89:
If you want to keep using 1.11 provider, you can do a partial plan first with something like `terraform plan -target module.gke.google_container_cluster.primary`.

EricHorst:
Upstream issue has been resolved with 1.11.1: terraform-providers/terraform-provider-kubernetes#759

@morgante Would you consider moving the kubernetes provider version up to ">= 1.11.1"?

In my shop we are locked at kubernetes provider 1.9 due to terraform-providers/terraform-provider-kubernetes/issues/787 which was just fixed in 1.11.1. This means we're also stuck on terraform-google-kubernetes-engine 6.2 since 7.x is locked on kubernetes provider 1.10. Moving up to 1.11.1 should get beyond both problems. We desperately want some of the 7.3 knobs.

Thanks.

morgante:
I'd be happy to accept/review a PR which increases/changes the provider version constraint assuming tests pass.

PR #453, opened 2020-03-11
update and pin kubernetes provider to >= 1.11.1

EricHorst:
Re-visting #429, the upstream issue has been resolved with 1.11.1 (terraform-providers/terraform-provider-kubernetes#759). This update sets the kubernetes provider version at or greater than the problematic version 1.11.0

Advancing to 1.11.1 also addresses terraform-providers/terraform-provider-kubernetes/issues/787 which was broken in kubernetes 1.10 and 1.11 and prevented running Terraform in a pod, ie Atlantis.

### DIFF
--- a/auth.tf
+++ b/auth.tf
@@ -27,7 +27,7 @@ data "google_client_config" "default" {
   Configure provider
  *****************************************/
 provider "kubernetes" {
-  version                = ">= 1.11.1"
+  version                = "~> 1.10.0, ~> 1.11.1"
   load_config_file       = false
   host                   = "https://${local.cluster_endpoint}"
   token                  = data.google_client_config.default.access_token
--- a/auth.tf
+++ b/auth.tf
@@ -27,7 +27,7 @@ data "google_client_config" "default" {
   Configure provider
  *****************************************/
 provider "kubernetes" {
-  version                = "~> 1.10.0"
+  version                = ">= 1.11.1"
   load_config_file       = false
   host                   = "https://${local.cluster_endpoint}"
   token                  = data.google_client_config.default.access_token
--- a/auth.tf
+++ b/auth.tf
@@ -27,7 +27,7 @@ data "google_client_config" "default" {
   Configure provider
  *****************************************/
 provider "kubernetes" {
-  version                = "~> 1.10.0, ~> 1.11.1"
+  version                = "!= 1.11.0, < 2.0.0"
   load_config_file       = false
   host                   = "https://${local.cluster_endpoint}"
   token                  = data.google_client_config.default.access_token
--- a/auth.tf
+++ b/auth.tf
@@ -27,7 +27,7 @@ data "google_client_config" "default" {
   Configure provider
  *****************************************/
 provider "kubernetes" {
-  version                = "!= 1.11.0, < 2.0.0"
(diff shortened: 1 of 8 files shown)


bharathkkb:
@EricHorst thanks for the contribution. I think we should add a <2.0.0 constraint because there maybe breaking changes. Thoughts?

EricHorst:
@bharathkkb I considered that. My reasoning was that prior to the version being pinned to 1.10, there was no version constraint so when undoing that I returned to effectively no version constraint (upward) except for the very specific one required.

morgante:
Unfortunately my confidence in the stability of the Kubernetes provider is shaken by releasing a breaking change in a major version. In order to allow a wider variety of setups while preventing known issues, I suggest we use `~> 1.10.0, ~> 1.11.1` as the version constraint.

EricHorst:
Incorporated @morgante suggested constraint. An enlightened solution I probably wouldn't have come up with!

EricHorst:
Whoops. I trusted and did not test that this would work. Looking...

EricHorst:
This is more version exploration than I've done before. It seems that while compound expressions are allowed, it doesn't work when the expression results in two disjoint intervals.

Terraform doesn't find a suitable version for any of these:
`~> 1.10.0, ~> 1.11.1`
`>= 1.10.0, < 1.11, >= 1.11.1, < 1.12`
`~> 1.10.0, > 1.11`

How do you feel about simply `!= 1.11.0` which does work?

EricHorst:
Trying to make this palatable and pass tests.

Terraform did not like the other compound forms but is happy with a compound of `!= 1.11.0, < 2.0.0`. This will avoid version 1.11.0 that was reported problematic in #429 but allow us to use either the fallback 1.10.0 or the fixed 1.11.1. And also keep us below 2.0.0 as requested by @bharathkkb.

morgante:
I think we should still pin to at least 1.10. How about `~> 1.10, ~= 1.11.0`?

EricHorst:
@morgante your last suggestion doesn't work as typed. ~> 1.10 means `>= 1.10, < 2.0` which renders the `~= 1.11.0` meaningless.

I think you probably meant to type `~> 1.10, != 1.11.0` and in retrospect I think that is exactly what we want based on the conversation in this PR. In `master` it is restrictively pinned to 1.10.x. It is desired to have `>= 1.10` and `< 2.0` but avoid the known bad version `1.11.0`  The last commit implements this.

EricHorst:
Closed/reopened not paying attention to what tab I was in, not trying to be annoying.

morgante:
> I think you probably meant to type ~> 1.10, != 1.11.0 and in retrospect I think that is exactly what we want based on the conversation in this PR

Indeed I did, sorry for the typo!

EricHorst:
We're seeing sporadic problems with Terraform Registry on `init` here locally so I assume that's why the triggered test here failed after I bumped it by closing and opening. It was passing before What's the easiest way to retrigger the test?

morgante:
I just manually queued a test run. You can also push an (empty) commit to force it as well.

bharathkkb:
@EricHorst  @morgante test failures are due to a change in the proto for GKE and the way we were asserting eq; I opened an issue #454 and  fix is in PR #442

PR #456, opened 2020-03-16
Feature: Add autoscaling profiles

### LINKED ISSUE #455: Add support for autoscaling_profile
Add support for autoscaling_profile in the cluster_autoscaling block. 
https://www.terraform.io/docs/providers/google/r/container_cluster.html#cluster_autoscaling

bharathkkb:
Fixes #455 
- Adds autoscaling_profile to the cluster_autoscaling obj
- Added test


autogen/main/cluster.tf.tmpl
```diff
@@ -68,7 +68,8 @@ resource "google_container_cluster" "primary" {
 
 {% if beta_cluster %}
   cluster_autoscaling {
-    enabled = var.cluster_autoscaling.enabled
+    enabled             = var.cluster_autoscaling.enabled
+    autoscaling_profile = var.cluster_autoscaling.autoscaling_profile
```
morgante: Should we set a default here (BALANCED)?

bharathkkb: I have the default in the cluster_autoscaling to BALANCED.

morgante: Right, but that default will only be applied if `cluster_autoscaling` is omitted entirely. I should be able to specify `autoscaling_profile = null` to get a sane default (or should the default be disabled?).

bharathkkb: ah, yes good point; will update :)

PR #462, opened 2020-03-22
Add Istio config block

### LINKED ISSUE #457: Add support to mTLS for Istio Auth
In the google provider there is an option to require mTLS for Istio Auth, seen in the GCP console as `Strict`.

bharathkkb:
- fixes #457 
- switch from a boolean to config block for istio
- support additional parameters auth in the config block


autogen/main/variables.tf.tmpl
```diff
@@ -387,8 +387,15 @@ variable "master_ipv4_cidr_block" {
 {% if beta_cluster %}
 
 variable "istio" {
-  description = "(Beta) Enable Istio addon"
-  default     = false
+  description = "Istio configs"
+  type        = object({
```
morgante: I'd like to avoid changing this variable type. Instead can we add a new variable `istio_auth` and leave the existing setting untouched?

bharathkkb: @morgante my reasoning for using the object type was to keep it aligned with the way we are using `cluster_autoscaling` as we may have new arguments come in. But happy to change it if necessary.

morgante: Since doing defaults for objects is still complicated, I'd prefer to keep this flat (with `var.istio ` and `var.istio_auth`) for now.

bharathkkb: gotcha, sg

bharathkkb: updated to use istio_auth

Issue #393, opened 2020-01-02
Generate kubeconfig from output


yashbhutwala:

morgante:
Can you clarify your ask here?

yashbhutwala:
Right now, afaik, there is no way to take the output of this module and turn it into a valid `~/.kube/config` file.  Correct me if I'm wrong :)

morgante:
The outputs include the `ca_certificate` and `endpoint`, which should be sufficient to construct a kubeconfig file: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine#outputs

yashbhutwala:
Is that enough to generate a kubeconfig file, what about `client-certificate-data` and `client-key-data`

PR #469, opened 2020-03-30
Add an `auth` submodule outputting a `kubeconfig`

rileykarson:
As well as convenience values for the `kubernetes` and `helm` providers.

I'm running the tests following the guide at https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#noninteractive-execution, `make docker_test_integration` has been running since morning though.

/cc @morgante 

Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/393


morgante:
@rileykarson Since you're only editing/writing one test, you might want to try using interactive mode to only run that suite.

bharathkkb:
@rileykarson I believe you have not added the fixture that invokes the example and hence the int tests are not being called.

For adding a new test you will also need to make an entry [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/.kitchen.yml) which will let kitchen know the existence of the test. After that you should be able to do `kitchen_do list` and find the `simple_regional_with_kubeconfig` test suite.

Then add an entry [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/build/int.cloudbuild.yaml)  which will let cloud build build that particular suite in parallel.

rileykarson:
I added the additional boilerplate, and it appeared to succeed when running in interactive mode. Let me know if you'd like me to exercise the generated kubeconfig.

bharathkkb:
LGTM

PR #470, opened 2020-03-31
Firewall support

### LINKED ISSUE #452: Add optional firewall support
Consider adding a opt in method for this module to create common/useful firewalls related to the cluster. I tend to always have to create 2 firewall rules due to running in a slightly more restricted VPC or for certain workloads (Admission Controllers).

### Explict Egress rule for cluster traffic.

GKE already creates a `Ingress` rule for the cluster and then relies on the fact there is a [implicit](https://cloud.google.com/vpc/docs/firewalls#default_firewall_rules) allow-all-egress rule for VPCs by default. That assumption doesn't work when you are running locked down VPCs that create a `default-deny-all` egress rule (also w/ intranode visibility enabled)

```tf
resource "google_compute_firewall" "egress" {
  name      = "gke-${var.name}-allow-intra-cluster-egress"
  project   = var.network_project_id
  network   = var.network
  direction = "EGRESS"
  priority  = 2000
  target_tags = ["gke-${var.name}"]  
  destination_ranges = [
    # GKE Master - Only for Private clusters!
    var.master_ipv4_cidr_block,
    # GKE Node Subnet
    data.google_compute_subnetwork.gke_subnetwork.ip_cidr_range,
    # GKE Alias Ranges
    local.alias_ranges_cidr[var.ip_range_pods],
    local.alias_ranges_cidr[var.ip_range_services]
  ]

  allow { protocol = "tcp"}
  allow { protocol = "udp"}
  allow { protocol = "icmp"}
}
```

### Allow non 443 ports for GKE Master -> nodes due to admission controllers

This is for when cluster workloads include webhooks/admission controllers that run on non 443 ports. Without this you will end up seeing timeouts and pods are blocked from being created.

https://github.com/kubernetes/kubernetes/issues/79739

Examples: 9443 (Elasticsearch Operator https://github.com/elastic/cloud-on-k8s/issues/896), 15017 (Istio 1.5+ https://github.com/istio/istio/issues/19532)

```tf
# Allow GKE master to hit non 443 ports for Webhooks/Admission Controllers
# https://github.com/kubernetes/kubernetes/issues/79739
resource "google_compute_firewall" "master_webhooks" {
  name      = "gke-${var.name}-allow-master-webhooks"
  project   = var.network_project_id
  network   = var.network
  direction = "INGRESS"
  priority  = 2000

  allow {
    protocol = "tcp"
    ports    = var.webhook_ports
  }

  source_ranges = [var.master_ipv4_cidr_block] # Public Cluster: "${module.gke.endpoint}/32"
  target_tags   = ["gke-${var.name}"]
}
```

Dev25:
Closes #452  

Add the 2 firewall rules raised in that issue, which can be opted out using `var.firewall_enabled`.  

Common inbound ports are also white listed by default (8443/9443 and 15017 for istio 1.5) for a better UX (port list controlled by `var.firewall_inbound_ports`)

Refactoring:
- Save `"gke-${var.name}"` network tag format into locals, this module already adds that network tag to all node pools
- Move `networks.tf` to `autogen/` so module variants have access to `data.google_compute_subnetwork.gke_subnetwork` which can be used to determine pod CIDR
- Misc formatting fixes to templates so the latest docker image is happy for `make build docker_test_lint`

TODO:
- README update, this requires the terraform runner to have necessary IAM permissions to make firewall changes in the VPC project

### DIFF
--- a/autogen/main/README.md
+++ b/autogen/main/README.md
@@ -240,7 +240,7 @@ The [project factory](https://github.com/terraform-google-modules/terraform-goog
 In order to execute this module you must have a Service Account with the
 following project roles:
 - roles/compute.viewer
-- roles/compute.securityAdmin
+- roles/compute.securityAdmin (only required if `add_cluster_firewall_rules` is set to `true`)
 - roles/container.clusterAdmin
 - roles/container.developer
 - roles/iam.serviceAccountAdmin
--- a/autogen/main/README.md
+++ b/autogen/main/README.md
@@ -240,6 +240,7 @@ The [project factory](https://github.com/terraform-google-modules/terraform-goog
 In order to execute this module you must have a Service Account with the
 following project roles:
 - roles/compute.viewer
+- roles/compute.securityAdmin
 - roles/container.clusterAdmin
 - roles/container.developer
 - roles/iam.serviceAccountAdmin
--- a/autogen/main/firewall.tf.tmpl
+++ b/autogen/main/firewall.tf.tmpl
@@ -25,7 +25,7 @@
   a default-deny egress rule
  *****************************************/
 resource "google_compute_firewall" "intra_egress" {
-  count       = var.firewall_enabled ? 1 : 0
+  count       = var.add_cluster_firewall_rules ? 1 : 0
   name        = "gke-${substr(var.name, 0, min(25, length(var.name)))}-intra-cluster-egress"
   description = "Managed by terraform gke module: Allow pods to communicate with each other and the master"
   project     = local.network_project_id
@@ -61,7 +61,7 @@ resource "google_compute_firewall" "intra_egress" {
   https://github.com/kubernetes/kubernetes/issues/79739
  *****************************************/
 resource "google_compute_firewall" "master_webhooks" {
-  count       = var.firewall_enabled ? 1 : 0
+  count       = var.add_cluster_firewall_rules ? 1 : 0
   name        = "gke-${substr(var.name, 0, min(25, length(var.name)))}-webhooks"
   description = "Managed by terraform gke module: Allow master to hit pods for admission controllers/webhooks"
   project     = local.network_project_id
--- a/autogen/main/firewall.tf.tmpl
+++ b/autogen/main/firewall.tf.tmpl
@@ -1,0 +1,84 @@
+/**
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+{{ autogeneration_note }}
+
+
+/******************************************
+  Match the gke-<CLUSTER>-<ID>-all INGRESS
+  firewall rule created by GKE but for EGRESS
+
+  Required for clusters when VPCs enforce
+  a default-deny egress rule
+ *****************************************/
+resource "google_compute_firewall" "intra_egress" {
+  count       = var.firewall_enabled ? 1 : 0
+  name        = "gke-${substr(var.name, 0, min(25, length(var.name)))}-intra-cluster-egress"
+  description = "Managed by terraform gke module: Allow pods to communicate with each other and the master"
+  project     = local.network_project_id
+  network     = var.network
+  priority    = var.firewall_priority
+  direction   = "EGRESS"
+
+  target_tags = [local.cluster_network_tag]
+  destination_ranges = [
+    local.cluster_endpoint_for_nodes,
+    local.cluster_subnet_cidr,
+    local.cluster_alias_ranges_cidr[var.ip_range_pods],
+  ]
+
+  # Allow all possible protocols
+  allow { protocol = "tcp" }
+  allow { protocol = "udp" }
+  allow { protocol = "icmp" }
+  allow { protocol = "sctp" }
+  allow { protocol = "esp" }
+  allow { protocol = "ah" }
+
+  depends_on = [
+    google_container_cluster.primary,
+  ]
+}
+
+
+/******************************************
+  Allow GKE master to hit non 443 ports for
+  Webhooks/Admission Controllers
+
+  https://github.com/kubernetes/kubernetes/issues/79739
+ *****************************************/
+resource "google_compute_firewall" "master_webhooks" {
+  count       = var.firewall_enabled ? 1 : 0
+  name        = "gke-${substr(var.name, 0, min(25, length(var.name)))}-webhooks"
+  description = "Managed by terraform gke module: Allow master to hit pods for admission controllers/webhooks"
+  project     = local.network_project_id
+  network     = var.network
+  priority    = var.firewall_priority
+  direction   = "INGRESS"
+
+  source_ranges = [local.cluster_endpoint_for_nodes]
+  target_tags   = [local.cluster_network_tag]
+
+  allow {
+    protocol = "tcp"
+    ports    = var.firewall_inbound_ports
+  }
+
+  depends_on = [
+    google_container_cluster.primary,
+  ]
+
+}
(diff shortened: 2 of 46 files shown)


autogen/main/firewall.tf.tmpl
```diff
@@ -0,0 +1,84 @@
...
+/******************************************
+  Match the gke-<CLUSTER>-<ID>-all INGRESS
+  firewall rule created by GKE but for EGRESS
+
+  Required for clusters when VPCs enforce
+  a default-deny egress rule
+ *****************************************/
+resource "google_compute_firewall" "intra_egress" {
+  count       = var.firewall_enabled ? 1 : 0
+  name        = "gke-${substr(var.name, 0, min(25, length(var.name)))}-intra-cluster-egress"
```
bharathkkb, line 29: Thanks for the contribution, looks like tests are failing due to duplicate fw names being given to two fixtures `beta-cluster-local` and `sandbox-enabled-local`. Both use the same example [simple-regional-beta](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/examples/simple_regional_beta/main.tf) with a unique prefix. With the substring in the FW name, it takes off that prefix and hence the clash. 

One possibility is to provide a default value for cluster name in the example and override them in the fixtures to set names as beta-cluster and sandbox-enabled to get around this?

morgante: Maybe we could switch from suffix to prefix?

autogen/main/README.md
```diff
@@ -240,6 +240,7 @@ The [project factory](https://github.com/terraform-google-modules/terraform-goog
 In order to execute this module you must have a Service Account with the
 following project roles:
 - roles/compute.viewer
+- roles/compute.securityAdmin
```
morgante, line 243: Can we note this is only required if firewalls are enabled?

autogen/main/networks.tf
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+{{ autogeneration_note }}
+
+data "google_compute_network" "gke_network" {
```
morgante, line 19: Note this might raise some issues with networks created inline. Could we make these data sources conditional on the firewall_enabled var?

autogen/main/variables.tf.tmpl
```diff
@@ -488,3 +488,22 @@ variable "enable_shielded_nodes" {
   default     = true
 }
 {% endif %}
+
+
+variable "firewall_enabled" {
```
morgante, line 493: Let's be more explicit about what firewall rules we're allowing. Maybe `add_cluster_firewall_rules`?

autogen/main/variables.tf.tmpl
```diff
@@ -488,3 +488,22 @@ variable "enable_shielded_nodes" {
...
+  type        = bool
+  description = "Create additional firewall rules"
+  default     = true
```
morgante, line 496: Let's default to false so this can be included in a minor release.

autogen/main/variables.tf.tmpl
Dev25: Changed to your suggestion

autogen/main/README.md
Dev25: Done

autogen/main/firewall.tf.tmpl
Dev25: Do we want firewall support enabled on all examples/tests or should we create a dedicated example for this instead? Because this is going to be disabled by default now so these breaking tests will presumably pass again.

morgante: Let's enable it one one of the examples, maybe `shared_vpc`. We can leave the default on the rest.

Dev25: Enabled on `shared_vpc`

Dev25: Could either of you explain why https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/runs/605538699 failed? I did just run the `shared-vpc-local` test locally and it passed.

autogen/main/networks.tf
Dev25: Will look into doing that.

morgante:
@Dev25 This is the error which shows up for that test:
```
       Error: project: required field is not set
       
         on ../../../main.tf line 22, in data "google_compute_zones" "available":
         22: data "google_compute_zones" "available" {
       
       
       
       Error: project: required field is not set
       
         on ../../../main.tf line 125, in data "google_container_engine_versions" "region":
        125: data "google_container_engine_versions" "region" {
```

bharathkkb:
That does seem weird, let me try re running.

autogen/main/firewall.tf.tmpl
bharathkkb: @Dev25 looks like it was some thing unrelated and worked when I reran it.

autogen/main/networks.tf
bharathkkb: I think the regular count conditional approach should work on datasources too

morgante: I think this is the only thing blocking merge now.

Dev25: Should be fixed in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/470/commits/8ffa5e8326de3cab6db7cbc11c8019e4e1ecc9df 
Also removed the top level network datasource since it seems unused

Dev25:
btw a old CI job seems to be stuck, you may want to explicitly cancel it: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/runs/605516376

>Google Cloud Build / Trigger: 10ee403a-235b-42d6-8acc-5ebfe78cfcae
started 1d 6h 13m 32s ago

bharathkkb:
@Dev25 that succeeded on cloud build, for some reason it didnt report back to GH

autogen/main/networks.tf
Dev25: Plus https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/470/commits/ceef39536910631b11db8af544e06eaeb437c019 to fix attribute errors, not pretty but it works/keeps it all conditional.

```tf
Error: Unsupported attribute

  on .terraform/modules/blue_gke.gke/main.tf line 75, in locals:
  75:   cluster_alias_ranges_cidr = { for range in toset(data.google_compute_subnetwork.gke_subnetwork.*.secondary_ip_range) : range.range_name => range.ip_cidr_range }
```

modules/beta-private-cluster-update-variant/main.tf
```diff
@@ -71,6 +71,9 @@ locals {
   // auto upgrade by defaults only for regional cluster as long it has multiple masters versus zonal clusters have only have a single master so upgrades are more dangerous.
   default_auto_upgrade = var.regional ? true : false
 
+  cluster_subnet_cidr       = var.add_cluster_firewall_rules ? data.google_compute_subnetwork.gke_subnetwork[0].ip_cidr_range : null
+  cluster_alias_ranges_cidr = var.add_cluster_firewall_rules ? { for range in toset(data.google_compute_subnetwork.gke_subnetwork[0].secondary_ip_range) : range.range_name => range.ip_cidr_range } : {}
```
erichaase, line 75: I'm getting the following error caused by this line:
```
Error: Iteration over null value

  on .terraform/modules/corp-dev-us-east1-c-01.gke/terraform-google-kubernetes-engine-10.0.0/modules/beta-private-cluster-update-variant/main.tf line 72, in locals:
  72:   cluster_alias_ranges_cidr = var.add_cluster_firewall_rules ? { for range in toset(data.google_compute_subnetwork.gke_subnetwork[0].secondary_ip_range) : range.range_name => range.ip_cidr_range } : {}
    |----------------
    | data.google_compute_subnetwork.gke_subnetwork[0].secondary_ip_range is null

A null value cannot be used as the collection in a 'for' expression.
```

I'm having a hard time troubleshooting this on our end and would appreciate any feedback on this.

morgante: Could you open a new issue w/ your Terraform config?

erichaase: Yeah, sure:

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/608

Issue #449, opened 2020-03-07
Safer Cluster Update Variant does not propogate zones


jamesthimont:
The gke module is missing the `zones` property
```
module "gke" {
  source             = "../beta-private-cluster-update-variant/"
  project_id         = var.project_id
  name               = var.name
  regional           = var.regional
  region             = var.region
 ***zones              = var.zones***
  network            = var.network
  network_project_id = var.network_project_id
```

morgante:
Indeed it is! We could should certainly add it.

PR #474, opened 2020-04-03
Fix zonal bug

jmymy:
The Safer modules support regional and zonal clusters. However, the zones variable is never passed so the config errors when setting regional = false

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/449


jmymy:
I honestly have no idea how the linting works.

the error is 
`Error: submodule's files generation has not been run, please run the
'make build' command and commit changes`

make build removes my changes.

bharathkkb:
LGTM

Issue #486, opened 2020-04-13
safer-cluster-update-variant always forces upgrade to latest GKE version.


skinlayers:
Transfered from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/451

@morgante I traced this out last week, and found that it does not appear to conform to the GKE hardening recommendations.

https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster#upgrade_your_infrastructure_in_a_timely_fashion_default_2019-11-11

"CIS GKE Benchmark Recommendation: 6.5.3. Ensure Node Auto-Upgrade is enabled for GKE nodes
[...]
In Google Kubernetes Engine, the masters are patched and upgraded for you automatically. Node auto-upgrade also automatically upgrades nodes in your cluster."

https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-upgrades

"Node pools with auto-upgrades enabled are scheduled for upgrades when they meet the selection criteria (announced in the release notes). Rollouts are phased across multiple weeks to ensure cluster and fleet stability.
[...]
Note: Enabling auto-upgrades does not cause your nodes to upgrade immediately. For more information, see Cluster and node upgrades."

https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#upgrading_automatically

"When you create a cluster using the Google Cloud Console, auto-upgrade is enabled on the cluster and its node pools by default, and Google upgrades your clusters when a new GKE version is selected for auto-upgrade.

When you create a cluster using the gcloud command or the GKE API, node auto-upgrade is currently enabled by default."

https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-upgrades#auto-upgrade-version-selection

"**New GKE versions are released regularly, but a version is not selected for auto-upgrade right away. When a GKE version has accumulated enough cluster usage to prove stability over time, Google selects it as an auto-upgrade target for clusters running a subset of older versions.**

New auto-upgrade targets are announced in the release notes. Until an available version is selected for auto-upgrade, you can upgrade to it manually. Occasionally, a version is selected for cluster auto-upgrade and node auto-upgrade during different weeks."

https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels

"Regular: 
Production clusters that need features not yet offered in the Stable channel.
These versions are considered production-quality. Known issues generally have known workarounds.

Stable:
[...]

PR #487, opened 2020-04-13
[safer-cluster] Replace "kubernetes_version" with "release_channel"

skinlayers:
See https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/486

### DIFF
--- a/autogen/safer-cluster/main.tf.tmpl
+++ b/autogen/safer-cluster/main.tf.tmpl
@@ -34,14 +34,14 @@ module "gke" {
 
   // We need to enforce a minimum Kubernetes Version to ensure
   // that the necessary security features are enabled.
-  kubernetes_version = "latest"
+  kubernetes_version = var.kubernetes_version
 
   // Nodes are created with a default version. The nodepool enables
   // auto_upgrade so that the node versions can be kept up to date with
   // the master upgrades.
   //
   // https://cloud.google.com/kubernetes-engine/versioning-and-upgrades
-  node_version = ""
+  node_version = var.node_version
 
   master_authorized_networks = var.master_authorized_networks
 
--- a/autogen/safer-cluster/main.tf.tmpl
+++ b/autogen/safer-cluster/main.tf.tmpl
@@ -41,8 +41,6 @@ module "gke" {
   // the master upgrades.
   //
   // https://cloud.google.com/kubernetes-engine/versioning-and-upgrades
-  node_version = var.node_version
-
   release_channel = var.release_channel
 
   master_authorized_networks = var.master_authorized_networks
--- a/autogen/safer-cluster/main.tf.tmpl
+++ b/autogen/safer-cluster/main.tf.tmpl
@@ -43,6 +43,8 @@ module "gke" {
   // https://cloud.google.com/kubernetes-engine/versioning-and-upgrades
   node_version = var.node_version
 
+  release_channel = var.release_channel
+
   master_authorized_networks = var.master_authorized_networks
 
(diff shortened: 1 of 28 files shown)


autogen/safer-cluster/variables.tf.tmpl
```diff
@@ -68,7 +68,7 @@ variable "subnetwork" {
 variable "kubernetes_version" {
   type        = string
   description = "The Kubernetes version of the masters. If set to 'latest' it will pull latest available version in the selected region. The module enforces certain minimum versions to ensure that specific features are available. "
-  default     = "latest"
+  default     = ""
```
morgante, line 71:
```suggestion
  default     = null
```

morgante: It seems like we want to avoid setting kubernetes_version at all by default?

skinlayers: That makes sense.

autogen/safer-cluster/main.tf.tmpl
```diff
@@ -34,14 +34,16 @@ module "gke" {
...
-  kubernetes_version = "latest"
+  kubernetes_version = var.kubernetes_version
 
   // Nodes are created with a default version. The nodepool enables
   // auto_upgrade so that the node versions can be kept up to date with
   // the master upgrades.
   //
   // https://cloud.google.com/kubernetes-engine/versioning-and-upgrades
-  node_version = ""
+  node_version = var.node_version
```
skinlayers, line 44: @morgante Is node_version actually used anywhere in any of the modules? I what grepping through the code last night, but AFAICT, node_version is never consumed by any resource. Am I crazy, or can we just remove node_version, node_version_regional, and node_version_zonal from autogen/main & autogen/safer-cluster?

morgante: You're right, it looks like we might have dropped it entirely. Feel free to remove the variable in that case.

autogen/safer-cluster/variables.tf.tmpl
```diff
@@ -77,6 +77,12 @@ variable "node_version" {
   default     = ""
 }
 
+variable "release_channel" {
+  type        = string
+  description = "(Beta) The release channel of this cluster. Accepted values are `UNSPECIFIED`, `RAPID`, `REGULAR` and `STABLE`. Defaults to `REGULAR`."
+  default     = "REGULAR"
```
umairidris, line 83: Out of curiosity, why pick REGULAR as default over STABLE?

skinlayers: See https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/486 & https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/451 for additional context.

umairidris: Right, I looked through it but didn't see any particular reason for picking regular over stable.

While I don't have a strong opinion, based on the GKE documentation I think a *default* setting would be better suited to the more stable option. Users who want newer features etc can switch to regular etc if they would like by overriding. 

I will leave it up to your judgment, just doing a drive-by here. Thanks for the PR, we're hoping to use it as soon as it's released :)

morgante: I think regular is a reasonable default, because we'd potentially like to add enhanced security features in the future w/o waiting for them to appear in the stable channel. For clusters which require extra stability, they can override.

Issue #316, opened 2019-11-12
add GKE metering (now in GA) for all module and submodule versions


gcamus59:
GKE metering is not anymore a Beta feature, it's now in General Availability.
So, the feature should be available for use in every kind of modules.

It's already available in the beta' module versions. It seems that moving the code outside from the beta_cluster IF conditions (inside the autogen repository) will enable the feature for all module versions.

After beginning work on the modification in a pull request, I saw that the parameter is not yet available in the non beta google Terraform provider (see https://github.com/terraform-providers/terraform-provider-google/issues/4634)

morgante:
Thanks, this should be easy to fix once https://github.com/terraform-providers/terraform-provider-google/issues/4634 is resolved.

thomas-riccardi:
https://github.com/terraform-providers/terraform-provider-google/commit/a646d86669acdf34eedf8515521a3bf56d8b0c4b is now included in `google` 3.16.0.

Also, would it be possible to make `enable_network_egress_metering` configurable (which is set to true here whereas in the provider it defaults to false)? (And maybe `enable_resource_consumption_metering` too for completeness).

gcamus59:
Thanks, I'll work on it this week and I will craft a pull request.

PR #491, opened 2020-04-20
feat: Add fully configurable resource usage export block in GA

gcamus59:
Add GKE metering block which it's now in GA into Google Provider version since 3.16. 
In the previous implementation on the BETA modules, only the BigQuery dataset was editable. In this version, all parameters can be updated such as: 

- Network egress
- Resource consumption metering
- BigQuery dataset ID

⚠️ Tests cases with fixed version of the Google provider older than 3.16 are failing.

See #316


gcamus59:
Waiting your feedback regarding the requirement of the Google provider in 3.16+ version (is it the right implementation for this kind fo new feature only available in the latest google provider versions, should I update other tests cases requirements, etc..)

bharathkkb:
Yes I think its reasonable to bump provider versions. In fact I think we should be doing `~> 3.x` instead of `~> 3.x.0` so that we keep getting the latest 3.x provider. Any thoughts on changing all examples to  `~> 3.x` @morgante ?

morgante:
In terms of upgrading provider versions:
1. In the examples, feel free to upgrade to a new pinned provider (ex. `~> 3.16.0`). The reason we pin to a specific minor version is to avoid cases where the minimum required provider version is *silently* changed w/o us knowing. We want to be able to call that out in release notes.
2. In the module versions.tf we can change the pin to the minimum required (ex. `~> 3.16`).

autogen/main/variables.tf.tmpl
```diff
@@ -188,6 +188,12 @@ variable "node_pools_metadata" {
     default-node-pool = {}
   }
 }
+
+variable "resource_usage_export" {
```
morgante, line 192: Instead of switching to a direct object list (which is also a very confusing data type), we should retain the existing `resource_usage_export_dataset_id` variable and add `enable_network_egress_export` and `enable_resource_consumption_export` variables.

gcamus59: Ok, thanks for the feedback. I'll update it.

bharathkkb:
LGTM, CI error seems unrelated but there are some conflicts. Please rebase

PR #497, opened 2020-04-27
Add GCE PD CSI Driver beta support

Dev25:
[3.19](https://github.com/terraform-providers/terraform-provider-google-beta/releases/tag/v3.19.0) beta provider was released today that includes the CSI addon support

This bumps the min beta provider to 3.19, GA provider is still 3.16

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -152,6 +152,10 @@ resource "google_container_cluster" "primary" {
     dns_cache_config {
       enabled = var.dns_cache
     }
+
+    gce_persistent_disk_csi_driver_config {
+      enabled = var.gce_pd_csi_driver
+    }
     {% endif %}
   }
 
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -153,8 +153,12 @@ resource "google_container_cluster" "primary" {
       enabled = var.dns_cache
     }
 
-    gce_persistent_disk_csi_driver_config {
-      enabled = var.gce_pd_csi_driver
+    dynamic "gce_persistent_disk_csi_driver_config" {
+      for_each = local.cluster_gce_pd_csi_config
+
+      content {
+        enabled = var.gce_persistent_disk_csi_driver_config.value.enabled
+      }
     }
     {% endif %}
   }
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -157,7 +157,7 @@ resource "google_container_cluster" "primary" {
       for_each = local.cluster_gce_pd_csi_config
 
       content {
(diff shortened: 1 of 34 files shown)


autogen/main/cluster.tf.tmpl
```diff
@@ -152,6 +152,10 @@ resource "google_container_cluster" "primary" {
     dns_cache_config {
       enabled = var.dns_cache
     }
+
+    gce_persistent_disk_csi_driver_config {
```
morgante, line 156: We should make this a dynamic block, as I've observed consistent issues with the behavior of addon blocks in `disabled` states.

autogen/main/outputs.tf.tmpl
```diff
@@ -147,6 +147,11 @@ output "dns_cache_enabled" {
   value       = local.cluster_dns_cache_enabled
 }
 
+output "cluster_gce_pd_csi_driver_enabled" {
+  description = "Whether GCE Persistent Disk CSI Driver enabled"
```
morgante, line 151:
```suggestion
  description = "Whether GCE Persistent Disk CSI Driver is enabled"
```

autogen/main/outputs.tf.tmpl
```diff
@@ -147,6 +147,11 @@ output "dns_cache_enabled" {
...
+  value       = local.cluster_gce_pd_csi_driver_enabled
```
morgante, line 152: This should be a reference to the cluster itself, not merely passing back the input.

Or omit entirely, not sure this output is needed.

Dev25: You're right, i'm going to drop the output/local reference completely

autogen/main/cluster.tf.tmpl
Dev25: Right, i based this on the initial DNS cache PR. Did you want me to fix that addon in this PR too?

Dev25:
```tf
-  cluster_gce_pd_csi_config = var.gce_pd_csi_driver ? [{ enabled = true }] : []
+  cluster_gce_pd_csi_config = var.gce_pd_csi_driver ? [{ enabled = true }] : [{ enabled = false }]
```

I couldn't disable this addon with the initial dynamic block approach (based on istio addon), one difference with the older addons is that newer addon config blocks (dns/csi) use `enabled` rather then `disabled` states. When trying to plan it doesn't indicate a change needs to be made when trying to flip back to disabled

I was only able to disable when forcing a dynamic block by proving a explicit `[{ enabled = false }]`. So the dynamic block is always going to trigger now, do you still want to keep this approach then?

morgante: Hmm, that's odd. It might indeed by driven by the defaults.

Can you try importing an existing cluster (with this disabled) using the static block approach? If import is successful, then I'm fine with using that approach.

PR #500, opened 2020-04-30
Fixes #489 Identity namespace output for beta clusters

### LINKED ISSUE #489: identity_namespace output is "enabled" instead of [project_id].svc.id.goog for beta-private-cluster
We are using module version `8.1.0` and when I create a beta-private-cluster with the default `identity_namespace = "enabled"` the output value of `identity_namespace` is also `"enabled"`. I would expect `[project_id].svc.id.goog` as output here. Any opinions on this? Happy to issue a PR if needed :)

sebastianneb:
The identity namespace flag was "enabled". Changed the output value to reference the actual identity namespace of the cluster / the project.

This PR fixes #489.

### DIFF
--- a/autogen/main/outputs.tf.tmpl
+++ b/autogen/main/outputs.tf.tmpl
@@ -169,7 +169,7 @@ output "release_channel" {
 
 output "identity_namespace" {
   description = "Workload Identity namespace"
-  value       = var.identity_namespace
+  value       = length(local.cluster_workload_identity_config) > 0 ? local.local.cluster_workload_identity_config[0].identity_namespace : null
   depends_on = [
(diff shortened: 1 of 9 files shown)


sebastianneb:
I ran all integration tests locally. All tests are running through apart from the following with the following errors:

- `simple-zonal-local` --> `error: the path "/workspace/acm.yaml" does not exist`. Obvious, where do I get this `acm.yaml` from?

So maybe you could also give me hint why the pipeline is falling? I did run all steps described in `CONTRIBUTING.md` and at least the tests I changed for beta-cluster are green...

morgante:
The acm error looks unrelated, so we can ignore it.

PR #512, opened 2020-05-06
Enable GKE_METADATA_SERVER as node_metadata for beta-clusters (#490)

### LINKED ISSUE #490: Beta clusters don't automatically enable Workload Identity
When using a beta private cluster in version `8.1.0` identity namespace gets enabled by default, but not workload identity. 

We delete the default node-pool after creation and use a node pool with the following basic settings: 
```
node_pools = [
    {
      name         = "main"
      node_count   = 2
      image_type   = "COS_CONTAINERD"
      min_count    = 2
      max_count    = 4
      machine_type = "n1-standard-8"
    }
  ]
```

When looking at the GCloud UI, we can see, that identity namespace for the cluster is set. Looking at the node pool we can see, that Security -> Enable GKE Metadata Server is not enabled (which is required for Workload Identity).

Looking at the code it seems like we have to set `node_metadata = "GKE_METADATA_SERVER"` on the module on our own to enable that feature.

Since beta cluster should enable Identity Workload as default, this should also be the default for `node_metadata`. Also happing to issue a PR if you guys say that this is really a bug that I encounter :)

sebastianneb:
Fixes #490.

Since workload identity is enabled by default for beta-clusters, GKE_METADATA should be the default for exposing metadata to running pods as well in order to enabled workload identity in full.

### DIFF
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -454,7 +454,7 @@ variable "pod_security_policy_config" {
 
 variable "node_metadata" {
   description = "Specifies how node metadata is exposed to the workload running on the node"
-  default     = "SECURE"
+  default     = "GKE_METADATA_SERVER"
   type        = string
 }
 
--- a/examples/simple_regional_beta/README.md
+++ b/examples/simple_regional_beta/README.md
@@ -18,7 +18,6 @@ This example illustrates how to create a simple cluster with beta features.
 | ip\_range\_services | The secondary ip range to use for services | string | n/a | yes |
 | istio | Boolean to enable / disable Istio | string | `"true"` | no |
 | network | The VPC network to host the cluster in | string | n/a | yes |
-| node\_metadata | Specifies how node metadata is exposed to the workload running on the node | string | `"SECURE"` | no |
 | node\_pools | List of maps containing node pools | list(map(string)) | `<list>` | no |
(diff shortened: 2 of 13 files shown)


sebastianneb:
Some local tests fail atm due to `Google Compute Engine: Not all instances running in IGM after 16.026802185s. Expect 1. Current errors: [GCE_STOCKOUT]: Instance 'gke-node-pool-cluster-rq5o-pool-02-91cb342f-j5lm' creation failed: The zone 'projects/ci-gke-b86c/zones/us-central1-a' does not have enough resources available to fulfill the request.  Try a different zone, or try again later.`. Since this has nothing to do with my changes, I will run the tests later today / tomorrow once I am able to create the resources and run the tests...

The tests in `beta_cluster` are working fine and therefore I already present this PR for review.

sebastianneb:
Some local tests fail atm due to `Google Compute Engine: Not all instances running in IGM after 16.026802185s. Expect 1. Current errors: [GCE_STOCKOUT]: Instance 'gke-node-pool-cluster-rq5o-pool-02-91cb342f-j5lm' creation failed: The zone 'projects/ci-gke-xxxx/zones/us-central1-a' does not have enough resources available to fulfill the request.  Try a different zone, or try again later.`. Since this has nothing to do with my changes, I will run the tests later today / tomorrow once I am able to create the resources and run the tests...

The tests in `beta_cluster` are working fine and therefore I already present this PR for review.

morgante:
@sebastianneb Can you make sure linters pass at least?

sebastianneb:
Sure, missed one reference. Should be fixed now, at least the linters are green now. Will try to run the tests again tomorrow (every test that references the beta-cluster).

PR #523, opened 2020-05-17
Support cluster with network outside of its own project

withnale:
At present the network is assumed to be under `projects/${local.network_project_id}/global/networks/${var.network}` which is not always the case.

This PR checks to see if the specified network starts with `projects/` and if so treats it as an explicit network.


morgante:
This *should* already be supported by specifying `var.network_project_id` to target a network outside the cluster project. Is that not working for you?

Issue #428, opened 2020-02-10
Add NodeLocal DNSCache beta addon support


Dev25:
[NodeLocal DNSCache](https://cloud.google.com/kubernetes-engine/docs/how-to/nodelocal-dns-cache) was recently promoted to beta, add a flag to enable it for beta modules.

Blocked until beta provider supports it, looking at the GKE REST API this should just be a simple [boolean flag](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#Cluster.DnsCacheConfig)

morgante:
Please make sure you open a provider issue as well.

Dev25:
Existing Issue: https://github.com/terraform-providers/terraform-provider-google/issues/5150

Dev25:
This is no longer blocked, [3.14](https://github.com/terraform-providers/terraform-provider-google/releases/tag/v3.14.0) was released yesterday that includes a `dns_cache_config` field.

sebastianortizs4n:
This change always forces replacement of the cluster, even when is set to false and is also disabled on the cluster. This should not happen... Is the same configured value.

morgante:
This seems like a bug in the provider, but we could probably make it a dynamic block.

sebastianortizs4n:
@morgante  this is it, is solved already https://github.com/terraform-providers/terraform-provider-google/issues/6031

sebastianortizs4n:
The solution use `google-beta` provider 3.17.0

morgante:
Just to confirm, it sounds like this is fixed if you upgrade to the latest provider version?

In that case I'm going to close this issue.

jmymy:
I am still seeing this behavior on an existing cluster with nodelocal cache disabled.

 ```+ 1.16.8-gke.4
Terraform v0.12.24
+ provider.google v3.17.0
+ provider.google-beta v3.17.0

+ dns_cache_config { # forces replacement
              + enabled = false
            }
```

jmymy:
according to the repo, the fix didnt make it in the v3.17.0 release 
https://github.com/terraform-providers/terraform-provider-google-beta/blob/v3.17.0/google-beta/resource_container_cluster.go#L263

sebastianortizs4n:
Ohh man, sorry about that. What I did was enabled it through the gcloud web, changed the config to true, removed the cluster from the state and import it again.

morgante:
[...]

PR #528, opened 2020-05-21
Add KALM config

c0ffeec0der:
Add KALM config

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -160,6 +160,10 @@ resource "google_container_cluster" "primary" {
         enabled = gce_persistent_disk_csi_driver_config.value.enabled
       }
     }
+
+    kalm_config {
+      enabled = var.kalm_config
+    }
     {% endif %}
   }
 
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -423,6 +423,12 @@ variable "gce_pd_csi_driver" {
   default     = false
 }
 
+variable "kalm_config" {
+  type        = bool
+  description = "(Beta) Whether KALM is enabled for this cluster."
+  default     = false
+}
+
 variable "database_encryption" {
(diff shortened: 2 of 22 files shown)


c0ffeec0der:
Okay, beta provider updated in /autogen/version.tf.tmpl

bharathkkb:
@c0ffeec0der you will need to update the examples which use the beta-modules as well. If you do `make docker_test_lint` that should show you which modules are failing validation due to this.

c0ffeec0der:
I ran make docker_test_lint and it found a lot of errors (attached). But i think most is not related to the change, rather to windows character formatting. So I did some guesswork here and think the modules in examples folder get the beta version from test/setup/versions.tf . I have updated the file. Is the test on your end still failing? If you could supply the result of make docker_test_lint on your end, I'll make the changes in the code

[test_lint_error.txt](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/files/4666014/test_lint_error.txt)

bharathkkb:
Sure, here are the logs: https://gist.github.com/bharathkkb/f1298ef2629c1ce0f25f30fea3de0c71
Basically any example like `./examples/node_pool ` you see that has the error `No provider "google-beta" plugins meet the constraint` is failing because we have the required beta provider to be at least `3.21.0` but those examples are using an older version.

c0ffeec0der:
Ah okay, so I have updated the failing version in /examples

But the ones in /test/fixtures doesn't seem to specify provider version. It looks like it use the same one in /examples?

I see the second trigger of the cloud build successful. Is that the lint check? If there is still error, let me know how to fix the one in /test/fixtures, because I couldn't find the provider version explicitly mentioned in that one. Thanks

bharathkkb:
I apologize for the back and forth, we are working to surface these logs to make contributing easier 😃  And yes, the ones in fixtures don't need them as we have set in the examples.

LGTM overall. However we need to make sure that the issue mentioned here https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/428 does not happen. I think this was fixed in the provider but I am happy to give this a whirl and confirm. 

@morgante I think int tests are failing due to the cloudbuild service account unable to copy the acm yaml from GCS. Could you PTAL?

c0ffeec0der:
No worries, always happy to help :)

PR #534, opened 2020-05-22
Remove incorrect note on whitelisting IPs

redbaron:
Docs on GKE authorized networks dont mention anywhere that IP must belong to Google network.

Ref: https://cloud.google.com/kubernetes-engine/docs/how-to/authorized-networks


redbaron:
@morgante , done

PR #536, opened 2020-05-22
Bugfix/acm param defaults

linde:
some bug fixes for param behavior for acm and configsync. removes shadow default values in these modules and allows module users to forgo params like policy dir and sync branch that have default values in the configmanagement resource.


modules/acm/variables.tf
```diff
@@ -15,17 +15,17 @@
  */
 
 variable "cluster_name" {
-  description = "The unique name to identify the cluster in ACM."
+  description = "Name used to reach cluster by way of a google_container_cluster terraform resource. This value becomes the cluster name in the Config Sync kubernetes custom resource."
```
bharathkkb, line 18: By `...reach cluster by way of a google_container_cluster terraform resource` do you mean that internally we use this cluster_name as an input for the `google_container_cluster` data source? 
`google_container_cluster resource` refers to the [actual GKE resource](https://www.terraform.io/docs/providers/google/r/container_cluster.html), whereas within the crd module we use the `google_container_cluster` [data source](https://www.terraform.io/docs/providers/google/d/google_container_cluster.html). 

And just for my clarification, is there a specific reason that we have to call out we are using x resource/data source?

linde: suggesting it because we gotten the feedback it was surprising that this and the other params were used to connect to the cluster and that the description for this had made it seem it was only for use in the configmanagent resource which (for better or worse) doesnt have to match the GKE cluster name. i am very open to alternatives and will be the first to admit i may be over fitting on that feedback.

morgante: I agree about making the intent clearer, but we probably don't need to call out the specific internals:

```suggestion
  description = "Name used to reach cluster for installing ACM. This value becomes the cluster name in the Config Sync Kubernetes custom resource."
```

modules/acm/variables.tf
```diff
@@ -15,17 +15,17 @@
...
   type        = string
 }
 
 variable "project_id" {
-  description = "The project in which the resource belongs."
+  description = "project_id used to reach cluster by way of a google_container_cluster terraform resource."
```
morgante, line 23:
```suggestion
  description = "The project_id where the cluster is running (used to reach the cluster for installing ACM)."
```

modules/acm/variables.tf
```diff
@@ -15,17 +15,17 @@
...
   type        = string
 }
 
 variable "location" {
-  description = "The location (zone or region) this cluster has been created in."
+  description = "location used to reach cluster by way of a google_container_cluster terraform resource."
```
morgante, line 28:
```suggestion
  description = "The Google Cloud location (region or zone) where the cluster is running (used to reach the cluster for installing ACM)."
```

linde:
hey, sorry -- i made a simplication but missed your helpful suggestion.  can i trouble you to review and if ok, accept the PR; if you'd prefer the language suggested, that's cool and i can merge it in and regen docs and resubmit

PR #542, opened 2020-05-26
fix: Add additional guardrails for disabled workload identity.

### LINKED ISSUE #535: Not possible to disable workload identity /modules/beta-private-cluster v8.1.0
This is the variable
```hcl
variable "identity_namespace" {
  description = "Workload Identity namespace. (Default value of `enabled` automatically sets project based namespace `[project_id].svc.id.goog`)"
  type        = string
  default     = "enabled"
}
```
This is the local variable 
```hcl
cluster_workload_identity_config = var.identity_namespace == null ? [] : var.identity_namespace == "enabled" ? [{
    identity_namespace = "${var.project_id}.svc.id.goog" }] : [{ identity_namespace = var.identity_namespace
  }]
```
I would like to use the variable in this way 
```hcl
identity_namespace = null
```

**Expected:**
if `var.identity_namespace == null` it should use an empty list, like so
```
       workload_identity_config {
           identity_namespace = []
        }
```

**Actual result:**
```
      + workload_identity_config {
          + identity_namespace = "null"
        }
```
It basically converts `null` to `"null"`

morgante:
Test that #535 doesn't occur.

Fixes #535.

### DIFF
--- a/examples/simple_regional_beta/main.tf
+++ b/examples/simple_regional_beta/main.tf
@@ -47,6 +47,10 @@ module "gke" {
   enable_binary_authorization = var.enable_binary_authorization
   pod_security_policy_config  = var.pod_security_policy_config
   release_channel             = "REGULAR"
+
+  # Disable workload identity
+  identity_namespace = null
+  node_metadata      = "UNSPECIFIED"
 }
 
 data "google_client_config" "default" {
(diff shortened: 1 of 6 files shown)


examples/simple_regional_beta/main.tf
```diff
@@ -47,6 +47,10 @@ module "gke" {
   enable_binary_authorization = var.enable_binary_authorization
   pod_security_policy_config  = var.pod_security_policy_config
   release_channel             = "REGULAR"
+
+  # Disable workload identity
+  identity_namespace = null
```
bharathkkb, line 52: Should we update test to check for `"null"` since we added guard rails for that?

morgante: No need, since GKE will not accept a `"null"` as a valid value (ie. creation fails).

PR #547, opened 2020-05-28
Add ConfigConnector configuration option (beta)

redbaron:


redbaron:
@bharathkkb , thank, I did just that, but still no luck :(

bharathkkb:
@redbaron looks like the main issue is that only the beta modules provider version was bumped to `3.23.0`. For integration tests for those modules, we use the examples, so we will need to update the providers for all the examples that use the beta provider. [Example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d1e5d123b54e48afa60515b3ff30d7c4daa1882e/examples/simple_regional_beta/main.tf#L21). 
Here is another example PR that did something similar: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/528

After this, you can also run the lint test locally by doing `make docker_test_lint`.

redbaron:
@bharathkkb , thanks, will do. It's a bit  inconvenient that external contributors can't see CI logs :(

bharathkkb:
@redbaron absolutely agree, we are working on something that will let external contributors see CI logs soon 😃 
Look like a test if failing. We need to add`"configConnectorConfig"=>{}` to list of expected addons for this test, like we have for `kalmConfig`: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d1e5d123b54e48afa60515b3ff30d7c4daa1882e/test/integration/beta_cluster/controls/gcloud.rb#L58

redbaron:
@bharathkkb , any idea what it might be this time? :)

bharathkkb:
error doesnt seem related, this LGTM

PR #557, opened 2020-06-11
Enhance WI module usability with existing KSA

### LINKED ISSUE #555: workload identity existing sa iam.gke.io/gcp-service-account overwrite
Following the documented approach to use existing sa would result in an error 

```
resource "kubernetes_service_account" "iden-test" {
  metadata {
    name = "iden-test"
    annotations = {
      "iam.gke.io/gcp-service-account": "iden-test@myproject.iam.gserviceaccount.com"
    }
  }
  secret {
    name = my-custom-secret
  }
}

module "workload_identity" {
  source              = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
  project_id          = var.project
  name                = "iden-test"
  use_existing_k8s_sa = true
}
```

`Output: error: --overwrite is false but found the following declared annotation(s): 'iam.gke.io/gcp-service-account' already has a value`

I can see two options to fix this:

1. add --overwrite parameter to module "annotate-sa", which would require users to implement lifecycle ignore annotations
2. remove module "annotate-sa" completely (and have users define their own annotations as per documentation)

I would prefer 2. approach to keep user defined terraform files clean

kubectl version: v1.18.2
cluster version: v1.16.8-gke.15

bharathkkb:
fixes #555 
- Adds kubectl_wrapper
- Adds overwrite flag for annotate command
- Update docs

### DIFF
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -19,9 +19,23 @@ locals {
   gcp_sa_email            = google_service_account.cluster_service_account.email
 
   # This will cause terraform to block returning outputs until the service account is created
-  k8s_given_name       = var.k8s_sa_name != null ? var.k8s_sa_name : var.name
-  output_k8s_name      = var.use_existing_k8s_sa ? local.k8s_given_name : kubernetes_service_account.main[0].metadata[0].name
-  output_k8s_namespace = var.use_existing_k8s_sa ? var.namespace : kubernetes_service_account.main[0].metadata[0].namespace
+  k8s_given_name         = var.k8s_sa_name != null ? var.k8s_sa_name : var.name
+  output_k8s_name        = var.use_existing_k8s_sa ? local.k8s_given_name : kubernetes_service_account.main[0].metadata[0].name
+  output_k8s_namespace   = var.use_existing_k8s_sa ? var.namespace : kubernetes_service_account.main[0].metadata[0].namespace
+  token                  = var.use_existing_k8s_sa ? data.google_client_config.default.0.access_token : ""
+  cluster_ca_certificate = var.use_existing_k8s_sa ? data.google_container_cluster.primary.0.master_auth.0.cluster_ca_certificate : ""
+  cluster_endpoint       = var.use_existing_k8s_sa ? "https://${data.google_container_cluster.primary.0.endpoint}" : ""
+}
+
+data "google_container_cluster" "primary" {
+  count    = var.use_existing_k8s_sa ? 1 : 0
+  name     = var.cluster_name
+  project  = var.project_id
+  location = var.location
+}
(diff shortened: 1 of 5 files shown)


modules/workload-identity/main.tf
```diff
@@ -19,9 +19,23 @@ locals {
...
-  output_k8s_namespace = var.use_existing_k8s_sa ? var.namespace : kubernetes_service_account.main[0].metadata[0].namespace
+  k8s_given_name         = var.k8s_sa_name != null ? var.k8s_sa_name : var.name
+  output_k8s_name        = var.use_existing_k8s_sa ? local.k8s_given_name : kubernetes_service_account.main[0].metadata[0].name
+  output_k8s_namespace   = var.use_existing_k8s_sa ? var.namespace : kubernetes_service_account.main[0].metadata[0].namespace
+  token                  = var.use_existing_k8s_sa ? data.google_client_config.default.0.access_token : ""
+  cluster_ca_certificate = var.use_existing_k8s_sa ? data.google_container_cluster.primary.0.master_auth.0.cluster_ca_certificate : ""
+  cluster_endpoint       = var.use_existing_k8s_sa ? "https://${data.google_container_cluster.primary.0.endpoint}" : ""
+}
+
+data "google_container_cluster" "primary" {
```
morgante, line 30: Should we use the auth submodule?

bharathkkb:
@morgante WDYT about adding an additional invocation of the WI module in the current example to test the `use_existing_ksa` control flow?

modules/workload-identity/scripts/kubectl_wrapper.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
morgante, line 1: Should we move this into a common location? Maybe the gcloud module?

morgante:
> @morgante WDYT about adding an additional invocation of the WI module in the current example to test the use_existing_ksa control flow?

Good idea, let's add it.

modules/workload-identity/main.tf
bharathkkb: I decided against the auth submodule because we instantiate this optionally if `use_existing_k8s_sa` is true.

modules/workload-identity/scripts/kubectl_wrapper.sh
bharathkkb: I think thats a great idea, considering we now have copies here, in the crd module, asm module etc. Do you mean as a gcloud submodule or did you have another approach? Also might it might better to do it alongside #404

morgante: Yeah, maybe a `gcloud` submodule?

PR #538, opened 2020-05-23
Add ASM module

bharathkkb:
- Adds ASM module
    - Downloads and installs ASM
    - Produces ASM resource patch yamls
    - Optionally enables registration to GKE Hub
- Adds example with ASM module
- Adds tests for ASM module

TODO:
- [x] Add readme
- [x] Add kustomize to tools image - GoogleCloudPlatform/cloud-foundation-toolkit/pull/698

PR #571, opened 2020-06-22
Add bool option for automount_service_account_token

davidholsgrove:
Allow setting kubernetes service account resource attribute `automount_service_account_token`


davidholsgrove:
Thanks @bharathkkb - I see one of the checks timed out, does it need to be rerun?

bharathkkb:
fix for lint test timing out is in #538 so I think this should be okay.

PR #574, opened 2020-06-23
Fix integration test project creation

bharathkkb:
- fixes integration test project creation failure due to different PF versions
- add more random chars to project id to prevent overlap with previous projects pending deletion


morgante:
@bharathkkb Can you also fix #578 and get this PR back to green please?

PR #577, opened 2020-06-23
Remove file function wrapper from ssh_auth_key variable

### LINKED ISSUE #576: Unable to use terraform data resource as ssh_auth_key value
Hi!

I'm in the market to use the Config Sync (9.3.0) submodule with my clusters and I want to be able to pull an ssh key from Google secrets manager in the terraform plan (via the `google_secret_manager_secret_version` data resource). Right now, it looks like the module forces the input to be a file on the local system. 

Does anyone have any suggestions on how to accomplish this? Or would this need a change in the modules to be supported

dpatrick:
Fixes #576 

I looked back at the git history and found this regression was introduced when the config sync module was added, though it applies to the acm module as well.


bharathkkb:
@morgante do you know what those extra triggers are? I see that in the main build too where both int and lint are green.

morgante:
> @morgante do you know what those extra triggers are? I see that in the main build too where both int and lint are green.

It's the same triggers, but it looks like Cloud Build has yet again chained the names they send back.

PR #584, opened 2020-06-27
added variable for service dependency in binary_authorization sub module

### LINKED ISSUE #575: Services in Binary Authorization modules
Binary authorization module is enabling needed services in project with following two parameters using default value true:

disable_services_on_destroy
disable_dependent_services

This will cause problem if there are other attestors in same project. Is it possible to change the code in 
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/binary-authorization/main.tf

**from:**

```
module "project-services" {
  source  = "terraform-google-modules/project-factory/google//modules/project_services"
  version = "~> 8.0"

  project_id = var.project_id

  activate_apis = local.required_enabled_apis
}

```
**To:**

```
module "project-services" {
  source  = "terraform-google-modules/project-factory/google//modules/project_services"
  version = "~> 8.0"

  project_id = var.project_id
  activate_apis = local.required_enabled_apis
  disable_services_on_destroy = false
  disable_dependent_services  = false
} 
```

imrannayer:
Fix #575


modules/binary-authorization/variables.tf
```diff
@@ -34,3 +34,15 @@ variable crypto-algorithm {
   default     = "RSA_SIGN_PKCS1_4096_SHA512"
   description = "Algorithm used for the async signing keys"
 }
+
+variable "disable_services_on_destroy" {
+  description = "Whether project services will be disabled when the resources are destroyed. https://www.terraform.io/docs/providers/google/r/google_project_service.html#disable_on_destroy"
+  default     = "false"
```
morgante, line 40: These should be booleans.

imrannayer: Done. I was thinking about making them bool but kept it string as these variables are defined as string in sub-module project_services

https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/project_services

morgante:
You also need to regenerate docs (`make generate_docs`).

Issue #585, opened 2020-07-01
PodSecurityPolicy implementation changing the cluster on every iteration


msgongora:
**Summary:**
_pod_security_policy_config_ with default value is causing to update the cluster every time you run terraform using `beta-private-cluster-update-variant`. I was able to verify that this issue came from the provider itself but since I'm facing it because the way it was implemented in this module, I'm reporting it here.

**How to reproduce it:**
``` HCL
module "cicd_cluster" {
 source                       = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster-update-variant"
  version                      = "v8.1.0"
  project_id                   = var.project
  region                       = var.region
  service_account              = var.service_account
  name                         = local.cluster_name
  regional                     = false
  zones                        = local.zones
  network                      = local.primary_network
  subnetwork                   = local.primary_subnetwork
  ip_range_pods                = local.cicd_pods_range
  ip_range_services            = local.cicd_services_range
  release_channel              = local.release_channel
  master_ipv4_cidr_block       = local.master_ipv4_cidr_block
  cluster_resource_labels      = local.cluster_labels
  remove_default_node_pool     = true
  enable_private_nodes         = true
  create_service_account       = false
  node_metadata                = "GKE_METADATA_SERVER"
  identity_namespace           = "enabled"
  authenticator_security_group = local.authenticatorsg_email
  maintenance_start_time       = "07:00"
  node_pools = [
    {
      name               = "main-pool"
      machine_type       = local.main_instance_type
      disk_size_gb       = local.main_disk_size
      disk_type          = local.main_disk_type
      min_count          = 1
      max_count          = 20
      image_type         = "COS"
      auto_repair        = true
      auto_upgrade       = true
      preemptible        = false
      initial_node_count = 1
    },
  ]

  node_pools_taints = {
    all       = []
    main-pool = []
  }
}
```
```shell
$ terraform init 

Initializing provider plugins...
- Checking for available provider plugins...
- Downloading plugin for provider "google-beta" (terraform-providers/google-beta) 3.14.0...
- Downloading plugin for provider "kubernetes" (hashicorp/kubernetes) 1.11.3...
- Downloading plugin for provider "random" (hashicorp/random) 2.2.1...
- Downloading plugin for provider "null" (hashicorp/null) 2.1.2...
- Downloading plugin for provider "google" (hashicorp/google) 3.14.0...
...
...

$ terraform apply

[...]

PR #601, opened 2020-07-13
Add master_global_access_config

### LINKED ISSUE #579: How to enable master global access
Hi Folks,

I want to enable master global access in a cluster, I checked the code but couldn't find the input to enable master_global_access.

Can someone please help me to add this in configuration.

Thanks in advance

ilkelma:
Fixes #579  and implements https://www.terraform.io/docs/providers/google/r/container_cluster.html#master_global_access_config

I ran the integration tests and got errors because the examples pin the google provider versions to `~> 3.16.0` which does not include the new `master_global_access_config` block.

I have a stashed commit I can push that bumps all those example versions to the latest `3.29.0` but I didn't know whether that should be part of this commit or there are other things that need to fall in line to bump all those versions.

### DIFF
--- a/modules/beta-private-cluster/variables.tf
+++ b/modules/beta-private-cluster/variables.tf
@@ -397,7 +397,7 @@ variable "master_global_access_enabled" {
   type        = bool
   description = "(Beta) Whether the cluster master is accessible globally (from any region) or only within the same region as the private endpoint."
 
-  default = false
+  default = true
 }
 
 
--- a/modules/beta-private-cluster/variables.tf
+++ b/modules/beta-private-cluster/variables.tf
@@ -397,7 +397,7 @@ variable "master_global_access_enabled" {
   type        = bool
   description = "(Beta) Whether the cluster master is accessible globally (from any region) or only within the same region as the private endpoint."
 
-  default = true
+  default = false
 }
 
 
--- a/modules/beta-private-cluster/variables.tf
+++ b/modules/beta-private-cluster/variables.tf
@@ -393,6 +393,17 @@ variable "master_ipv4_cidr_block" {
   default     = "10.0.0.0/28"
 }
 
+variable "master_global_access_config" {
+  type        = list(object({ enabled = bool }))
+  description = "(Beta) Whether the cluster master is accessible globally (from any region) or only within the same region as the private endpoint. "
+
+  default = [{
+    enabled = true
+  }]
+}
+
+
+
 variable "istio" {
   description = "(Beta) Enable Istio addon"
   default     = false
--- a/modules/beta-private-cluster/variables.tf
+++ b/modules/beta-private-cluster/variables.tf
@@ -393,15 +393,12 @@ variable "master_ipv4_cidr_block" {
   default     = "10.0.0.0/28"
 }
 
-variable "master_global_access_config" {
-  type        = list(object({ enabled = bool }))
-  description = "(Beta) Whether the cluster master is accessible globally (from any region) or only within the same region as the private endpoint. "
+variable "master_global_access_enabled" {
+  type        = bool
+  description = "(Beta) Whether the cluster master is accessible globally (from any region) or only within the same region as the private endpoint."
 
-  default = [{
-    enabled = true
-  }]
+  default = true
 }
-
 
 
 variable "istio" {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Extending the adopted spec, each change should have a link to its corresponding pull request appended.
 
+## [10.0.0](https://www.github.com/terraform-google-modules/terraform-google-kubernetes-engine/compare/v9.4.0...v10.0.0) (2020-07-10)
+
+
+### ⚠ BREAKING CHANGES
+
+See the [upgrade guide](./docs/upgrading_to_v10.0.md) for details.
+
+* The default machine type has been changed to `e2-medium`. If you want the old default, you should specify it explicitly: `machine_type = "n1-standard-2"`.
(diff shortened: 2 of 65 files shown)


ilkelma:
> Thanks for the PR! I think it makes sense to follow this approach [#585 (comment)](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/585#issuecomment-652753906) to just use a boolean flag for this. Bumping dependencies as necessary also sgtm.

Done - I've bumped all versions in test files + examples - some may not be strictly necessary and I can pare back to only the ones that involve a beta private cluster (and thus would need this change) if needed.

morgante:
@ilkelma I don't think we need to upgrade the test version—the beta examples are already in 3.29. Do you mind reverting that?

ilkelma:
looks like the integration tests succeed if I upgrade everything but fail otherwise so something needs upgrading it would seem but i'm not sure exactly where... is it possible y'all could let me know? Likely would be faster than me trying to run the tests on my end

morgante:
@ilkelma The error is on the `simple-regional-with-networking` example:

```
       Error: Invalid template interpolation value
       
         on ../../../cluster.tf line 261, in module "gcloud_wait_for_cluster":
        261:   create_cmd_body        = "${var.project_id} ${var.name}"
           |----------------
           | var.project_id is null
       
       The expression result is null. Cannot include a null value in a string
       template.
```

ilkelma:
> @ilkelma The error is on the `simple-regional-with-networking` example:
> 
> ```
>        Error: Invalid template interpolation value
>        
>          on ../../../cluster.tf line 261, in module "gcloud_wait_for_cluster":
>         261:   create_cmd_body        = "${var.project_id} ${var.name}"
>            |----------------
>            | var.project_id is null
>        
>        The expression result is null. Cannot include a null value in a string
>        template.
> ```

I'm not sure how this relates to my changes 🤔 i'll see if I can reproduce in local integration tests for some clues...

morgante:
> I'm not sure how this relates to my changes 🤔 i'll see if I can reproduce in local integration tests for some clues...

Yeah I'm not sure either but we can see that it's only happening on your branch. I'm especially confused why the GKE version upgrade would affect it.

bharathkkb:
@ilkelma @morgante its on the prepare step actually. The projects are not getting created. Seems like API activations failure. It's not related to this PR as I am seeing it in other modules as well.
```
Error: Error enabling the Compute Engine API required to delete the default network: Error waiting for Enable Project "ci-gke-b8681adb-e462" Services: [compute.googleapis.com]: timeout while waiting for state to become 'done: true' (last state: 'done: false', timeout: 4m0s) 

  on .terraform/modules/gke-project-2/terraform-google-project-factory-8.0.1/modules/core_project_factory/main.tf line 96, in resource "google_project" "main":
  96: resource "google_project" "main" {



Error: Error enabling the Compute Engine API required to delete the default network: Error waiting for Enable Project "ci-gke-asm-b8681adb-03ca" Services: [compute.googleapis.com]: timeout while waiting for state to become 'done: true' (last state: 'done: false', timeout: 4m0s) 

  on .terraform/modules/gke-project-asm/terraform-google-project-factory-8.0.1/modules/core_project_factory/main.tf line 96, in resource "google_project" "main":
  96: resource "google_project" "main" {

```

ilkelma:
@bharathkkb @morgante I bumped the version from 3.25 -> 3.29 in the prep step (the versions.tf file in test/setup) on a hunch but i can't see the results. It got about 20 more minutes in I think so possibly got past that point?

bharathkkb:
I think https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/601/commits/f6096e84f94f7ca6aeb8a2f6f1e792bd121c3ad4 was flaky https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/601/commits/1695094d39d6f9718965e026f371b5c67efa808b has passed.

modules/beta-private-cluster/variables.tf
```diff
@@ -393,6 +393,14 @@ variable "master_ipv4_cidr_block" {
   default     = "10.0.0.0/28"
 }
 
+variable "master_global_access_enabled" {
+  type        = bool
+  description = "(Beta) Whether the cluster master is accessible globally (from any region) or only within the same region as the private endpoint."
+
+  default = true
```
bharathkkb, line 400: I dont have strong feelings but since provider default if false, I wonder if we should keep it that way. WDYT @ilkelma @morgante

modules/beta-private-cluster/variables.tf
```diff
@@ -393,6 +393,14 @@ variable "master_ipv4_cidr_block" {
   default     = "10.0.0.0/28"
 }
 
+variable "master_global_access_enabled" {
```
bharathkkb, line 396: Can we expose this via the [safer-cluster](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/autogen/safer-cluster) too?

modules/beta-private-cluster/variables.tf
ilkelma: Yeah that's my bad, we should definitely match the provider default imho. Will push a fix.

morgante: Actually I like defaulting to true. I think it's the best practice.

modules/beta-private-cluster/variables.tf
morgante: Actually let's not over-parameterize. Let's just set it to true on the safer cluster.

ilkelma: ok sounds good

modules/beta-private-cluster/variables.tf
ilkelma: yeah there's kinda bad feels in either direction "why can't I connect to my cluster?" vs "why did something get set that I wasn't expecting". But given that modules are generally collections of best practices I can see the logic in avoiding the former problem as the latter will likely be much rarer. I'll push a fix flipping it back.

PR #615, opened 2020-07-29
11.0 upgrade guide

bharathkkb:
Adds upgrade guide for 11.0 release

### DIFF
--- a/docs/upgrading_to_v11.0.md
+++ b/docs/upgrading_to_v11.0.md
@@ -40,3 +40,8 @@ module "asm" {
 -  use_tf_google_credentials_env_var = true
 }
 ```
+
+### Workload Identity
+
+- Use of Workload Identity module without `use_existing_k8s_sa` enabled is no-op as it does not trigger the annotation of existing KSA.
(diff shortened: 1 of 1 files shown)


docs/upgrading_to_v11.0.md
```diff
@@ -40,3 +40,8 @@ module "asm" {
 -  use_tf_google_credentials_env_var = true
 }
 ```
+
+### Workload Identity
+
+- Use of Workload Identity module without `use_existing_k8s_sa` enabled is no-op as it does not trigger the annotation of existing KSA.
+- Use of Workload Identity module with `use_existing_k8s_sa` enabled is no-op due to usage of `--overwrite` flag while annotating KSA.
```
morgante, line 47: These two lines seem to conflict with each other?

bharathkkb: @morgante I meant to say that if they are not using the `use_existing_k8s_sa` it does not affect them at all, other than the random cache being removed which is no-op. 

If they are using `use_existing_k8s_sa` then also it is technically no-op, as it will just re-trigger the `kubectl annotate` command which will succeed as we have the ` --overwrite` flag.

morgante: Got it, thanks! I missed the "without" part.

Issue #560, opened 2020-06-12
Set auto_provisioning_defaults.service_account ? (beta-public)


lemaral:
When enabling node autoprovisioning, the service account for it stays at "default" and nodes in "automatic" node pools have no rights on Google APIs (e.g. pull from private GCR)
I would have expected that the service account would be set to the same one that is created for the "regular" node pools so that pods scheduled or rescheduled on the automatic node pools would work the same (at least by default).
I believe it is in auto_provisioning_defaults.service_account

morgante:
Yes we should fill in this property: https://www.terraform.io/docs/providers/google/r/container_cluster.html#service_account

Happy to review a PR.

morgante:
Fixed by #639.

PR #639, opened 2020-08-21
Set auto_provisioning_defaults.service_account

dpetersen:
This sets the Service Account that should be used by node VMs created by
node auto-provisioning. This should cause the auto-provisioned nodes to
have the same permissions as the nodes that are manually provisioned.

This is to address #560. I will be testing this in my cluster that has NAP enabled. As far as I can tell, it's safe to set this value whether auto-provisioning is actually enabled or disabled, and that using `local.service_account` is the right default. That's the default for any manual `node_pool` that isn't specifically overridden.

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -79,9 +79,13 @@ resource "google_container_cluster" "primary" {
   cluster_autoscaling {
     enabled = var.cluster_autoscaling.enabled
 {% if beta_cluster %}
-    auto_provisioning_defaults {
-      service_account = local.service_account
-      oauth_scopes = local.node_pools_oauth_scopes["all"]
+    dynamic "auto_provisioning_defaults" {
+      for_each = var.cluster_autoscaling.enabled ? [1] : []
+
+      content {
+        service_account = local.service_account
+        oauth_scopes = local.node_pools_oauth_scopes["all"]
+      }
     }
     autoscaling_profile = var.cluster_autoscaling.autoscaling_profile != null ? var.cluster_autoscaling.autoscaling_profile : "BALANCED"
 {% endif %}
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -77,8 +77,11 @@ resource "google_container_cluster" "primary" {
 {% endif %}
 
   cluster_autoscaling {
-    enabled             = var.cluster_autoscaling.enabled
+    enabled = var.cluster_autoscaling.enabled
 {% if beta_cluster %}
+    auto_provisioning_defaults {
+      service_account = local.service_account
+    }
     autoscaling_profile = var.cluster_autoscaling.autoscaling_profile != null ? var.cluster_autoscaling.autoscaling_profile : "BALANCED"
 {% endif %}
     dynamic "resource_limits" {
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -81,6 +81,7 @@ resource "google_container_cluster" "primary" {
 {% if beta_cluster %}
     auto_provisioning_defaults {
       service_account = local.service_account
+      oauth_scopes = var.cluster_autoscaling.oauth_scopes
     }
     autoscaling_profile = var.cluster_autoscaling.autoscaling_profile != null ? var.cluster_autoscaling.autoscaling_profile : "BALANCED"
 {% endif %}
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -81,7 +81,7 @@ resource "google_container_cluster" "primary" {
 {% if beta_cluster %}
     auto_provisioning_defaults {
       service_account = local.service_account
-      oauth_scopes = var.cluster_autoscaling.oauth_scopes
+      oauth_scopes = local.node_pools_oauth_scopes["all"]
     }
     autoscaling_profile = var.cluster_autoscaling.autoscaling_profile != null ? var.cluster_autoscaling.autoscaling_profile : "BALANCED"
 {% endif %}
--- a/autogen/main/sa.tf.tmpl
+++ b/autogen/main/sa.tf.tmpl
@@ -23,7 +23,7 @@ locals {
       ["dummy"],
     ),
   )
-  // if user set var.service_accont it will be used even if var.create_service_account==true, so service account will be created but not used
+  // if user set var.service_account it will be used even if var.create_service_account==true, so service account will be created but not used
   service_account = (var.service_account == "" || var.service_account == "create") && var.create_service_account ? local.service_account_list[0] : var.service_account
 }
 
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -225,7 +225,6 @@ variable "cluster_autoscaling" {
     max_cpu_cores       = number
     min_memory_gb       = number
     max_memory_gb       = number
-    oauth_scopes        = list(string)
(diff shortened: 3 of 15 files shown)


dpetersen:
I have rebased this with master and also added one additional overridable attribute for node auto-provisioning that came up in my testing. I noticed that my NAP (node auto-provisioning) nodes could not pull images from my private registry, despite running as a service account that had access. I realized that those nodes had different oAuth scopes to my manual node:

```
https://www.googleapis.com/auth/logging.write
https://www.googleapis.com/auth/monitoring
```
my other pool, set up with this module, had the module's default:
```
https://www.googleapis.com/auth/cloud-platform
```

The `logging.write` + `monitoring` scopes came from GKE defaults, as far as I can tell. I have left those as the default in my additions to this module but allowed it to be overridden. I'm not sure if the stance of this module is to provide suggestions or to just stick with GKE defaults. Personally, I would be less confused if the default for all node pools created were identical, and the other nodes use the full `cloud-platform` scope. So I have GKE defaults in right now, but I'm certainly happy to change it.

morgante:
@dpetersen If scopes are not provided, we should probably pick up whatever scopes are specified on the existing node pools (maybe the first node pool), which also default to `cloud-platform`. Does that make sense?

dpetersen:
That makes sense, I will try and figure it out. Unfortunately, I'm running into a few unrelated issues here. First, I think #676 is breaking my local `terraform apply`. I will make a separate issue for that and un-rebase this branch for now. I also cannot see the lint output from this failing CI check, so I don't know what to fix. I can see the lint output from my other PR, #682, and it is complaining about code I didn't change.

dpetersen:
Well, I learned something new today. The `cluster_autoscaling` object I have added the `oauth_scopes` to cannot have any optional fields.

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/dffb04774f81cd0a9e2459c411eb75cde8b705d2/autogen/main/variables.tf.tmpl#L211-L229

There is a [long-running, open Github issue](https://github.com/hashicorp/terraform/issues/19898) on the topic, but if I add `oauth_scopes` to this list, it becomes a required field on that object. If you want to override any of these fields, you have to override them all.

It probably doesn't make sense for oAuth scopes to be required to enable NAP. And this isn't the last NAP-related override that should probably be optional. At some point, surge upgrade settings for NAP [will be supported in the provider](https://github.com/hashicorp/terraform-provider-google/pull/5676), and that should certainly be optional. I think you could probably make the argument that `autoscaling_profile` should be optional, too.

So the scenario your described above is not possible with this configuration, you can't enable NAP and not specify your oAuth scopes. I suppose you could specify an empty list, but that would be confusing to use. I think I should probably make a new, separate variable, like `cluster_autoscaling_oauth_scopes` or something. Does that reasonable? I'm not a particularly experienced Terraform module author, so I'm not sure what other tricks may exist to keep this ergonomic.

morgante:
Thanks for investigating and I had forgotten the issue around defaults for objects.

In this case, I think we actually have a nice solution: simply reference the "all" value for `var.node_pools_oauth_scopes` which we have [already defined](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/5f5c8d895933693073969d63372d6ce6580fb226/modules/private-cluster-update-variant/variables_defaults.tf#L53). This gives a sensible default and it makes sense that auto-provisioned node pools should have scopes matching "all" your node pools.

dpetersen:
That makes sense and works great for me. I have updated the PR and removed that variable.

dpetersen:
Ugh, I found one additional issue. I set up a minimal cluster to test one of my other PRs, and that cluster did not have NAP enabled. But the plan would continually say it wanted to change my cluster with:
```
~ cluster_autoscaling {
      autoscaling_profile = "BALANCED"
      enabled             = false

    + auto_provisioning_defaults {
        + oauth_scopes    = [
            + "https://www.googleapis.com/auth/cloud-platform",
          ]
        + service_account = "tf-gke-prtest-sde5@content-don-5ea5.iam.gserviceaccount.com"
      }
  }
```
It would apply cleanly, but then continue to have the same desired changes over and over. So I think maybe the API doesn't save your auto-provisioning defaults when autoscaling is disabled. That code for a conditional block (which I stole from [a medium article](https://medium.com/@business_99069/terraform-0-12-conditional-block-7d166e4abcbf)) feels like a hack, but it does seem to work.

dpetersen:
Hey everybody, I'm back to pester you about this pull request. We have been running this for a couple of months, but we're upgrading to Terraform 0.14 so I have rebased this pull request on `master` (which was the same as `v12.3.0` at the moment I pulled and rebased). I am testing the updated version of this pull request against on our staging clusters (it was previously based on `v12.0.0`, I believe). Rebasing doesn't appear to have changed anything, which is great.

As far as I know, I have implemented the requested changes from your feedback. Ever since I first created this pull request, CI has been failing. Unfortunately, I don't have rights to view the failure to see what the issue is.

Please let me know if there's anything I can do to get this in a release version of the module. Thanks so much, this module saves us a ton of time!

aaronwebre (approved):

morgante (changes requested):
Sorry for the delayed response.

The failing test is:
```
     expected: {"autoprovisioningNodePoolDefaults"=>{"oauthScopes"=>["https://www.googleapis.com/auth/logging.write"...inimum"=>"5", "resourceType"=>"cpu"}, {"maximum"=>"30", "minimum"=>"10", "resourceType"=>"memory"}]}
          got: {"autoprovisioningNodePoolDefaults"=>{"oauthScopes"=>["https://www.googleapis.com/auth/cloud-platform...inimum"=>"5", "resourceType"=>"cpu"}, {"maximum"=>"30", "minimum"=>"10", "resourceType"=>"memory"}]}
```

It is configured [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/integration/node_pool/controls/gcloud.rb#L36) with [this fixture](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/fixtures/node_pool/example.tf).

dpetersen:
Oh, thank you. I wish it wasn't truncating the hash like that, but the scopes being different does make sense. I have just pushed up a commit that should fix that, but I don't expect the tests to pass. I'm trying to understand how these specs are run. I'm fairly certain that CI will still fail, but now about the node pool service account.

The specs [expect `"default"`](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/integration/node_pool/controls/gcloud.rb#L41), but that's no longer the case. What it will be now is sort of a mystery to me that I think will be dynamic for each run. It [comes from a `var`](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/fixtures/node_pool/example.tf#L28)  in the fixture. Can I somehow make that available in my tests [like how the Project ID is available](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/integration/node_pool/controls/gcloud.rb#L15)? I'm fairly proficient with Ruby and RSpec from my past, but how these are executing is a bit of a mystery. I can't find any existing tests that are checking the service account elsewhere. I can test everything else and ignore the service account name if you'd rather.

morgante:
@dpetersen Looks like it succeeded after all. :)

FYI for the future you can run tests locally to verify: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#integration-testing

morgante (approved):

dpetersen:
Doh! I went looking for that in the README, I didn't notice the other file. Thanks again, really glad this is in master, and I look forward to the next release. I appreciate all your help with this, and your work on the modules in this ecosystem, they're working great for us.

Issue #661, opened 2020-09-10
[bug] enable_gke_hub_registration variable not respected in asm module


cloud-pharaoh:
#611 introduced an issue where the enable_gke_hub_registration variable is no longer respected and the ASM module will try to register the cluster even when set to false.

I have 2 solutions:
1. Less elegant but compatible with 0.12
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/compare/master...CloudPharaoh:hub_enable_asm?expand=1

2. Requires 0.13
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/compare/master...CloudPharaoh:hub_enable_asm_13?expand=1

cloud-pharaoh:
@bharathkkb can you take a look at these solutions and let me know which one to submit PR for.

morgante:
We're not ready to require 0.13, so your first approach is probably better.

bharathkkb:
Hi @cloud-pharaoh 
We can remove GKE hub from ASM. If users want they can now use the hub submodule directly. There is some context here #611

bharathkkb:
fixed by #665

PR #665, opened 2020-09-10
refactor!:remove hub from asm module

cloud-pharaoh:
address #661

### DIFF
--- a/examples/simple_zonal_with_asm/hub.tf
+++ b/examples/simple_zonal_with_asm/hub.tf
@@ -1,0 +1,23 @@
+/**
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


comment-bot-dev:

bharathkkb (changes requested):
Thanks for the PR @cloud-pharaoh 
We have this [example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/examples/simple_zonal_with_asm/) which is actually an [integration test](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/182dded476ddd8eb6f95a800f7a5bb6541c9fcbe/test/integration/simple_zonal_with_asm/controls/gcloud.rb#L43). To keep those passing, I propose we invoke the hub module within that same example. 

Another alternative is to use a new test just for hub but I dont think thats necessary as we should be consolidating wherever possible.

bharathkkb (approved):
LGTM

cloud-pharaoh:
w00t!

bharathkkb:
@cloud-pharaoh one minor nit. The documentation for `asm` module still has some references to hub like `Optionally registering your cluster with GKE Hub.` We should clean those up too.

cloud-pharaoh:
Good catch. I totally forgot there were words in there apart from the variables and outputs :) 
Done.

PR #669, opened 2020-09-14
fix: Fix skip_provisioners enabled flag for wait_for_cluster

bharathkkb:
IIUC we should set `enabled = false` when `skip_provisioners = true`. Looks like a regression from #404


morgante (approved):

comment-bot-dev:

bharathkkb:
also fixes cloud run int tests due to API change ref: b/159766801

PR #676, opened 2020-09-16
Make wait-for-cluster logic more robust

MrBlaise:
The status is PROVISIONING at the beginning, thus the original logic will always exit and say "Cluster is ready!". If we wait for the RUNNING status, we will know for sure the cluster is up.

Note: if the initial cluster node pool is removed it will say RUNNING for a second then go back to RECONCILING again while the node pool is removed. This might cause issues as well.

### DIFF
--- a/autogen/main/scripts/wait-for-cluster.sh
+++ b/autogen/main/scripts/wait-for-cluster.sh
@@ -27,7 +27,7 @@ echo "Waiting for cluster $CLUSTER_NAME in project $PROJECT to reconcile..."
 
 while
   current_status=$(gcloud container clusters list --project="$PROJECT" --filter=name:"$CLUSTER_NAME" --format="value(status)")
-  [[ "${current_status}" == "RECONCILING" ]]
+  [[ "${current_status}" != "RUNNING" ]]
 do printf ".";sleep 5; done
 
 echo "Cluster is ready!"
--- a/modules/beta-private-cluster-update-variant/scripts/wait-for-cluster.sh
+++ b/modules/beta-private-cluster-update-variant/scripts/wait-for-cluster.sh
@@ -27,7 +27,7 @@ echo "Waiting for cluster $CLUSTER_NAME in project $PROJECT to reconcile..."
 
 while
   current_status=$(gcloud container clusters list --project="$PROJECT" --filter=name:"$CLUSTER_NAME" --format="value(status)")
-  [[ "${current_status}" == "RECONCILING" ]]
+  [[ "${current_status}" != "RUNNING" ]]
 do printf ".";sleep 5; done
 
 echo "Cluster is ready!"
--- a/modules/beta-private-cluster/scripts/wait-for-cluster.sh
+++ b/modules/beta-private-cluster/scripts/wait-for-cluster.sh
@@ -27,7 +27,7 @@ echo "Waiting for cluster $CLUSTER_NAME in project $PROJECT to reconcile..."
 
 while
   current_status=$(gcloud container clusters list --project="$PROJECT" --filter=name:"$CLUSTER_NAME" --format="value(status)")
-  [[ "${current_status}" == "RECONCILING" ]]
+  [[ "${current_status}" != "RUNNING" ]]
 do printf ".";sleep 5; done
 
 echo "Cluster is ready!"
--- a/modules/beta-public-cluster-update-variant/scripts/wait-for-cluster.sh
+++ b/modules/beta-public-cluster-update-variant/scripts/wait-for-cluster.sh
@@ -27,7 +27,7 @@ echo "Waiting for cluster $CLUSTER_NAME in project $PROJECT to reconcile..."
 
 while
   current_status=$(gcloud container clusters list --project="$PROJECT" --filter=name:"$CLUSTER_NAME" --format="value(status)")
-  [[ "${current_status}" == "RECONCILING" ]]
+  [[ "${current_status}" != "RUNNING" ]]
 do printf ".";sleep 5; done
 
 echo "Cluster is ready!"
--- a/modules/beta-public-cluster/scripts/wait-for-cluster.sh
+++ b/modules/beta-public-cluster/scripts/wait-for-cluster.sh
@@ -27,7 +27,7 @@ echo "Waiting for cluster $CLUSTER_NAME in project $PROJECT to reconcile..."
 
 while
   current_status=$(gcloud container clusters list --project="$PROJECT" --filter=name:"$CLUSTER_NAME" --format="value(status)")
-  [[ "${current_status}" == "RECONCILING" ]]
+  [[ "${current_status}" != "RUNNING" ]]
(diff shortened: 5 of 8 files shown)


comment-bot-dev:

bharathkkb (commented):
Thanks for the PR. This makes sense although I am curious. The provider should ensure that we are always past the `PROVISIONING` [state](https://github.com/hashicorp/terraform-provider-google/blob/4d60ab75ceea255d8062db209a407225e7b85b15/google/resource_container_cluster.go#L2235). Did you encounter this particular case where the cluster was provisioned and Terraform exited successfully but the cluster was still in a `PROVISIONING` state?

> if the initial cluster node pool is removed it will say RUNNING for a second then go back to RECONCILING again while the node pool is removed. This might cause issues as well.

[We wait](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/0b5bd3d7a1f079ab98bc294f64a4f6b27ee34fc7/autogen/main/cluster.tf.tmpl#L561) for the new nodepool to be provisioned before running the wait for cluster. Does that help?

MrBlaise:
> Thanks for the PR. This makes sense although I am curious. The provider should ensure that we are always past the `PROVISIONING` [state](https://github.com/hashicorp/terraform-provider-google/blob/4d60ab75ceea255d8062db209a407225e7b85b15/google/resource_container_cluster.go#L2235). Did you encounter this particular case where the cluster was provisioned and Terraform exited successfully but the cluster was still in a `PROVISIONING` state?
> 
> > if the initial cluster node pool is removed it will say RUNNING for a second then go back to RECONCILING again while the node pool is removed. This might cause issues as well.
> 
> [We wait](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/0b5bd3d7a1f079ab98bc294f64a4f6b27ee34fc7/autogen/main/cluster.tf.tmpl#L561) for the new nodepool to be provisioned before running the wait for cluster. Does that help?

You are right, my bad. I used the same wait-for-cluster logic from the module for a different module of mine and I switched up the two and I thought I encountered this error in the gke module. It works well, it only starts checking after the node pool is created.

It might still be worth considering my PR in my opinion because we are waiting for the cluster to be RUNNING rather than waiting for it to be not RECONCILING. At least to me that makes more sense.

bharathkkb (approved):
LGTM
/cc @morgante

morgante (approved):
This logic looks more robust, though I see tests are failing?

bharathkkb:
@morgante ah yes these were not made in autogen. 
@MrBlaise could you make these changes [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/scripts/wait-for-cluster.sh) and run `make build`.

MrBlaise:
@bharathkkb Sure, I’ll do it today :)

MrBlaise:
> @morgante ah yes these were not made in autogen.
> @MrBlaise could you make these changes [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/scripts/wait-for-cluster.sh) and run `make build`.

Should be done.

greg-bumped:
I don't know if it's somehow just me, but this change to make the script more robust breaks the functionality that worked for terraform cloud.  I've tried a variety of workarounds, including "skip_provisioners" and nothing worked.  
Gets stuck in the loop of "gcloud not found". The script needs to revert to previous behavior if it can't find gcloud!
This is not robust enough.

greg-bumped:
The specific line that changed and has caused an infinite loop is:
`  [[ "${current_status}" == "RECONCILING" ]]`
which became 
` [[ "${current_status}" != "RUNNING" ]]`
And in the case where terraform does NOT have access to gcloud (such as in terraform cloud) this causes an infinite loop, because current_status will never be "RUNNING"...

morgante:
@greg-bumped gcloud is a dependency, your issue isn't related to this change.

Please review the update notes about how to handle forcing the install of gcloud: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/docs/upgrading_to_v12.0.md#dropped-support-for-gcloud_skip_download-variable

greg-bumped:
The other modules seem to be interacting with gcloud just fine.  But this script, for some reason beyond me, doesn't seem to be running in a shell that has access to it.  And the nature of terraform cloud isn't such that I can remote to it for diagnostics or to see why.  This issue was swallowed previously, but this fix seems to have brought it to the forefront, and I've had to lock into a previous version.

How can it deploy clusters previously, but it's the script that causes an issue?  Does that still mean that gcloud installation is the issue?  It's getting the service account credentials from an env variable for the other infrastructure that's being deployed.

I apologize for my lack of experience in this regard...  When I did give the script access to gcloud, it generated a different set of errors:

`to select an already authenticated account to use.
.WARNING: Could not open the configuration file: [/home/terraform/.config/gcloud/configurations/config_default].
ERROR: (gcloud.container.clusters.list) You do not currently have an active account selected.
Please run:

  $ gcloud auth login

to obtain new credentials, or if you have already logged in with a
different account:

  $ gcloud config set account ACCOUNT`

Is this a bug in terraform cloud?  I can't emphasize enough that this script is the only failure point.

Thanks for your time.

morgante:
@greg-bumped This script uses `gcloud`, but your error isn't related to the change made in this PR. It has *always* required gcloud.

You need to do two things:
1. Ensure gcloud is installed (either by adding it to your environment or setting the environment variable `GCLOUD_TF_DOWNLOAD="always"`).
2. Ensure gcloud is [authenticated](https://cloud.google.com/docs/authentication/getting-started#setting_the_environment_variable). This can be done by setting the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

greg-bumped:
GOOGLE_APPLICATION_CREDENTIALS is set, otherwise none of the infrastructure would be deployed.
When I set this:
```  source      = "terraform-google-modules/kubernetes-engine/google//modules/private-cluster"
  version     = "~> 11.1.0"
```
Everything works.  Wasn't gcloud always a requirement?  I don't think anything would deploy in the above situation if the two suggestions you made weren't already moot?

The authentication issue seems unique to the script.  Even when I set GCLOUD_TF_DOWNLOAD, the version of gcloud the script runs doesn't seem to be leveraging the GOOGLE_APPLICATION_CREDENTIALS environment variable.  Where I know the rest of the terraform scripting does.  I hope I'm making sense, I can tell I wasn't being clear earlier, I should have mentioned explicitly that I'd done those steps, and that's what caused the auth errors.

I don't think the terraform would get past the first line of configuration without the credentials set, as it uses them to collect state, doesn't it?  How could it even get to the script at all if it wasn't configured?

morgante:
@greg-bumped We changed how gcloud is installed/used so that might be why.

Terraform and `gcloud` actually authenticate slightly differently 

Can you try setting the `CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE` environment variable?

PR #679, opened 2020-09-17
ACM submodule: fix bug when not using `ssh` secret type

cloud-pharaoh:
Module failed as it tried to create `git-creds` secret regardless of secret type. Add enabled flag to k8sop_creds_secret module.

### DIFF
--- a/modules/k8s-operator-crd-support/main.tf
+++ b/modules/k8s-operator-crd-support/main.tf
@@ -63,6 +63,8 @@ resource "tls_private_key" "k8sop_creds" {
 module "k8sop_creds_secret" {
   source                   = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
   version                  = "~> 2.0.2"
+  
+  enabled                  = var.secret_type == ssh ? 1 : 0
   module_depends_on        = [module.k8s_operator.wait]
   skip_download            = var.skip_gcloud_download
   cluster_name             = var.cluster_name
--- a/modules/k8s-operator-crd-support/main.tf
+++ b/modules/k8s-operator-crd-support/main.tf
@@ -24,8 +24,6 @@ locals {
   policy_dir_node               = var.policy_dir != "" ? format("policyDir: %s", var.policy_dir) : ""
   hierarchy_controller_map_node = var.hierarchy_controller == null ? "" : format("hierarchy_controller:\n    %s", yamlencode(var.hierarchy_controller))
   source_format_node            = var.source_format != "" ? format("sourceFormat: %s", var.source_format) : ""
-  secret_kubectl_command        = local.private_key != null ? "kubectl create secret generic ${var.operator_credential_name} -n=${var.operator_credential_namespace} --from-literal=${local.k8sop_creds_secret_key}='${local.private_key}'" : ""
-
 }
 
 module "k8sop_manifest" {
@@ -74,7 +72,7 @@ module "k8sop_creds_secret" {
   project_id               = var.project_id
   service_account_key_file = var.service_account_key_file
 
-kubectl_create_command  = local.secret_kubectl_command
+kubectl_create_command  = local.private_key != null ? "kubectl create secret generic ${var.operator_credential_name} -n=${var.operator_credential_namespace} --from-literal=${local.k8sop_creds_secret_key}='${local.private_key}'" : ""
 kubectl_destroy_command = "kubectl delete secret ${var.operator_credential_name} -n=${var.operator_credential_namespace}"
 }
 
--- a/modules/k8s-operator-crd-support/main.tf
+++ b/modules/k8s-operator-crd-support/main.tf
@@ -16,7 +16,7 @@
 
 locals {
   cluster_endpoint              = "https://${var.cluster_endpoint}"
-  private_key                   = var.ssh_auth_key == null ? tls_private_key.k8sop_creds.private_key_pem : var.ssh_auth_key
+  private_key                   = var.create_ssh_key && var.ssh_auth_key == null ? tls_private_key.k8sop_creds[0].private_key_pem : var.ssh_auth_key
   k8sop_creds_secret_key        = var.secret_type == "cookiefile" ? "cookie_file" : var.secret_type
   should_download_manifest      = var.operator_path == null ? true : false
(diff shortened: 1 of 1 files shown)


comment-bot-dev:
Thanks for the PR! 🚀
- ⚠️check_whitespace
 Failed whitespace check. More details below. 
 ```bash 
Checking for trailing whitespace
./modules/k8s-operator-crd-support/main.tf:66:  
Error: Trailing whitespace found in the lines above.
Checking for missing newline at end of file
 ``` 
- ⚠️check_terraform
 ```bash 
Running terraform fmt
modules/k8s-operator-crd-support/main.tf
--- old/modules/k8s-operator-crd-support/main.tf
+++ new/modules/k8s-operator-crd-support/main.tf
@@ -61,9 +61,9 @@
 }
 
 module "k8sop_creds_secret" {
-  source                   = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
-  version                  = "~> 2.0.2"
-  
+  source  = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
+  version = "~> 2.0.2"
+
   enabled                  = var.secret_type == ssh ? "true" : "false"
   module_depends_on        = [module.k8s_operator.wait]
   skip_download            = var.skip_gcloud_download
 ```

modules/k8s-operator-crd-support/main.tf
```diff
@@ -63,6 +63,8 @@ resource "tls_private_key" "k8sop_creds" {
 module "k8sop_creds_secret" {
   source                   = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
   version                  = "~> 2.0.2"
+  
+  enabled                  = var.secret_type == ssh ? "true" : "false"
```
bharathkkb, line 67:
```suggestion
  enabled                  = var.secret_type == "ssh" ? "true" : "false"
```

bharathkkb (changes requested):
Thanks for the PR!

bharathkkb:
@cloud-pharaoh looks like there maybe some nit formatting issue. `terraform fmt --recursive` should fix those.

cloud-pharaoh:
Hold off on merging this one. I found an issue with template interpolation.

cloud-pharaoh:
@bharathkkb I am not super happy with creating an ssh key and not using but couldn't find a better solution to the interpolation error. Can you take a look and let me know if there is a better way?

Basically if an existing key is used, `local.private_key` was null and the `k8sop_creds_secret` errored out on plan due to an interpolation error.

bharathkkb:
@cloud-pharaoh if an existing key is used and `var.secret_type == "ssh"` shouldn't `ssh_auth_key` be passed in? That would make `local.private_key` not null right? Is there another case I am missing?

cloud-pharaoh:
In existing code:

In case of using a different secret type. Not ssh.
secret_type=“gcenode”
create_ssh_key=false
ssh_auth_key=null
local.private_key=null

Module k8sop_cred_secret.enabled will be false and won’t run but even though that’s the case terraform plan runs into an interpolation error because local.private_key is null.

bharathkkb:
I see
One kinda hacky way would be to add some logic to the expression itself. This will get rid of the need for unnecessarily creating the ssh key.
```hcl
kubectl_create_command  = "kubectl create secret generic ${var.operator_credential_name} -n=${var.operator_credential_namespace} --from-literal=${local.k8sop_creds_secret_key}='${var.secret_type != "ssh" && local.private_key == null ? "" : local.private_key}'"
```

This does seem like a tf limitation. I was able to repro with this minimal config.
```hcl
locals {
  test_var = null
}

resource "random_pet" "foo" {
  count  = local.test_var == "bar" ? 1 : 0
  prefix = "foo-${local.test_var}"
}

```

@morgante any thoughts?

morgante:
What is the underlying error we're encountering here? I don't want to assume that SSH is the only permissable credential type.

It seems like a few fixes are needed:

1. Update the module to only be activated if `var.create_ssh_key = true || var.ssh_auth_key != null`
2. Within the create command, we just need to change the literal to use an inline expression:

```
kubectl_create_command  = "${local.private_key != null ? "kubectl create secret generic ${var.operator_credential_name} -n=${var.operator_credential_namespace} --from-literal=${local.k8sop_creds_secret_key}='${local.private_key}' : ''}"
```

bharathkkb:
@morgante IIUC  the underlying error is we do not want `k8sop_creds_secret` to be activated if we are not creating a k8s secret for example https://cloud.google.com/anthos-config-management/docs/how-to/installing#using_a_google_source_repository_with_workload_identity. @cloud-pharaoh please correct me if I misunderstood.

cloud-pharaoh:
@bharathkkb correct. The secret was created regardless of secret type. After adding the enabled flag to `k8sop_creds_secret` the plan command ran into an interpolation error caused by the local.private_key being null. I am testing now with the inline expression as suggested by Morgante to confirm that will resolve it.

cloud-pharaoh:
@morgante The test with suggested changes yielded the same issue
```
Step #8 - "tf-acm-prod": Error: Invalid template interpolation value
Step #8 - "tf-acm-prod": 
Step #8 - "tf-acm-prod":   on .terraform/modules/acm_gke4/modules/k8s-operator-crd-support/main.tf line 75, in module "k8sop_creds_secret":
Step #8 - "tf-acm-prod":   75:   kubectl_create_command  = "kubectl create secret generic ${var.operator_credential_name} -n=${var.operator_credential_namespace} --from-literal=${local.k8sop_creds_secret_key}='${local.private_key}' : ''}"
Step #8 - "tf-acm-prod":     |----------------
Step #8 - "tf-acm-prod":     | local.private_key is null
Step #8 - "tf-acm-prod": 
Step #8 - "tf-acm-prod": The expression result is null. Cannot include a null value in a string
Step #8 - "tf-acm-prod": template.
Step #8 - "tf-acm-prod": 
Finished Step #8 - "tf-acm-prod" 
```

Variables:
```
create_ssh_key = false
secret_type       = "gcenode"
ssh_auth_key    = null # default
```

morgante:
@cloud-pharaoh Looking good, just need to reformat.

cloud-pharaoh:
@morgante done :)

bharathkkb (approved):
LGTM
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/688 for the issue we discussed offline

PR #705, opened 2020-10-08
feat: node_pool_taints available to all modules

marko7460:
This PR enables node pool taints on all the cluster types , not only beta clusters

### DIFF
--- a/test/setup/main.tf
+++ b/test/setup/main.tf
@@ -20,7 +20,7 @@ resource "random_id" "random_project_id_suffix" {
 
 module "gke-project-1" {
   source  = "terraform-google-modules/project-factory/google"
-  version = "~> 8.0"
+  version = "~> 9.1.0"
 
   name                 = "ci-gke-${random_id.random_project_id_suffix.hex}"
   random_project_id    = true
@@ -39,11 +39,17 @@ module "gke-project-1" {
     "serviceusage.googleapis.com",
     "storage-api.googleapis.com",
   ]
+  activate_api_identities = [
+    {
+      api   = "container.googleapis.com"
+      roles = ["roles/cloudkms.cryptoKeyEncrypterDecrypter", "roles/container.serviceAgent"]
+    },
+  ]
 }
 
 module "gke-project-2" {
   source  = "terraform-google-modules/project-factory/google"
-  version = "~> 8.0"
+  version = "~> 9.1.0"
 
   name                 = "ci-gke-${random_id.random_project_id_suffix.hex}"
   random_project_id    = true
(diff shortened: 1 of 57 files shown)


comment-bot-dev:

test/setup/main.tf
```diff
@@ -56,16 +63,23 @@ module "gke-project-2" {
     "cloudkms.googleapis.com",
     "cloudresourcemanager.googleapis.com",
     "container.googleapis.com",
+    "compute.googleapis.com",
     "pubsub.googleapis.com",
     "serviceusage.googleapis.com",
     "storage-api.googleapis.com",
   ]
+  activate_api_identities = [
```
bharathkkb, line 71: @marko7460 Did you get an error without this? We should only need this if are using this identity cross project (which we are not) or we need to apply additional IAM roles on top of the existing ones (in which case we will need to supply `roles/container.serviceAgent` + whatever additional roles we need.)

test/setup/main.tf
```diff
@@ -35,15 +35,22 @@ module "gke-project-1" {
     "cloudkms.googleapis.com",
     "cloudresourcemanager.googleapis.com",
     "container.googleapis.com",
+    "compute.googleapis.com",
```
bharathkkb, line 38: We removed this on purpose as it was increasing test flakiness. ref: https://github.com/hashicorp/terraform-provider-google/issues/7103

bharathkkb (approved):
Thanks for the PR! 
I do see some odd CI errors unrelated to this change, will take a look at them tomorrow.

test/setup/main.tf
marko7460: I was getting the error ```Step #1 - "prepare": [1m[31mError: [0m[0m[1mRequest "Set IAM Binding for role \"roles/cloudkms.cryptoKeyEncrypterDecrypter\" on \"project \\\"ci-gke-39e386ee-bb70\\\"\"" returned error: Batch request and retried single request "Set IAM Binding for role \"roles/cloudkms.cryptoKeyEncrypterDecrypter\" on \"project \\\"ci-gke-39e386ee-bb70\\\"\"" both failed. Final error: Error applying IAM policy for project "ci-gke-39e386ee-bb70": Error setting IAM policy for project "ci-gke-39e386ee-bb70": googleapi: Error 400: Service account service-520904206997@container-engine-robot.iam.gserviceaccount.com does not exist., badRequest[0m``` . After talking to @morgante I decided to try `active_api_identities` and the error stoped after adding it.

test/setup/main.tf
marko7460: I added it to try to fix the error
```
Error: googleapi: Error 403: Google Compute Engine: Required 'compute.regions.get' permission for 'projects/ci-gke-3ba695b1-b5d9/regions/us-central1'., forbidden
```
Adding the api didn't fix the problem. I will remove compute.googleapis.com

test/setup/main.tf
bharathkkb: ah yes you are correct we do assign `additional` roles to that identity. https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/0bcf3ca0dfebcdd99084d43eb6833bb7e55ae434/test/setup/iam.tf#L109

Issue #644, opened 2020-08-26
Promote previously beta features to GA modules


bharathkkb:
Quite a few beta features are now GA and we should enable these in our GA modules.
TODO:
- release_channel https://github.com/hashicorp/terraform-provider-google/pull/6955
- database_encryption https://github.com/terraform-providers/terraform-provider-google/pull/6757
- workload_identity_config https://github.com/hashicorp/terraform-provider-google/pull/6490
- enable_shielded_nodes https://github.com/terraform-providers/terraform-provider-google/pull/6303
- node_locations
- enable_binary_authorization

mik-laj:
+1

edwardday7:
+1 Would love the workload identity features without having to use the beta module since it's GA now.

PR #706, opened 2020-10-09
Promote previously beta features to GA modules

marko7460:
This PR fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/644

PR #709, opened 2020-10-10
Promote previously beta features to GA modules

marko7460:
This PR fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/644

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ Then perform the following commands on the root folder:
 | default\_max\_pods\_per\_node | The maximum number of pods to schedule per node | string | `"110"` | no |
 | description | The description of the cluster | string | `""` | no |
 | disable\_legacy\_metadata\_endpoints | Disable the /0.1/ and /v1beta1/ metadata server endpoints on the node. Changing this value will cause all node pools to be recreated. | bool | `"true"` | no |
+| enable\_binary\_authorization | Enable BinAuthZ Admission controller | string | `"false"` | no |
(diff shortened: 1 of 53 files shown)


comment-bot-dev:

bharathkkb (commented):
Overall LGTM
Can we implement swap a couple of checks as described https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/706#pullrequestreview-505964436 ?

marko7460:
@bharathkkb I have missed to update the `source` in examples/node_pool. All other non beta examples have been updated to point to non beta modules including workload identity. Maybe I'm not understanding the task. Are there other examples that need updating?

examples/node_pool/variables.tf
```diff
@@ -54,20 +54,22 @@ variable "compute_engine_service_account" {
...
   type = object({
-    enabled             = bool
-    autoscaling_profile = string
-    min_cpu_cores       = number
-    max_cpu_cores       = number
-    min_memory_gb       = number
-    max_memory_gb       = number
+    enabled = bool
+    //Enable when autoscaling_profile becomes GA
+    //autoscaling_profile = string
```
marko7460, line 59: @bharathkkb I commented out `autoscaling_profile` because it is a beta feature. Comment is there so that we know to add this back once it becomes GA. If you think that this is not necessary then I will remove it.

bharathkkb: I think in that case its better to switch back to `"../../modules/beta-public-cluster/"` and keep the test as is. We have a couple of tests for the others in GA and we dont need to be exhaustive.

bharathkkb (approved):
LGTM

Issue #719, opened 2020-10-19
Add support for artifact registry


weeco:
Currently it's possible to set the `registry_project_id` which grants storage.objectViewer so that the GKE service account can pull images from the container registry in the respective project. See: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/beta-private-cluster/variables.tf#L320-L324

If I want to use the artifact registry (which seems to be the successor of the container registry), I do need different IAM permissions - such as the Artifact Registry Reader permission.

bharathkkb:
fixed by https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/748

PR #748, opened 2020-11-27
#719 Add roles/artifactregistry.reader to serviceaccount

davidholsgrove:
Added a new var `grant_artifact_registry_access` but could also see the logic in using the existing `grant_registry_access` and just adding the new `roles/artifactregistry.reader` as well as `roles/storage.objectViewer` to the service accounts.


comment-bot-dev:

morgante (changes requested):
I don't think we need a new variable, let's just reuse grant_registry_access. Thanks for working on this!

morgante (approved):
Thanks for the addition!

PR #752, opened 2020-12-01
feat: added notification_config block to beta submodules

odise:
Added missing `notification_config` parameter block.

Reference: 
* https://github.com/GoogleCloudPlatform/magic-modules/pull/4012
* https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-upgrade-notifications

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -307,6 +307,13 @@ resource "google_container_cluster" "primary" {
     }
   }
 {% endif %}
+
+  notification_config {
+    pubsub {
+      enabled = var.enable_notification_config
+      topic = var.notification_config_topic
+    }
+  }
 }
 
 /******************************************
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -309,7 +309,7 @@ resource "google_container_cluster" "primary" {
 
   notification_config {
     pubsub {
-      enabled = var.enable_notification_config
+      enabled = var.notification_config_topic != "" ? true : false
       topic = var.notification_config_topic
     }
   }
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -306,7 +306,6 @@ resource "google_container_cluster" "primary" {
       security_group = authenticator_groups_config.value.security_group
     }
   }
-{% endif %}
 
   notification_config {
     pubsub {
@@ -314,6 +313,7 @@ resource "google_container_cluster" "primary" {
       topic = var.notification_config_topic
     }
   }
+{% endif %}
 }
 
 /******************************************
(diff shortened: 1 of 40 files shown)


comment-bot-dev:
Thanks for the PR! 🚀
- ⚠️check_terraform
 ```bash 
Running terraform fmt
examples/safer_cluster/main.tf
--- old/examples/safer_cluster/main.tf
+++ new/examples/safer_cluster/main.tf
@@ -62,7 +62,7 @@
   istio    = true
   cloudrun = true
 
-  notification_config_topic = google_pubsub_topic.updates.id
+  notification_config_topic  = google_pubsub_topic.updates.id
   enable_notification_config = true
 }
 
 ```

autogen/main/cluster.tf.tmpl
```diff
@@ -307,6 +307,13 @@ resource "google_container_cluster" "primary" {
     }
   }
 {% endif %}
+
+  notification_config {
```
bharathkkb, line 311: Looks like `notification_config` is in beta, could you move this to into beta conditional block and regenerate? 
```
{% if beta_cluster %}
notification_config {
...
}
{% endif %}
```

autogen/main/variables.tf.tmpl
```diff
@@ -574,3 +574,15 @@ variable "impersonate_service_account" {
   description = "An optional service account to impersonate for gcloud commands. If this service account is not specified, the module will use Application Default Credentials."
   default     = ""
 }
+
+variable "enable_notification_config" {
```
bharathkkb, line 578: variables will also need to be gated

bharathkkb (changes requested):
Thanks for the PR @odise

autogen/main/cluster.tf.tmpl
odise: Done.

autogen/main/variables.tf.tmpl
odise: I also put variables into beta conditional blocks.

autogen/main/cluster.tf.tmpl
```diff
@@ -306,6 +306,13 @@ resource "google_container_cluster" "primary" {
       security_group = authenticator_groups_config.value.security_group
     }
   }
+
+  notification_config {
+    pubsub {
+      enabled = var.enable_notification_config
```
bharathkkb, line 312: Can we do this with a single var? I was thinking we could just utilize `notification_config_topic` like 
```hcl
pubsub{
enabled = var.notification_config_topic != "" ? true : false
topic = var.notification_config_topic
}
```

odise: Absolutely. But technically it is possible to set the topic reference but disable the update itself.

bharathkkb: I don't see an apparent use case that needs disabled but with `notification_config_topic` set. Happy to revisit if needed.

bharathkkb (approved):

PR #753, opened 2020-12-02
Add dataplane-v2 provisioning support

### LINKED ISSUE #656: GKE Dataplane v2 provisioning support
I apologize in advance if this is the wrong place or if this already exits (I may just not be able to find how to do it if so I would appreciate a link to the docs).

Google recently [announced Cilium support in the form of the "dataplane-v2" feature](https://cloud.google.com/blog/products/containers-kubernetes/bringing-ebpf-and-cilium-to-google-kubernetes-engine), it would be great if it was possible to provision a cluster with this feature enabled.

I've never worked with terraform modules but with some hints, I would be happy to add this functionality if it doesn't already exist! :)

ryan-atkins:
Fixes #656

The **datapath_provider** was released in [terraform-provider-google-beta v3.41.0](https://github.com/hashicorp/terraform-provider-google-beta/releases/tag/v3.41.0)

- adds **datapath_provider** var for beta clusters, defaults to "DATAPATH_PROVIDER_UNSPECIFIED"
- as-is, would need to set ```network_policy = false``` whenever desiring to use the "ADVANCED_DATAPATH" provider due to module currently setting CALICO as the default network_policy_provider.

Could use suggestions on how best to handle introducing this and better working with the **network_policy config**.


comment-bot-dev:

examples/simple_regional_beta/main.tf
```diff
@@ -51,6 +51,10 @@ module "gke" {
   # Disable workload identity
   identity_namespace = null
   node_metadata      = "UNSPECIFIED"
+
+  # Support Dataplane Setup
+  network_policy    = var.network_policy
+  datapath_provider = var.datapath_provider
```
bharathkkb, line 57: nit: We can hardcode these in examples instead of plumbing from fixture.
```suggestion
  datapath_provider = "ADVANCED_DATAPATH"
```

bharathkkb (commented):
Thanks for the PR @ryan-atkins
> as-is, would need to set network_policy = false whenever desiring to use the "ADVANCED_DATAPATH" provider due to module currently setting CALICO as the default network_policy_provider

It makes sense to automatically set that for the user if they are explicitly enabling dataplanev2. We should also document this in the variable description. For implementation we would need to check if `ADVANCED_DATAPATH` is set and use that to control the `network_policy` behavior in places below.

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/4a85321214247a1b83e751c45dfd71f4e3c017b1/autogen/main/cluster.tf.tmpl#L145

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/4a85321214247a1b83e751c45dfd71f4e3c017b1/autogen/main/main.tf.tmpl#L82

xaocon:
@bharathkkb, are you working on this already? If not would you feel like maybe doing some mentoring around it? I'd like to see it implemented but I'm still pretty new with Terraform and don't know much about this module. It looks like it's pretty close already but I may need some guidance around network policy.

bharathkkb:
Hi @xaocon 
Feel free to pick this up and happy to help support. I had two comments [above](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/753#pullrequestreview-543429194), one was just hard coding the example. The other was to automatically set `network_policy = false` depending on if `datapath_provider` is `ADVANCED_DATAPATH`

keskiju:
I'm not quite sure if `network_policy` should be automatically disabled if `datapath_provider` is set to `ADVANCED_DATAPATH`. I guess the following configuration is perfectly valid for the [google_container_cluster](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster):

```
  datapath_provider = "ADVANCED_DATAPATH"

  network_policy {
    enabled = true
    provider = "PROVIDER_UNSPECIFIED"
  }
```

You can even leave the network_policy.provider unset since `PROVIDER_UNSPECIFIED` is the default value for it there. I wonder why this module uses `CALICO` as the default instead.

sharkymcdongles:
> I'm not quite sure if `network_policy` should be automatically disabled if `datapath_provider` is set to `ADVANCED_DATAPATH`. I guess the following configuration is perfectly valid for the [google_container_cluster](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster):
> 
> ```
>   datapath_provider = "ADVANCED_DATAPATH"
> 
>   network_policy {
>     enabled = true
>     provider = "PROVIDER_UNSPECIFIED"
>   }
> ```
> 
> I would assume that with these settings it would automatically choose `CILIUM` as the network policy provider. You can even leave the `network_policy.provider` unset since `PROVIDER_UNSPECIFIED` is the default value for it there. I wonder why this module uses `CALICO` as the default `network_policy.provider` instead of `PROVIDER_UNSPECIFIED`. There would be no problem if `PROVIDER_UNSPECIFIED` was the default.

I think you might be incorrect as when I try to deploy a cluster using ADVANCED_DATAPATH and network policy true I get:

`Error: googleapi: Error 400: Enabling NetworkPolicy for clusters with DatapathProvider=ADVANCED_DATAPATH is not allowed., badRequest`

ryan-atkins:
Sorry ignored this for awhile! 
With https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/809 being merged, I believe that resolves needing to touch the `network_policy` here as the default now is false.
Updated the example and reran integration tests 👍

bharathkkb (approved):
Thanks @ryan-atkins

gw0:
Why is this so complicated? Wouldn't it make much more sense to extend `network_policy.provider` by adding `dataplane-v2` (or `cilium`)?

morgante:
@gw0 This was the simplest non-breaking fix, but we could probably re-evaluate interaction with `network_policy_provider`.

PR #756, opened 2020-12-04
Make bash scripts more portable

johnae:
On my distro these scripts won't work since there is no `/bin/bash`, if instead we use `/usr/bin/env bash` for shebang these would work in more places.

Thanks for a great tf module!


comment-bot-dev:

morgante (approved):
Thanks for the PR.

Issue #715, opened 2020-10-16
support for linux_node_config


Schnitzel:
looks like the main terraform provider has support for `linux_node_config` in `node_configs`:
https://github.com/hashicorp/terraform-provider-google/issues/6773

our usecase for this is to set `vm.max_map_count: "262144"` sysctl settings for elasticsearch clusters, while we can do this via init containers, the `sysctl-monitor` of GKE complains about an unmanaged sysctl setting.

bharathkkb:
Thanks for the report @Schnitzel 
SGTM and will be happy to review a PR

m10ev:
Hi, I'm planning to work on this today. Could you assign me?

PR #782, opened 2021-01-09
feat: Linux node config

m10ev:
Not sure if this is the right approach but thought I'd give it a try.

References:
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#linux_node_config
- https://cloud.google.com/sdk/gcloud/reference/beta/container/node-pools/create#--system-config-from-file
- https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/715


comment-bot-dev:

modules/beta-private-cluster/cluster.tf
```diff
@@ -428,6 +428,19 @@ resource "google_container_node_pool" "pools" {
...
+    dynamic "linux_node_config" {
+      for_each = contains(
+        keys(
+          merge(
+            local.node_pools_linux_node_configs["all"],
+        local.node_pools_linux_node_configs[each.value["name"]])),
+      "sysctls") ? [1] : []
+
+      content {
+        sysctls = merge(local.node_pools_linux_node_configs["all"], local.node_pools_linux_node_configs[each.value["name"]])["sysctls"]
```
bharathkkb, line 440: for a config like
```tf
all = {
      sysctls = {
        "foo"  = 1
      }
}
pool-01 = {
      sysctls = {
       "bar" = 2
      }
}
```
Wouldn't merge of the second map override the key `sysctls` from the first unless we do 
```
sysctls = merge(lookup(local.node_pools_linux_node_configs["all"],"sysctls",{}),lookup(local.node_pools_linux_node_configs[each.value["name"]],"sysctls",{}))
```

bharathkkb (commented):
Thanks for the PR @m10ev 
Would be great if we could add a test, maybe [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/examples/node_pool/main.tf)

modules/beta-private-cluster/cluster.tf
m10ev: Yes, you are right. Are you in favour of keeping this structure and doing this more complex "deep" merge or do you prefer to flatten the structure one level because sysctls is currently the only key in this map?

m10ev:
I'm just guessing about the test assertion right now because I don't have an existing cluster in order to see the output format of the describe command. If you have access to view the output from the integration tests, could you help me with that?

modules/beta-private-cluster/cluster.tf
bharathkkb: @m10ev I think it makes sense to flatten as only `sysctls` is available.

test/integration/node_pool/controls/gcloud.rb
```diff
@@ -174,6 +174,22 @@
...
         end
+
+        it "has the expected linux node config sysctls" do
+          expect(data['nodePools']).to include(
+            including(
+              "name" => "pool-01",
+              "config" => including(
+                "linuxNodeConfig" => including(
+                  "sysctls" => including(
+                    "net.core.netdev_max_backlog" => "10000"
```
bharathkkb, line 185:
```suggestion
                    "net.core.netdev_max_backlog" => "10000",
```

bharathkkb (commented):
The tests LGTM, there was a small typo.

test/integration/node_pool/controls/gcloud.rb
bharathkkb: We can probably assert the whole object instead of `including` because there should not be any other keys in `sysctls`.
```ruby
"linuxNodeConfig" => including(
                  "sysctls" => {
                    "net.core.netdev_max_backlog" => "10000",
                    "net.core.rmem_max" => "10000"
                  }
                )
```

morgante: Actually, it's preferable to use `including` as asserting the whole object is more brittle to API changes.

bharathkkb: @morgante In this particular case, wouldn't the API including additional configs within `sysctls` be an error? My reasoning for suggesting whole object for `sysctls` was in case of any bug being introduced in the logic where we merge `all` and nodepool specific config.

morgante: It could be an API error, but it could also be a new default that we haven't included. In either case we couldn't fix it in the module so I'd prefer to avoid brittle logic.

autogen/main/cluster.tf.tmpl
```diff
@@ -543,6 +543,20 @@ resource "google_container_node_pool" "pools" {
...
+    dynamic "linux_node_config" {
+      for_each = merge(
+        local.node_pools_linux_node_configs_sysctls["all"],
+        local.node_pools_linux_node_configs_sysctls[each.value["name"]]
+      ) != {} ? [1] : []
+
+      content {
+        sysctls = merge(
+          local.node_pools_linux_node_configs["all"],
+          local.node_pools_linux_node_configs[each.value["name"]]
```
tenevdev, line 556:
```suggestion
          local.node_pools_linux_node_configs_sysctls["all"],
          local.node_pools_linux_node_configs_sysctls[each.value["name"]]
```

bharathkkb (approved):

tenevdev:
Hello, I'm keen to use this feature, too. Is it blocked waiting for one more approval from the owners or it needs extra work?

bharathkkb:
@tenevdev you can reference the main branch or this will be available with the 0.13 release- #779

PR #809, opened 2021-02-03
Fix: Change default value of network_policy variable to false

### LINKED ISSUE #804: Disable network_policy by default again
First of all: thanks to everyone who is working actively on this repo to improve and maintain it! Keep up the great work!

I'm suggesting to **disable the network_policy option by default again**, reverting PR #356 (linked to issue #138). 

I understand that the CIS GCP Benchmark recommends this policy to be enabled, but this option causes an overhead that can't be neglected in my opinion. As [documented by Google](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#overhead_limitations_and_caveats), this option increases the memory footprint by approximately 128 MB and requires approximately 300 millicores of CPU. Because of this, it is no longer possible to run all `kube-system` pods on a single `e2-small` or `e2-medium` instance. Instead, you need at least 2 `e2-medium` instances, and [Google's documentation](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy#overhead_limitations_and_caveats) even recommends 3 `e2-medium` instances.

This overhead could be acceptable if it would improve the security out-of-the-box, but this option does not contribute anything to the security of your system as long as it's not configured. I'm sure that a lot of users don't know that this option is enabled by default, which means they will probably not configure the `network_policy`, resulting in 300 mCPU and 128 MB of wasted resources (and as previously explained, probably also a wasted VM instance). 

I would finally also like to add that the `network_policy` option is not enabled by default when a Kubernetes cluster is created through Google Cloud Platform console.

Given all these arguments, I feel like the cons outweigh the pros of enabling the `network_policy` by default. I actually feel like there are no real pros for this feature, since users who are planning to use the `network_policy` feature will probably inform themselves on how to enable it within this module. It also makes more sense that users who are not aware of this feature or don't want to use it, don't need to do anything. (In contrast to the current situation where this last category of users is probably experiencing the described overhead without realising it.) 

Thanks for taking this suggestion into consideration! I'm more than happy to create a PR if this gets approved.

JeremyKeustersML6:
This PR fixes #804.

The only thing I'm not sure about is whether or not the `network_policy` checks are necessary in the `stub` integration tests (`stub_domains`, `stub_domains_private`, `stub_domains_upstream_nameservers`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ module "gke" {
   ip_range_services          = "us-central1-01-gke-01-services"
   http_load_balancing        = false
   horizontal_pod_autoscaling = true
-  network_policy             = true
+  network_policy             = false
 
   node_pools = [
     {
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ Then perform the following commands on the root folder:
 | monitoring\_service | The monitoring service that the cluster should write metrics to. Automatically send metrics from pods in the cluster to the Google Cloud Monitoring API. VM metrics will be collected by Google Compute Engine regardless of this setting Available options include monitoring.googleapis.com, monitoring.googleapis.com/kubernetes (beta) and none | `string` | `"monitoring.googleapis.com/kubernetes"` | no |
 | name | The name of the cluster (required) | `string` | n/a | yes |
 | network | The VPC network to host the cluster in (required) | `string` | n/a | yes |
-| network\_policy | Enable network policy addon | `bool` | `true` | no |
+| network\_policy | Enable network policy addon | `bool` | `false` | no |
 | network\_policy\_provider | The network policy provider. | `string` | `"CALICO"` | no |
 | network\_project\_id | The project ID of the shared VPC's host (for shared vpc support) | `string` | `""` | no |
 | node\_metadata | Specifies how node metadata is exposed to the workload running on the node | `string` | `"GKE_METADATA_SERVER"` | no |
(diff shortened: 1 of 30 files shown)


comment-bot-dev:

morgante (changes requested):
Thanks! This is fine, but FYI we are going to hold off on merging until the next major release.

JeremyKeustersML6:
Thanks! Do you have any more information in the meantime on why the CI/CD pipeline is failing? Can't see the cloud build details.

bharathkkb (commented):
Hi @JeremyKeustersML6 
We have an upcoming breaking release coming up that we can roll this into. Could you add this to the [upgrade guide ](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/docs/upgrading_to_v14.0.md )showing how users can continue to use netpol with 14.0? A diff like [this](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/docs/upgrading_to_v10.0.md#default-machine-type) should work.

JeremyKeustersML6:
Sure @bharathkkb ! Will push something tomorrow.

JeremyKeustersML6:
I added some documentation @bharathkkb . Let me know if it's okay!

morgante:
@JeremyKeustersML6 Please resolve the linter errors ^^^

morgante (approved):

PR #814, opened 2021-02-05
Feat: Control logging access

rux616:
When using this module in a project that maintains control over project-level IAM with a `google_project_iam_policy` resource, adding IAM roles through `google_project_iam_member` resources becomes a bit problematic as the resources will fight. This PR adds a flag to control whether the service account is automatically granted access to the `logging.logWriter`, `monitoring.metricWriter`, `monitoring.viewer`, and `stackdriver.resourceMetadata.writer` roles or not.

While this could already be accomplished by setting your own service account, a user may wish to have the module still create the service account but at the same time have IAM roles managed elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ Then perform the following commands on the root folder:
 | firewall\_inbound\_ports | List of TCP ports for admission/webhook controllers | `list(string)` | <pre>[<br>  "8443",<br>  "9443",<br>  "15017"<br>]</pre> | no |
 | firewall\_priority | Priority rule for firewall rules | `number` | `1000` | no |
 | gcloud\_upgrade | Whether to upgrade gcloud at runtime | `bool` | `false` | no |
+| grant\_logging\_access | Grants created cluster-specific service account logging.logWriter, monitoring.metricWriter, monitoring.viewer, and stackdriver.resourceMetadata.writer roles. | `bool` | `true` | no |
 | grant\_registry\_access | Grants created cluster-specific service account storage.objectViewer role. | `bool` | `false` | no |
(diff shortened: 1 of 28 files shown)


morgante:
Thanks for the contribution, but I'm not sure we want to support this. Every additional flag complicates the module interface, and in this case there's a clear workaround: create the service account outside the module.

In fact, I would assume that in most cases if you're maintaining tight control over IAM (via `google_project_iam_policy`) you probably also want tighter control over the service accounts as well.

@bharathkkb Thoughts?

comment-bot-dev:

rux616:
That's fair. I would, however, argue that because the base module already has 63 inputs (and the beta modules have 80+), another one isn't going to have any measurable impact on the level of complexity.

As I see it, this merely offers a slightly different route for SA/IAM management (a zig, instead of a zag, as it were), as the default behavior doesn't change at all.

morgante:
> That's fair. I would, however, argue that because the base module already has 63 inputs (and the beta modules have 80+), another one isn't going to have any measurable impact on the level of complexity.

Fair, but we're trying to at least stop the growth where possible.

bharathkkb:
@morgante I agree that the workaround to create SA outside the module is straightforward. @rux616 you can also wrap this module + the SA creation into another module which you can then use as a single module. WDYT?

morgante:
I'm going to go ahead and close this for now since there is a viable approach without it, but feel free to continue discussion if there's points we've missed.

rux616:
The wrapping mechanism is what I was doing already, and I was just hoping that I could dispense with that, but I totally understand the reasons. Thanks for the discussion.

PR #815, opened 2021-02-07
Feat: Multiple registry projects

rux616:
In our organization, we have a few projects that we put container images into depending on the team, and sometimes a cluster needs access to more than one of them. Right now we use a custom module that _isn't_ a wrapper for this one, but I'd like to change that. This PR would allow that to happen, but it does mean a breaking change for those who utilize the variable as it is currently defined.

I've also updated the tests to handle the change of the variable name from `registry_project_id` to `registry_project_ids` and the type now being `list(string)`. The `workload-metadata-config-local` integration test (the only test to reference the former `registry_project_id` variable) passes locally as well. That being said, I've not used InSpec before and am not 100% sure everything's correct regarding my changes, so if it's decided that this PR merits inclusion, I'd appreciate a closer look at that part of the PR just to make sure nothing's broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ Then perform the following commands on the root folder:
 | project\_id | The project ID to host the cluster in (required) | `string` | n/a | yes |
 | region | The region to host the cluster in (optional if zonal cluster / required if regional) | `string` | `null` | no |
 | regional | Whether is a regional cluster (zonal cluster if set false. WARNING: changing this after cluster creation is destructive!) | `bool` | `true` | no |
+| registry\_project\_id | Deprecated. Replaced by `registry_project_ids`. Still works for the purpose of backwards compatibility, but will be removed in a future version. | `string` | `""` | no |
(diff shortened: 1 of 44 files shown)


comment-bot-dev:

morgante (changes requested):
For backwards compatibility, could you keep the existing `registry_project_id` as well? (Otherwise we'll have to wait until next major release.)

rux616:
> For backwards compatibility, could you keep the existing `registry_project_id` as well? (Otherwise we'll have to wait until next major release.)

Sure. I'm thinking that I could rename the `registry_project_ids` variable to something like `additional_registry_project_ids` for that.

Alternatively, adding some verbiage to `registry_project_id` saying that it's still functional but planned to be superseded by `registry_project_ids` could be a good idea. Thoughts?

morgante:
> Alternatively, not renaming the new registry_project_ids variable and simply adding some verbiage to registry_project_id saying that it's still functional but planned to be superseded by registry_project_ids could be a good idea. Thoughts?

Right, that's what I had in mind.

rux616:
@morgante The requested changes have been made. I've added the `registry_project_id` variable back in, and made its description a deprecation warning. I've left the tests as they are for the new `registry_project_ids` variable.

Once this PR is merged, I'll create another one that will actually remove the `registry_project_id` variable and then we can just let it sit until the next major version is imminent.

morgante (approved):

PR #840, opened 2021-03-04
feat!: Added support for multi-project GKE Hub registration

### LINKED ISSUE #837: Needs to use --gke-uri instead of --gke-cluster in a multi-project scenario
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/b3034492db3c14e499e89c62f417a87196496dec/modules/hub/scripts/gke_hub_registration.sh#L40

Suppose a GKE cluster resides in GKE_PROJECT_ID different from HUB_PROJECT_ID. In such a multi-project registration scenario, the command line param `--gke-cluster` won't work. I believe we'd have to use the `--gke-uri` param instead.

BTW, `--project` in this command refers to the HUB_PROJECT_ID. When GKE_PROJECT_ID == HUB_PROJECT_ID, this distinction isn't important. But in a multi-project registration scenario, it is better to rename `${PROJECT_ID}` to `${HUB_PROJECT_ID}` in order to be more clear on the concept.

JamesDuncanNz:
Added an optional variable HUB_PROJECT_ID which will allow you to specific a seperate
GCP project for the GKE HUB than the project the cluster is deployed to.

This included updating the 3 examples the leveraged the hub module.

Fixes #837

### DIFF
--- a/examples/simple_zonal_with_hub/hub.tf
+++ b/examples/simple_zonal_with_hub/hub.tf
@@ -17,6 +17,7 @@
 module "hub" {
   source           = "../../modules/hub"
   project_id       = var.project_id
+  hub_project_id   = var.hub_project_id
   location         = module.gke.location
   cluster_name     = module.gke.name
   cluster_endpoint = module.gke.endpoint
--- a/examples/simple_zonal_with_hub/hub.tf
+++ b/examples/simple_zonal_with_hub/hub.tf
@@ -17,7 +17,6 @@
 module "hub" {
   source           = "../../modules/hub"
   project_id       = var.project_id
-  hub_project_id   = var.hub_project_id
   location         = module.gke.location
   cluster_name     = module.gke.name
   cluster_endpoint = module.gke.endpoint
--- a/modules/hub/main.tf
+++ b/modules/hub/main.tf
@@ -19,26 +19,43 @@ locals {
 
   is_gke_flag               = var.use_kubeconfig ? 0 : 1
   create_cmd_gke_entrypoint = "${path.module}/scripts/gke_hub_registration.sh"
-  create_cmd_gke_body       = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${local.gke_hub_sa_key} ${var.project_id} ${var.labels}"
+  create_cmd_gke_body       = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${local.gke_hub_sa_key} ${var.project_id} ${var.hub_project_id == "" ? var.project_id : var.hub_project_id} ${var.labels}"
   destroy_gke_entrypoint    = "${path.module}/scripts/gke_hub_unregister.sh"
-  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id}"
+  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id} ${var.hub_project_id == "" ? var.project_id : var.hub_project_id}"
 }
 
 data "google_client_config" "default" {
 }
 
+data "google_project" "project" {
+  project_id = var.hub_project_id == "" ? var.project_id : var.hub_project_id
+}
+
 resource "google_service_account" "gke_hub_sa" {
   count        = var.use_existing_sa ? 0 : 1
   account_id   = var.gke_hub_sa_name
-  project      = var.project_id
+  project      = var.hub_project_id == "" ? var.project_id : var.hub_project_id
   display_name = "Service Account for GKE Hub Registration"
 }
 
 resource "google_project_iam_member" "gke_hub_member" {
   count   = var.use_existing_sa ? 0 : 1
-  project = var.project_id
+  project = var.hub_project_id == "" ? var.project_id : var.hub_project_id
   role    = "roles/gkehub.connect"
   member  = "serviceAccount:${google_service_account.gke_hub_sa[0].email}"
+}
+
+resource "google_project_iam_member" "hub_service_agent_project" {
+  project = var.project_id
+  role    = "roles/gkehub.serviceAgent"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-gkehub.iam.gserviceaccount.com"
+}
+
+resource "google_project_iam_member" "hub_service_agent_hub" {
+  count   = var.hub_project_id == "" ? 0 : 1
+  project = var.hub_project_id
+  role    = "roles/gkehub.serviceAgent"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-gkehub.iam.gserviceaccount.com"
 }
 
 resource "google_service_account_key" "gke_hub_key" {
--- a/modules/hub/main.tf
+++ b/modules/hub/main.tf
@@ -45,17 +45,18 @@ resource "google_project_iam_member" "gke_hub_member" {
   member  = "serviceAccount:${google_service_account.gke_hub_sa[0].email}"
 }
 
-resource "google_project_iam_member" "hub_service_agent_project" {
-  project = var.project_id
-  role    = "roles/gkehub.serviceAgent"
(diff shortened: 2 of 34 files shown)


comment-bot-dev:

examples/simple_zonal_with_hub/hub.tf
```diff
@@ -17,6 +17,7 @@
 module "hub" {
   source           = "../../modules/hub"
   project_id       = var.project_id
+  hub_project_id   = var.hub_project_id
```
bharathkkb, line 20: nit: since we are not using another hub project in the examples, we can drop this here and throughout.

modules/hub/scripts/gke_hub_registration.sh
```diff
@@ -37,18 +38,19 @@ echo "${SERVICE_ACCOUNT_KEY}" | base64 ${B64_ARG} > "$tmp_file"
 
 if [[ ${GKE_CLUSTER_FLAG} == 1 ]]; then
     echo "Registering GKE Cluster."
-    gcloud container hub memberships register "${MEMBERSHIP_NAME}" --gke-cluster="${CLUSTER_LOCATION}"/"${CLUSTER_NAME}" --service-account-key-file="${tmp_file}" --project="${PROJECT_ID}" --quiet
+    CLUSTER_URI="https://container.googleapis.com/projects/${PROJECT_ID}/locations/${CLUSTER_LOCATION}/clusters/${CLUSTER_NAME}"
```
bharathkkb, line 41: nit: Maybe we can collapse `PROJECT_ID`, `CLUSTER_LOCATION` and `CLUSTER_NAME` to a single URI param constructed in TF and pass it in to the script

modules/hub/main.tf
```diff
@@ -19,28 +19,45 @@ locals {
...
   destroy_gke_entrypoint    = "${path.module}/scripts/gke_hub_unregister.sh"
-  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id}"
+  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id} ${var.hub_project_id == "" ? var.project_id : var.hub_project_id}"
 }
 
 data "google_client_config" "default" {
 }
 
+data "google_project" "project" {
+  project_id = var.hub_project_id == "" ? var.project_id : var.hub_project_id
```
bharathkkb, line 31: nit: It looks like this conditional is used quite a few times, we could define this as a local  and we can just reference the local throughout.

modules/hub/main.tf
```diff
@@ -19,28 +19,45 @@ locals {
...
+  create_cmd_gke_body       = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${local.gke_hub_sa_key} ${var.project_id} ${var.hub_project_id == "" ? var.project_id : var.hub_project_id} ${var.labels}"
   destroy_gke_entrypoint    = "${path.module}/scripts/gke_hub_unregister.sh"
-  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id}"
+  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id} ${var.hub_project_id == "" ? var.project_id : var.hub_project_id}"
 }
 
 data "google_client_config" "default" {
 }
 
+data "google_project" "project" {
```
bharathkkb, line 30: for clarity btw hub and gke projects
```suggestion
data "google_project" "hub_project" {
```

modules/hub/main.tf
```diff
@@ -19,28 +19,45 @@ locals {
...
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-gkehub.iam.gserviceaccount.com"
+}
+
+resource "google_project_iam_member" "hub_service_agent_hub" {
+  count   = var.hub_project_id == "" ? 0 : 1
+  project = var.hub_project_id
+  role    = "roles/gkehub.serviceAgent"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-gkehub.iam.gserviceaccount.com"
+}
+
```
bharathkkb, line 60: IIUC when `hub_project_id ` is set we need to grant `roles/gkehub.serviceAgent` on `var.project_id`(GKE Project). This looks like we are granting across both projects?

bharathkkb (commented):
Thanks for the PR @JamesDuncanNz

examples/simple_zonal_with_hub/hub.tf
JamesDuncanNz: Thanks, I'll remove the `hub_project_id` reference from each of the examples `hub.tf` and `variables.tf`  files.

modules/hub/main.tf
```diff
@@ -19,28 +19,46 @@ locals {
...
+  project = var.hub_project_id == "" ? var.project_id : var.hub_project_id
   role    = "roles/gkehub.connect"
   member  = "serviceAccount:${google_service_account.gke_hub_sa[0].email}"
 }
 
+resource "google_project_iam_member" "hub_service_agent_hub" {
+  count   = var.hub_project_id == "" ? 0 : 1
+  project = var.hub_project_id
+  role    = "roles/gkehub.serviceAgent"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-gkehub.iam.gserviceaccount.com"
```
bharathkkb, line 52: With `google_project_service_identity` you can get the email dynamically via `google_project_service_identity.sa_gkehub.email`

modules/hub/main.tf
JamesDuncanNz: That is correct, I fixed this in the commit below. I removed adding the role to the local project and instead make sure the sa is enabled by using `google_project_service_identity`

modules/hub/main.tf
JamesDuncanNz: Thanks, I'll update this.

modules/hub/main.tf
```diff
@@ -17,30 +17,58 @@
...
+  create_cmd_gke_body       = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${local.gke_hub_sa_key} ${local.cluster_uri} ${local.hub_project} ${var.labels}"
   destroy_gke_entrypoint    = "${path.module}/scripts/gke_hub_unregister.sh"
-  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${var.location} ${var.cluster_name} ${var.project_id}"
+  destroy_gke_body          = "${local.is_gke_flag} ${var.gke_hub_membership_name} ${local.cluster_uri} ${local.hub_project}"
 }
 
 data "google_client_config" "default" {
 }
 
+data "google_project" "hub_project" {
```
bharathkkb, line 33: nit: do we need this since we are getting identity via `google_project_service_identity`

modules/hub/main.tf
```diff
@@ -17,30 +17,58 @@
...
+  role    = "roles/gkehub.serviceAgent"
+  member  = "serviceAccount:${google_project_service_identity.sa_gkehub[0].email}"
+}
+
+resource "google_project_iam_member" "hub_service_agent_hub" {
+  count   = var.hub_project_id == "" ? 0 : 1
+  project = local.hub_project
+  role    = "roles/gkehub.serviceAgent"
+  member  = "serviceAccount:${google_project_service_identity.sa_gkehub[0].email}"
+}
```
bharathkkb, line 63: Do we still need two? IIUC we only need this on the cluster project(i.e`var.project_id` if `var.hub_project_id != ""`)?

bharathkkb (commented):
Overall LGTM
@JamesDuncanNz can you test an upgrade from previous hub version to this. There maybe a delete recreate op as we are changing `create_cmd_gke_body` and would be good to capture this in the upgrade guide as it may result in the de registration and subsequent registration during upgrade.

modules/hub/main.tf
JamesDuncanNz: For the multi project registration the [gcp-sa-gkehub project SA requires the Service Agent role in both projects](https://cloud.google.com/anthos/multicluster-management/connect/prerequisites#gke-cross-project).

modules/hub/main.tf
JamesDuncanNz: No longer needed, I'll remove it.

bharathkkb (approved):

PR #841, opened 2021-03-05
Add new property to explicitly return GKE private_endpoint

### LINKED ISSUE #828: modules/private-cluster endpoint incorrect
Sorry if I've got the wrong end of the stick here but I am having a problem with my private clusters. I've been having this problem with my own project and decided to run the example project (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples/private_zonal_with_networking) but I get the same result.

If I look in the GCP UI at the cluster it tells me the endpoint is `172.16.0.2` which is what I would expect but the output within terraform has the endpoint `104.196.20.229`

```
Outputs:

ca_certificate = LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURLekNDQWhPZ0F3SUJBZ0lSQUtqM3pTK0VMSVFFQ1UxNEl4eitwNGt3RFFZSktvWklodmNOQVFFTEJRQXcKTHpFdE1Dc0dBMVVFQXhNa1lXVmtNVGhoTUdNdE56WXhZUzAwWkdZeExUaGpPREl0WmpZMk5HRTROVGhqTlROaQpNQjRYRFRJeE1ESXhPVEUxTlRZek4xb1hEVEkyTURJeE9ERTJOVFl6TjFvd0x6RXRNQ3NHQTFVRUF4TWtZV1ZrCk1UaGhNR010TnpZeFlTMDBaR1l4TFRoak9ESXRaalkyTkdFNE5UaGpOVE5pTUlJQklqQU5CZ2txaGtpRzl3MEIKQVFFRkFBT0NBUThBTUlJQkNnS0NBUUVBcTIxVFBmVHlmYWRobGFVT0ZWbjQvR3FPaWtNVU81TWpSaUErUThxVgptaTVsVklLa2JOcU9haXVsYzl0SnBtempwaUU3MjNRN1pUK2tBY3B0YXVyeGxweENrMmduK1o5RSs1QThBOC9QClZYcXBjdVhIUzB1a3h6RXZob2E1bmdDRmlSRGpqMWxiL2ZBbW80bldTKzVMc2I0S1dvazZnVzE4NG1SUWhydysKcC9TaFBNM1dTbVhDT0ZMZmswL0VVYVAyK3BZbTA4WVYrWUtqMU8vS1dsZFR6UWV3MnJRcXRHK29zR2lLVktnRApyR2V5bXhOYUJKdGcvdWFCYnZiQ0xGc2dTeGgrVE9GNkZ2WWRZMUNaQ25vSExYSDNvWXpvQ2pETllQNUZ5Z2dmCkFxQ1UwWFhPWlhPSmtrdHNMR3hXWVhlNXlvOTJJS3JJbUxOVDhsdG42ajJTaXdJREFRQUJvMEl3UURBT0JnTlYKSFE4QkFmOEVCQU1DQWdRd0R3WURWUjBUQVFIL0JBVXdBd0VCL3pBZEJnTlZIUTRFRmdRVXNPMDZJNlVaQ3ZhSApMRURKQS8zWVFrclRlVG93RFFZSktvWklodmNOQVFFTEJRQURnZ0VCQURaSWdFVDNaRG9rakZ6aWx4Z2VKbU9NCndOVVd4MUFUQmZmalZkd3NGL0tvMUtIZnlHYXNvWjlzWTQyclJJMENac1hWQUwvY0M4WEFmdWRvcTFDVTFxY0IKb2dFZDg4cU9LTVAvWFZrYm1kZS9oMitQUzRNTk94NHo0VkVETlppVzVRdFNCdCsrYU9UbFlGcHpBTGJUVkQ3SApoWDFpb1RhdGozM0Vuc0RsemlNOThEc3hTWXJkT0o5YnUrazNNTWZHR0RBV1RKUVdZU1djbGNqMWZDR3BOeHBJCmt5WXpicW9zY2h0b1FTaC9qU0tyTjVPeGkvMTYxb3pNbTc3cG9Vc2pqM3ZYTWpPRzhIRHdNbmdEMHR5a255L1cKcmxRN0xvZDNMbW9MUnF2amMwUUIwK3JMWVJ2TG5QWU1kRVkxSW9pdERYT3Y0NDh2akdkZ0crUnNUd0ZuV0xjPQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
client_token = <sensitive>
cluster_name = gke-on-vpc-cluster
ip_range_pods_name = ip-range-pods
ip_range_services_name = ip-range-scv
kubernetes_endpoint = 104.196.20.229
location = us-east1-b
master_kubernetes_version = 1.18.15-gke.1100
network = gke-network
network_name = gke-network
peering_name = gke-n9c50b1751d1d8287509-37e3-8f1f-peer
project_id = tf-example-test
region = us-east1
service_account = tf-gke-gke-on-vpc-clus-razb@tf-example-test.iam.gserviceaccount.com
subnet_name = [
  "gke-subnet",
]
subnet_secondary_ranges = [
  [
    {
      "ip_cidr_range" = "192.168.0.0/18"
      "range_name" = "ip-range-pods"
    },
    {
      "ip_cidr_range" = "192.168.64.0/18"
      "range_name" = "ip-range-scv"
    },
  ],
]
subnetwork = gke-subnet
zones = []

### PR DESCRIPTION
**Pull request by fstr**
This fixes #828 on the `auth` module. A new flag `use_private_endpoint` is introduced. If this flag is set to `true`, the `private_endpoint` property of the GKE cluster is looked up.

If `use_private_endpoint` is set to `false` (default), the `data.google_container_cluster.gke_cluster.private_cluster_config[0].public_endpoint` will be used if available with a fallback to the default GKE `endpoint` property.

### DIFF
--- a/modules/auth/main.tf
+++ b/modules/auth/main.tf
@@ -17,9 +17,8 @@
 locals {
   cluster_ca_certificate = data.google_container_cluster.gke_cluster.master_auth != null ? data.google_container_cluster.gke_cluster.master_auth[0].cluster_ca_certificate : ""
   private_endpoint       = try(data.google_container_cluster.gke_cluster.private_cluster_config[0].private_endpoint, "")
-  public_endpoint        = try(data.google_container_cluster.gke_cluster.private_cluster_config[0].public_endpoint, "")
   default_endpoint       = data.google_container_cluster.gke_cluster.endpoint != null ? data.google_container_cluster.gke_cluster.endpoint : ""
-  endpoint               = var.use_private_endpoint == true ? local.private_endpoint : local.public_endpoint != "" ? local.public_endpoint : local.default_endpoint
+  endpoint               = var.use_private_endpoint == true ? local.private_endpoint : local.default_endpoint
   host                   = local.endpoint != "" ? "https://${local.endpoint}" : ""
   context                = data.google_container_cluster.gke_cluster.name != null ? data.google_container_cluster.gke_cluster.name : ""
 }
--- a/modules/auth/main.tf
+++ b/modules/auth/main.tf
@@ -16,8 +16,11 @@
 
 locals {
   cluster_ca_certificate = data.google_container_cluster.gke_cluster.master_auth != null ? data.google_container_cluster.gke_cluster.master_auth[0].cluster_ca_certificate : ""
-  endpoint               = data.google_container_cluster.gke_cluster.endpoint != null ? data.google_container_cluster.gke_cluster.endpoint : ""
-  host                   = data.google_container_cluster.gke_cluster.endpoint != null ? "https://${data.google_container_cluster.gke_cluster.endpoint}" : ""
+  private_endpoint       = try(data.google_container_cluster.gke_cluster.private_cluster_config[0].private_endpoint, "")
+  public_endpoint        = try(data.google_container_cluster.gke_cluster.private_cluster_config[0].public_endpoint, "")
+  default_endpoint       = data.google_container_cluster.gke_cluster.endpoint != null ? data.google_container_cluster.gke_cluster.endpoint : ""
+  endpoint               = var.use_private_endpoint == true ? local.private_endpoint : local.public_endpoint != "" ? local.public_endpoint : local.default_endpoint
+  host                   = local.endpoint != "" ? "https://${local.endpoint}" : ""
   context                = data.google_container_cluster.gke_cluster.name != null ? data.google_container_cluster.gke_cluster.name : ""
 }
 
--- a/modules/auth/README.md
+++ b/modules/auth/README.md
@@ -9,15 +9,18 @@ This module retrieves a token for the account configured with the `google`
 provider as the Terraform runner using the provider's `credentials`,
 `access_token`, or other means of authentication.
 
+If you run a [private cluster](https://cloud.google.com/kubernetes-engine/docs/concepts/private-cluster-concept), you can set the `use_private_endpoint` property to return the GKE private_endpoint IP address.
+
 ## Usage
 
 ```tf
 module "gke_auth" {
-  source           = "terraform-google-modules/kubernetes-engine/google//modules/auth"
+  source               = "terraform-google-modules/kubernetes-engine/google//modules/auth"
 
-  project_id       = "my-project-id"
-  cluster_name     = "my-cluster-name"
-  location         = module.gke.location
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: modules/auth/main.tf [hunk 1]
```diff
@@ -16,8 +16,11 @@
 
 locals {
   cluster_ca_certificate = data.google_container_cluster.gke_cluster.master_auth != null ? data.google_container_cluster.gke_cluster.master_auth[0].cluster_ca_certificate : ""
-  endpoint               = data.google_container_cluster.gke_cluster.endpoint != null ? data.google_container_cluster.gke_cluster.endpoint : ""
-  host                   = data.google_container_cluster.gke_cluster.endpoint != null ? "https://${data.google_container_cluster.gke_cluster.endpoint}" : ""
+  private_endpoint       = try(data.google_container_cluster.gke_cluster.private_cluster_config[0].private_endpoint, "")
+  public_endpoint        = try(data.google_container_cluster.gke_cluster.private_cluster_config[0].public_endpoint, "")
+  default_endpoint       = data.google_container_cluster.gke_cluster.endpoint != null ? data.google_container_cluster.gke_cluster.endpoint : ""
+  endpoint               = var.use_private_endpoint == true ? local.private_endpoint : local.public_endpoint != "" ? local.public_endpoint : local.default_endpoint
```
morgante, line 22: Shouldn't we return the default endpoint if you don't express a preference?

```suggestion
  endpoint               = var.use_private_endpoint == true ? local.private_endpoint : local.public_endpoint != "" ? local.public_endpoint : local.default_endpoint
```

comment-bot-dev:

modules/auth/main.tf
fstr: I thought about this as well. It might lead to unexpected behaviour though.

E. g. you provide `use_private_endpoint = true`. No `private_endpoint` is found, and as a result the default endpoint is returned. I think I'd rather have an empty `host` in that case and a failing follow-up request. But I'm open to a change.

morgante: Right, I'm fine with the behavior when `use_private_endpoint = true`. But when `use_private_endpoint = false` we should simply return the default endpoint (not necessarily the public endpoint).

fstr: Agree, this will also guarantee backwards compatibility.

morgante (approved):

Issue #863, opened 2021-04-10
Missing option to install ASM managed control plane


LucaPrete:
I will send a PR to cover this.

bharathkkb:
fixed via https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/864

PR #864, opened 2021-04-10
[#863] Add managed ctrl plane option to ASM module

LucaPrete:


comment-bot-dev:
Thanks for the PR! 🚀
- ⚠️check_shell
 Failed shell check. More info on running shellcheck locally [here](https://www.shellcheck.net). 
 ```bash 
Running shellcheck
In ./modules/asm/scripts/install_asm.sh line 51:
./install_asm $(echo "${params[@]}")
              ^--------------------^ SC2046: Quote this to prevent word splitting.
              ^--------------------^ SC2116: Useless echo? Instead of 'cmd $(echo foo)', just use 'cmd foo'.
For more information:
  https://www.shellcheck.net/wiki/SC2046 -- Quote this to prevent word splitt...
  https://www.shellcheck.net/wiki/SC2116 -- Useless echo? Instead of 'cmd $(e...
 ```

bharathkkb (commented):
Thanks for the PR @LucaPrete

modules/asm/main.tf
```diff
@@ -33,6 +33,6 @@ module "asm_install" {
   service_account_key_file = var.service_account_key_file
 
 
-  kubectl_create_command  = "${path.module}/scripts/install_asm.sh ${var.project_id} ${var.cluster_name} ${var.location} ${var.asm_version}"
+  kubectl_create_command  = "${path.module}/scripts/install_asm.sh ${var.project_id} ${var.cluster_name} ${var.location} ${var.asm_version} ${var.managed}"
```
bharathkkb, line 36: This could cause a delete recreate operation since we are changing the command and would need a breaking release. An alternative could be to keep the command as is and switch based on `${var.managed}` and default managed to false in the script.

` kubectl_create_command = var.managed ? "${local.kubectl_create_command_base} ${var.managed} : local.kubectl_create_command_base`

morgante: Agreed, let's set a default and keep this backwards-compatible.

LucaPrete: Makes totally sense! Submitting the updated patch

LucaPrete:
@morgante @bharathkkb not sure why this is failing, but it doesn't seem related to the code. From the tests here on GitHub I see:

```
Error from server (NotFound): memberships.hub.gke.io "membership" not found
```

bharathkkb (approved):

PR #882, opened 2021-04-29
Introduce flag add_master_webhook_firewall_rules

### LINKED ISSUE #880: Only one flag for two unrelated additional firewall rules
I'm currently using the beta-private-cluster module and I've installed Istio. I need to open the port `15017` from the GCP managed control-plane to my nodes and pods.

The module has the property `firewall_inbound_ports` with the default values `["8443", "9443", "15017"]`. I made use of this property, but nothing changes. After reading the TF code, I figured out that I need to set `add_cluster_firewall_rules` to `true`, so my configured inbound ports would be opened.

The `add_cluster_firewall_rules` not only controls the `firewall_inbound_ports`, but also adds the resource `google_compute_firewall.intra_egress`, even if I don't want those firewall rules to be applied. They don't do any harm, but I have a shared networking project with many clusters attached and I'd like to keep the firewall rules to an absolute minimum. 

Proposals:
- Quick win: Improve docs and add a note to `firewall_inbound_ports` that `add_cluster_firewall_rules` must be set
- *Breaking change:* Remove default values from `firewall_inbound_ports`. In my case I only need one of those ports, so I can overwrite the defaults. But I don't see a reason why a default setting should open more ports than necessary.
- *Breaking change:* Use separate flags to control the creation of `google_compute_firewall.master_webhooks` and `google_compute_firewall.intra_egress`
  - a) Separate flag could be introduced to control creation of webhooks
  - b) `google_compute_firewall.master_webhooks` could only be applied if `length(var.firewall_inbound_ports) > 0`, making an additional flag obsolete. If there are no ports to be opened, there's no need for the firewall rule.
    - If we keep the default values for `firewall_inbound_ports` this would always be applied, adding rules for users that haven't had them before.
    -  If we also remove the default values for `firewall_inbound_ports`, the firewall rule would be removed for users that relied on the default values

fstr:
Resolves #880 

- Flag controls exclusive creation of a firewall rule for ports defined in firewall_inbound_ports
- The existing flag add_cluster_firewall_rules is kept on the firewall rule resource for backwards compatibility

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Then perform the following commands on the root folder:
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
 | add\_cluster\_firewall\_rules | Create additional firewall rules | `bool` | `false` | no |
+| add\_master\_webhook\_firewall\_rules | Create master\_webhook firewall rules for ports defined in `firewall_inbound_ports` | `bool` | `false` | no |
 | add\_shadow\_firewall\_rules | Create GKE shadow firewall (the same as default firewall rules with firewall logs enabled). | `bool` | `false` | no |
 | basic\_auth\_password | The password to be used with Basic Authentication. | `string` | `""` | no |
 | basic\_auth\_username | The username to be used with Basic Authentication. An empty value will disable Basic Authentication, which is the recommended configuration. | `string` | `""` | no |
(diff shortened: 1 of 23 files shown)


morgante (approved):

comment-bot-dev:

Issue #899, opened 2021-05-18
Cluster resource ephemeral_storage_config for localSSD


jeremyrcampb:
I'd like to [use localSSDs for ephemeral storage](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd) for specific node pools.  It looks like the Google provider has exposed this on the beta provider for the [container_cluster](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#ephemeral_storage_config) resource.  Is it on the radar to get added to this module?

Thank you!

morgante:
We'd be happy to review a PR adding this.

jeremyrcampb:
Addressed in merged PR.  Thanks!

PR #902, opened 2021-05-20
Add use local_ssd_ephemeral_count attribute in node_pool config (beta)

jeremyrcampb:
This addresses #899 by adding a new local_ssd_ephemeral_count attribute to the node_pool configuration.  This gets passed conditionally to a new `ephemeral_storage_config` block on the `google_container_node_pool` resource.

This is supported only on the google-beta provider, since 3.60.0, and is a beta feature in GKE 1.18 and later.

Feature and its use is documented [here](https://cloud.google.com/kubernetes-engine/docs/how-to/persistent-volumes/local-ssd).

Mutually Exclusive Configs:

Both local_ssd_count and local_ssd_ephemeral_count cannot be used on the same node pool, an error will return:

`Error: error creating NodePool: googleapi: Error 400: Can not use local SSDs as volumes when using ephemeral storage backed by local SSDs., badRequest`

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,20 +46,20 @@ module "gke" {
 
   node_pools = [
     {
-      name               = "default-node-pool"
-      machine_type       = "e2-medium"
-      node_locations     = "us-central1-b,us-central1-c"
-      min_count          = 1
-      max_count          = 100
-      local_ssd_count    = 0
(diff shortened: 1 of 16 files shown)


comment-bot-dev:

PR #920, opened 2021-06-07
use absolute path for overlay files similar to service account key file

Pega88:
Similar to the fix @ameer00 and I made for `KEY_FILE_COMMAND_SNIPPET` to add the full path to the service account key file, the `CUSTOM_OVERLAYS_COMMAND ` also needs the full path since the script generating the command is in a different location from where it is used, so it will not resolve relative paths correctly.


comment-bot-dev:

Pega88:
needs rework to work with multiple files, only works with a single file. will update.

PR #955, opened 2021-07-16
Enables an existing GSA to be used when setting up Workload Identity

jackwhelpton:
Resolves https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/956.

The workload-identity module currently allows for reusing an existing KSA, but not an existing GSA. This is a common requirement for us: typically, we will have two clusters within the same project (for example, a production cluster and a disaster recovery cluster), and want to configure KSAs in each cluster that are bound to the same GSA.

With the existing module, this isn't possible, as errors are thrown owing to the attempt to recreate an existing GSA.

This is a work-in-progress: I believe it's feature complete, but I'm still testing.

### DIFF
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -17,7 +17,7 @@
 locals {
   gcp_given_name = var.gcp_sa_name != null ? var.gcp_sa_name : var.name
   gcp_sa_name    = var.use_existing_gcp_sa ? local.gcp_given_name : google_service_account.main[0].name
-  gcp_sa_email   = var.use_existing_gcp_sa ? "${local.gcp_given_name}@${var.project_id}.iam.gserviceaccount.com" : google_service_account.main[0].email
+  gcp_sa_email   = google_service_account.cluster_service_account.email
   gcp_sa_fqn     = "serviceAccount:${local.gcp_sa_email}"
 
   k8s_sa_gcp_derived_name = "serviceAccount:${var.project_id}.svc.id.goog[${var.namespace}/${local.output_k8s_name}]"
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -15,8 +15,12 @@
  */
 
 locals {
+  gcp_given_name = var.gcp_sa_name != null ? var.gcp_sa_name : var.name
+  gcp_sa_name    = var.use_existing_gcp_sa ? local.gcp_given_name : google_service_account.cluster_service_account[local.gcp_given_name].name
+  gcp_sa_email   = var.use_existing_gcp_sa ? "${local.gcp_given_name}@${var.project_id}.iam.gserviceaccount.com" : google_service_account.cluster_service_account[local.gcp_given_name].email
+  gcp_sa_fqn     = "serviceAccount:${local.gcp_sa_email}"
+
   k8s_sa_gcp_derived_name = "serviceAccount:${var.project_id}.svc.id.goog[${var.namespace}/${local.output_k8s_name}]"
-  gcp_sa_email            = google_service_account.cluster_service_account.email
 
   # This will cause terraform to block returning outputs until the service account is created
   k8s_given_name       = var.k8s_sa_name != null ? var.k8s_sa_name : var.name
@@ -25,9 +29,11 @@ locals {
 }
 
 resource "google_service_account" "cluster_service_account" {
+  for_each = var.use_existing_gcp_sa ? [] : toset([local.gcp_given_name])
+
   # GCP service account ids must be < 30 chars matching regex ^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$
-  # KSA do not have this naming restriction.
-  account_id   = substr(var.name, 0, 30)
+  # KSAs do not have this naming restriction.
+  account_id   = substr(each.key, 0, 30)
   display_name = substr("GCP SA bound to K8S SA ${local.k8s_given_name}", 0, 100)
   project      = var.project_id
 }
@@ -40,7 +46,7 @@ resource "kubernetes_service_account" "main" {
     name      = var.name
     namespace = var.namespace
     annotations = {
-      "iam.gke.io/gcp-service-account" = google_service_account.cluster_service_account.email
+      "iam.gke.io/gcp-service-account" = local.gcp_sa_email
     }
   }
 }
@@ -61,7 +67,7 @@ module "annotate-sa" {
 }
 
 resource "google_service_account_iam_member" "main" {
-  service_account_id = google_service_account.cluster_service_account.name
+  service_account_id = local.gcp_sa_name
   role               = "roles/iam.workloadIdentityUser"
   member             = local.k8s_sa_gcp_derived_name
 }
@@ -72,5 +78,5 @@ resource "google_project_iam_member" "workload_identity_sa_bindings" {
 
   project = var.project_id
   role    = each.value
-  member  = "serviceAccount:${google_service_account.cluster_service_account.email}"
+  member  = local.gcp_sa_fqn
 }
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -17,7 +17,7 @@
 locals {
   gcp_given_name = var.gcp_sa_name != null ? var.gcp_sa_name : var.name
   gcp_sa_name    = var.use_existing_gcp_sa ? local.gcp_given_name : google_service_account.main[0].name
-  gcp_sa_email   = google_service_account.cluster_service_account.email
+  gcp_sa_email   = data.google_service_account.cluster_service_account.email
   gcp_sa_fqn     = "serviceAccount:${local.gcp_sa_email}"
 
   k8s_sa_gcp_derived_name = "serviceAccount:${var.project_id}.svc.id.goog[${var.namespace}/${local.output_k8s_name}]"
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -16,7 +16,7 @@
 
 locals {
   gcp_given_name = var.gcp_sa_name != null ? var.gcp_sa_name : var.name
-  gcp_sa_name    = var.use_existing_gcp_sa ? local.gcp_given_name : google_service_account.main[0].name
+  gcp_sa_name    = var.use_existing_gcp_sa ? local.gcp_given_name : google_service_account.main[0].account_id
   gcp_sa_email   = data.google_service_account.cluster_service_account.email
   gcp_sa_fqn     = "serviceAccount:${local.gcp_sa_email}"
 
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -28,6 +28,11 @@ locals {
   output_k8s_namespace = var.use_existing_k8s_sa ? var.namespace : kubernetes_service_account.main[0].metadata[0].namespace
 }
 
+data "google_service_account" "cluster_service_account" {
+  account_id = local.gcp_sa_name
+  project    = var.project_id
+}
+
 resource "google_service_account" "main" {
   count = var.use_existing_gcp_sa ? 0 : 1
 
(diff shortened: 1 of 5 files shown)


comment-bot-dev:

modules/workload-identity/variables.tf
```diff
@@ -32,22 +49,17 @@ variable "location" {
...
 }
 
 variable "namespace" {
-  description = "Namespace for k8s service account"
+  description = "Namespace for the Kubernetes service account"
   default     = "default"
   type        = string
 }
 
-variable "project_id" {
```
morgante, line 46: This variable should not be removed.

modules/workload-identity/variables.tf
```diff
@@ -32,22 +49,17 @@ variable "location" {
 }
 
 variable "k8s_sa_name" {
-  description = "Name for the existing Kubernetes service account"
+  description = "Name for the Kubernetes service account"
```
morgante, line 52:
```suggestion
  description = "Name for the Kubernetes service account, overrides `var.name`"
```

modules/workload-identity/variables.tf
```diff
@@ -19,6 +19,23 @@ variable "name" {
   type        = string
 }
 
+variable "project_id" {
+  description = "GCP project ID"
+  type        = string
+}
+
+variable "gcp_sa_name" {
+  description = "Name for the Google service account"
```
morgante, line 28:
```suggestion
  description = "Name for the Google service account, overrides `var.name`."
```

modules/workload-identity/output.tf
```diff
@@ -40,6 +40,6 @@ output "gcp_service_account_name" {
 }
 
 output "gcp_service_account" {
-  description = "GCP service account."
+  description = "GCP service account, if created; null if existing account was used."
```
morgante, line 43: We should always pass this output back. If an existing SA is used, just reflect that value back.

modules/workload-identity/main.tf
```diff
@@ -25,9 +29,11 @@ locals {
 }
 
 resource "google_service_account" "cluster_service_account" {
+  for_each = var.use_existing_gcp_sa ? [] : toset([local.gcp_given_name])
```
morgante, line 32: Count is preferred for conditionals.

```suggestion
  count = var.use_existing_gcp_sa ? 0 : 1
```

morgante (changes requested):

modules/workload-identity/variables.tf
jackwhelpton: It hasn't been removed: I just reordered the variables slightly, putting the required variables first and then grouping the optionals by those that pertain to the GSA, then the KSA, then implementation details. If you'd prefer this reverted or done differently, let me know.

modules/workload-identity/output.tf
jackwhelpton: Is there a way to lookup an existing GCP service account as a Terraform object in order to return it? I return all the other (string) output fields, but this one is the entirely (newly created) object...

modules/workload-identity/main.tf
jackwhelpton: OK. In our own code we've been moving the direction: collections keyed by a name rather than an index seem to be more resilient to changes.

morgante: Yes, collections keyed by name are preferred when it's actually a collection. This isn't a collection, just a binary (either 0 or 1 instances).

modules/workload-identity/variables.tf
morgante: Ah, missed that.

modules/workload-identity/output.tf
morgante: Yes, you can use the data source: https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/service_account

jackwhelpton: Ah, nice! Something like the most recently pushed change, then?

morgante: Yep, you'll still need to update the output to point to the data source.

jackwhelpton: I think I've done that... because I've used `main` as the name of the resource, to be consistent with the KSA, that's left me with `cluster_service_account` as a name for the data source. Given that, is the output just the same, i.e. `google_service_account.cluster_service_account`?

morgante: No. When the data source is used, you need to refer to it as `data.google_service_account.cluster_service_account`.

jackwhelpton:
@morgante I think all the changes you requested have been made now. The resulting module plans fine for me locally. I'm not sure why the `kubernetes-engine-int-trigger` check is failing, but unfortunately I don't seem to have permissions to view the Cloud Build logs... any further info you can give me to help me fix that up would be great.

morgante:
@jackwhelpton Here's the error:

```
       module.example.module.workload_identity_existing_ksa.google_service_account_iam_member.main: Creation complete after 4s [id=projects/ci-gke-c3429ab2-941a/serviceAccounts/existing-regional-cluster-4z3y@ci-gke-c3429ab2-941a.iam.gserviceaccount.com/roles/iam.workloadIdentityUser/serviceaccount:ci-gke-c3429ab2-941a.svc.id.goog[default/foo-ksa]]
       
       Error: "account_id" ("projects/ci-gke-c3429ab2-941a/serviceAccounts/existing-regional-cluster-4z3y@ci-gke-c3429ab2-941a.iam.gserviceaccount.com") doesn't match regexp "^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$"
       
         on ../../../modules/workload-identity/main.tf line 32, in data "google_service_account" "cluster_service_account":
         32:   account_id = local.gcp_sa_name
       
       
       
       Error: "account_id" ("projects/ci-gke-c3429ab2-941a/serviceAccounts/iden-regional-cluster-4z3y@ci-gke-c3429ab2-941a.iam.gserviceaccount.com") doesn't match regexp "^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$"
       
         on ../../../modules/workload-identity/main.tf line 32, in data "google_service_account" "cluster_service_account":
         32:   account_id = local.gcp_sa_name

```

jackwhelpton:
Thanks! I should have resolved that (using `account_id` rather than `name`), but still seeing an error. If there's a way for me to view these logs without bothering you every time, that probably be good for both of us!

I ran through the instructions for running tests locally, but owing to our organization restrictions it'd be preferable for me to use a service account in an existing project rather than having the test setup automatically create a new project. I've tried doing that, but end up getting credential errors (despite pointing it at a JSON key):

```
ERROR: (gcloud.auth.activate-service-account) Missing required argument [ACCOUNT]: An account is required when using .p12 keys
```

Given all that, it'd certainly be easier if I can just have the CI process run these tests for me.

morgante:
@jackwhelpton If you've having trouble setting up a test environment, this codelab might help: https://codelabs.developers.google.com/codelabs/cft-onboarding/#9

Unfortunately it's impossible for us to expose Cloud Build logs publicly. One other option for debugging is to just try `terraform applying` the example directly.

Here's the latest failure:
```
       module.example.module.workload_identity_existing_ksa.module.annotate-sa.module.gcloud_kubectl.null_resource.run_command[0]: Creation complete after 7s [id=2698573458861851095]
       
       Error: "service_account_id" ("iden-regional-cluster-mfsy") doesn't match regexp "projects/(?:(?:[-a-z0-9]{1,63}\\.)*(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?):)?(?:[0-9]{1,19}|(?:[a-z0-9](?:[-a-z0-9]{0,61}[a-z0-9])?)|-)/serviceAccounts/((?:(?:[-a-z0-9]{1,63}\\.)*(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?):)?(?:[0-9]{1,19}|(?:[a-z0-9](?:[-a-z0-9]{0,61}[a-z0-9])?))@[a-z]+.gserviceaccount.com$|[0-9]{1,20}-compute@developer.gserviceaccount.com|[a-z](?:[-a-z0-9]{4,28}[a-z0-9])@[-a-z0-9\\.]{1,63}\\.iam\\.gserviceaccount\\.com$)"
       
         on ../../../modules/workload-identity/main.tf line 75, in resource "google_service_account_iam_member" "main":
         75:   service_account_id = local.gcp_sa_name
       
       
       
       Error: "service_account_id" ("existing-regional-cluster-mfsy") doesn't match regexp "projects/(?:(?:[-a-z0-9]{1,63}\\.)*(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?):)?(?:[0-9]{1,19}|(?:[a-z0-9](?:[-a-z0-9]{0,61}[a-z0-9])?)|-)/serviceAccounts/((?:(?:[-a-z0-9]{1,63}\\.)*(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?):)?(?:[0-9]{1,19}|(?:[a-z0-9](?:[-a-z0-9]{0,61}[a-z0-9])?))@[a-z]+.gserviceaccount.com$|[0-9]{1,20}-compute@developer.gserviceaccount.com|[a-z](?:[-a-z0-9]{4,28}[a-z0-9])@[-a-z0-9\\.]{1,63}\\.iam\\.gserviceaccount\\.com$)"
       
         on ../../../modules/workload-identity/main.tf line 75, in resource "google_service_account_iam_member" "main":
         75:   service_account_id = local.gcp_sa_name
```

jackwhelpton:
Ta. I'm discussing with our ops team at the moment about creating a sandbox environment where we can use the integration test suite more as intended. In the meantime, I've addressed that most recent error and it seems like all the tests pass now. Any further changes I should make?

morgante (approved):

morgante:
> I'm discussing with our ops team at the moment about creating a sandbox environment where we can use the integration test suite more as intended. In the meantime, I've addressed that most recent error and it seems like all the tests pass now. Any further changes I should make?

Thanks for working through this. We're also working on a new test framework which should hopefully make contributions a bit easier.

jackwhelpton:
I know this is merged, but saw there was a slight problem with the renaming of `cluster_service_account` to `main`, so wanted to clear up a few questions to avoid that happening again:

* is it not expected that resources might change location between module versions, especially major releases, necessitating a `terraform state mv` when upgrading?
* this PR made the `google_service_account` optional, meaning that even with the original name, the resource would move from `google_service_account.cluster_service_account` to `google_service_account.cluster_service_account[0]`. Is the zero-index privileged in some way in the state file, such that this refactor does not require state moves?

I'd assumed that the change to `google_service_account.cluster_service_account[0]` _would_ be a breaking change, and given that it seemed to make sense to use the chance to refactor the GSA resources to line up more closely with the KSA ones.

bharathkkb:
Hi @jackwhelpton 

> is it not expected that resources might change location between module versions, especially major releases, necessitating a terraform state mv when upgrading?

Yes they may change, but we try to minimize it whenever possible and if necessary [add docs](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/docs) with helper commands (like `tf state mv`) to help with the upgrade. In this case, we thought it made sense to keep the resource name as is and because we missed identifying this as breaking during the release.

> this PR made the google_service_account optional, meaning that even with the original name, the resource would move from google_service_account.cluster_service_account to google_service_account.cluster_service_account[0]. Is the zero-index privileged in some way in the state file, such that this refactor does not require state moves?

I believe so, since technically when we use `count=1` it is possibly to assume the zero instance is the instance that existed before(vice versa if were to go from `count =1` to using no count) and TF probably takes this into account to prevent unnecessary diffs.

jackwhelpton:
Thanks for those clarifications. I knew writing, e.g. `cluster_service_account.name` rather than `cluster_service_account[0].name` in TF code would result in an error, but I hadn't realized the equivalence TF was drawing at plan-time; another compelling argument for using `count` over `for_each` for optional resources.

If I'd known that in advance, I wouldn't have done that rename, so fully agree with reverting it: the consistency between GSA and KSA names isn't worth the added update pain. Cheers for fixing the resulting mess too :)

PR #972, opened 2021-08-11
Consider k8s service account name from evaluated local parameter

sig-abyreddy:
Module supports providing a custom name for kubernetes service account through **k8s_sa_name** variable. But internally this value is not considered for kubernetes service account resource.

**k8s_given_name** has evaluated value for service account name under locals. Using this value will respect custom input from the user.


sig-abyreddy:
@googlebot I signed it!

morgante (approved):
Test failure is unrelated.

sig-abyreddy:
Thank for accepting the MR @morgante

PR #999, opened 2021-09-09
'pod_security_policy_config' block should be static

s1mark:
## What
The `pod_security_policy_config` block should be defined in all cases and not managed dynamically

## Why
In case someone enables the policy with the `enable_pod_security_policy` variable and later on decides to flag it as `false` then the apply fails due to the reason that "can not be null"

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -112,11 +112,8 @@ resource "google_container_cluster" "primary" {
   enable_kubernetes_alpha     = var.enable_kubernetes_alpha
   enable_tpu                  = var.enable_tpu
 
-  dynamic "pod_security_policy_config" {
-    for_each = var.enable_pod_security_policy ? [var.enable_pod_security_policy] : []
-    content {
-      enabled = pod_security_policy_config.value
-    }
+  pod_security_policy_config {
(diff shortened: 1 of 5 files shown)


s1mark:
If this is fine then it should be changed on all the module types

comment-bot-dev:
Thanks for the PR! 🚀
- ⚠️check_generate_modules
 The modules need to be regenerated. Please run `make_build`. 
 ```bash 
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=.git' /workspace/modules/beta-private-cluster/cluster.tf /tmp/tmp.kEfehV2jhU/workspace/modules/beta-private-cluster/cluster.tf
101,102c101,105
<   pod_security_policy_config {
<     enabled = var.enable_pod_security_policy
---
>   dynamic "pod_security_policy_config" {
>     for_each = var.enable_pod_security_policy ? [var.enable_pod_security_policy] : []
>     content {
>       enabled = pod_security_policy_config.value
>     }
Error: submodule's files generation has not been run, please run the
'make build' command and commit changes
 ```

morgante:
This seems like a bug in the Terraform provider. Please file an issue upstream: https://github.com/hashicorp/terraform-provider-google.

Removing the block should be equivalent to disabling it.

Issue #1000, opened 2021-09-13
Initial Node Pool Metadata


gfthybridlabs:
Hi,

We have policies enabled that mean we must disable serial-port-logging, enable-guest-attributes etc. I add these to the node-pool definitions:-

```
node_pools_metadata = {
  all = {
    serial-port-logging-enable = false
    enable-guest-attributes = false
    disable-legacy-endpoits = true
  }
}
```

But the default-pool that gets created still has those attached (unless I have done it wrong) so the cluster creation fails.

How can I add those to the default-pool before it gets deleted and replaced with my node pools ?

Thanks

morgante:
What value do you have for `remove_default_node_pool`?

We might need to fix the metadata here - https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/cluster.tf.tmpl#L250.

gfthybridlabs:
Thanks. It's set to "true"

morgante:
Thanks. If you're able to fix this a pull request would be appreciated.

gfthybridlabs:
I'll see if I have the skills to do that. Hard coding a "metadata = {map}" object of K/V pairs works fine so just need to expose that as a variable.

PR #1018, opened 2021-10-06
fix: adds metadata to the default node pool

g-awmalik:
This is a fix for #1000 where the default node pool doesn't accept the metadata provided as part of the module calling it. Now metadata included in "all" as well as "default-node-pool" will be applied to the default node pool preventing it from crashing in case of a specific policy constraint.

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -262,10 +262,7 @@ resource "google_container_cluster" "primary" {
         }
       }
 
-      metadata = merge(        
-        local.node_pools_metadata["all"],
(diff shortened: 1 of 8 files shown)


comment-bot-dev:

g-awmalik:
@bharathkkb - can you please review this?

bharathkkb:
@g-awmalik could you check the upgrade scenario from main to this patch if a user has already set `node_pools_metadata=all{k=v}`? I assume its noop but wanted to confirm so we can add upgrade docs if needed.

g-awmalik:
hey @bharathkkb - I already checked that and its noop like you said.

PR #1033, opened 2021-10-22
Add the option to set min_cpu_platform for nodepools

gajicdev:


gnalsa:
Any reason this was closed? It would be very helpful to get this option added as variable to the module.

PR #1038, opened 2021-10-29
Set security relevant settings on default node pool to comply with Sentinel

MajorBreakfast:
This PR adds some security relevant settings to the default node pool that exists temporarily during cluster creation to satisfy Terraform Enterprise Sentinel policies defined by CSO at my project.

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -252,6 +252,13 @@ resource "google_container_cluster" "primary" {
     initial_node_count = var.initial_node_count
 
     node_config {
+      {% if beta_cluster %}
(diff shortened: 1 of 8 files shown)


MajorBreakfast:
@morgante Thank you for the quick feedback!

I've removed the sandbox from the PR to make the tests pass. Sentinel policy enforcement currently only happens for the integrity monitoring, so it's not necessary at this point.

MajorBreakfast:
Change of strategy: Now setting the machine type and image as well based on the value of node pool 0. This should fix it and make it so that even the gvisor can be activated

MajorBreakfast:
Green tests ✌️

morgante (approved):

Issue #737, opened 2020-11-06
Confidential computing


sudipju:
Hello,

We are using this module for the cluster creation for our project. We are planning to use some of our applications to be running with confidential computing. I didn't notice that support yet by this module. Could you please enlighten, in case you have some plan in near future.
Any response on this topic will be well appreciated !

Thanks,
Sudip

bharathkkb:
Hi @sudipju Looks like this will need to be added in the provider first

bharathkkb:
Looks like this was added in the latest release. Happy to review a PR to add this
https://github.com/GoogleCloudPlatform/magic-modules/pull/4136

sudipju:
I observed that latest release is 12.0.0, which is released 24 days back. But, this specific change is merged 17 days back to master. Is there any beta release available after this merge? Not able to find the same. Or am I missing something?

As didn't see this variable in 12.0.0 version.
Thanks in advance.

bharathkkb:
So currently this feature was only added to the provider. In order to use it in the module, we need to make some changes to the source code in this repository. [Here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/601) is an example PR of how to add a new feature to a beta module.

sudipju:
Thanks for the reply. I wait for this to be integrated in the source code repo. Is there any way to know the priority of this issue to be integrated? Or can our need be an influencer of the priority?
Thanks in advance.

bharathkkb:
@sudipju currently its best effort, but happy to review if you can open a PR

AlfredHudson:
Is there any news about this issue?

safag05:
@bharathkkb  Is there an ETA on this?

stenalpjolly:
@bharathkkb Since we have added the beta support, can we close this issue?

PR #1040, opened 2021-10-30
Add beta support for confidential_nodes

stenalpjolly:
Reference: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/737


comment-bot-dev:

autogen/main/cluster.tf.tmpl
```diff
@@ -52,6 +52,14 @@ resource "google_container_cluster" "primary" {
       channel = release_channel.value.channel
     }
   }
+{% if beta_cluster %}
+  dynamic "confidential_nodes" {
+    for_each = local.confidential_node_config
+    content {
+      enabled = confidential_nodes.value.enabled
```
morgante, line 59: For future reference, it's usually more intuitive to just hard-code this (`enabled = true`).

autogen/main/variables.tf.tmpl
```diff
@@ -614,6 +614,12 @@ variable "shadow_firewall_rules_priority" {
 }
 
 {% if beta_cluster %}
+variable "enable_confidential_node" {
```
morgante, line 617:
```suggestion
variable "enable_confidential_nodes" {
```

examples/simple_regional_confidential_private_beta/main.tf
```diff
@@ -0,0 +1,72 @@
+/**
```
morgante, line 1: Thank you for adding a test! However, we're trying to avoid too many examples. Could you instead consolidate this by adding confidential notes to one of the existing examples/tests?

morgante (changes requested):

autogen/main/cluster.tf.tmpl
stenalpjolly: I thought this would be more aligned with existing code like
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/8e92f6ec0d71bfffee9fa3621b55bbbd9091b0d0/autogen/main/cluster.tf.tmpl#L43

morgante: Yes, that's an anti-pattern though.

examples/simple_regional_private_beta/main.tf
```diff
@@ -64,6 +64,8 @@ module "gke" {
     },
   ]
 
+  enable_confidential_node = true
```
morgante, line 67: This isn't correct.

morgante (changes requested):

morgante (approved):

PR #1061, opened 2021-11-15
fix: ignore ASM labels

### LINKED ISSUE #990: GKE modules squash ASM cluster labels
GKE submodules will delete cluster labels on any subsequent run after creation.  

ASM uses 'asmv" and "mesh_id" for observability and mesh management and are created dutifully by ASM module.   Any run with a GKE module (beta-private-cluster-update-variant, beta-private-cluster, private-cluster) will remove ASM cluster resource labels acting as the authorative source.    The module does not allow the use of the lifecycle meta argement to ignore them.    The modules cannot be used in conjuction or with separate terraform plans on the same cluster.

The ASM module never recognizes that the cluster resource labels are removed regardless of the 'enable_cluster_labels' input setting.

GKE should allow lifecycle meta or ignore labels as an 'additive' option.   ASM module should recognize and replace if removed.

morgante:
Fixes #990

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -246,7 +246,7 @@ resource "google_container_cluster" "primary" {
   }
 
   lifecycle {
-    ignore_changes = [node_pool, initial_node_count]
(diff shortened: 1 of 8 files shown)


comment-bot-dev:

bharathkkb (approved):
TIL :)

PR #1070, opened 2021-11-21
[Workload Identity] Set K8S SA token automount default value to true

trennepohl:
## What?
This PR sets the `automount_service_account_token` default value to `true` in the `workload-identity` module

## Why?
While using the module I bumped into an issue where one of my pods was complaining about not being able to mount
```
var/run/secrets/kubernetes.io/serviceaccount/token
```

After some looking around a bit I ended up finding this issue https://github.com/kubernetes/kubernetes/issues/27973, where folks mentioned that it could be an issue when using the `kubernetes` provider for terraform.

The kubernetes has the `automount_service_account_token`  set by default to `true` for a while now.
https://github.com/hashicorp/terraform-provider-kubernetes/pull/1054

So I was wondering if makes sense to follow the pattern here.


comment-bot-dev:

Issue #1045, opened 2021-11-02
Incorrect attribute value type  gpu_partition_size is required.


mike-serchenia:
TF version: 1.0.7
GCP providers: [3.90.1, 3.90.0]
Module version: 17.1.0

No issues if used with GCP provider: 3.89.0

```
Error: Incorrect attribute value type
  on .terraform/modules/gke/modules/beta-private-cluster/cluster.tf line 426, in resource "google_container_node_pool" "pools":
 426:     guest_accelerator = [
 427:       for guest_accelerator in lookup(each.value, "accelerator_count", 0) > 0 ? [{
 428:         type  = lookup(each.value, "accelerator_type", "")
 429:         count = lookup(each.value, "accelerator_count", 0)
 430:         }] : [] : {
 431:         type  = guest_accelerator["type"]
 432:         count = guest_accelerator["count"]
 433:       }
 434:     ]
    ├────────────────
    │ each.value is map of string with 18 elements

Inappropriate value for attribute "guest_accelerator": element 0: attribute
"gpu_partition_size" is required.
```

morgante:
If this is a regression in the provider, please open an issue there.

mike-serchenia:
> If this is a regression in the provider, please open an issue there.

looks like it isn't
```
IMPROVEMENTS:
container: added node_config.0.guest_accelerator.0.gpu_partition_size field to google_container_node_pool (#10339)
```
https://github.com/hashicorp/terraform-provider-google/pull/10339

morgante:
@mike-serchenia If your config was working in 3.89.0 but no longer in 3.90.0 that's a regression. /cc @rileykarson.

rileykarson:
See https://github.com/hashicorp/terraform-provider-google/issues/10415 for discussion, there's no way around this one unfortunately- we're using a half-baked SDK feature that was introduced in `0.12` to avoid breaking users, and it makes optional fields behave unusually. The only solution would be to yank the field, unfortunately. Note that https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#guest_accelerator documents that the field is using that feature, and links to the SDK's guide on how it may differ from other fields.

morgante:
@rileykarson Have we considered not using the new feature in 0.39 or at least pulling the field from 0.39 and instead only having it in 4.0? This seems to be clearly backwards-incompatible (though it is an SDK problem).

[...]

PR #1072, opened 2021-11-22
feat: Add support for `gpu_partition_size`

stanley98yu:
Fix for #1045. The `gpu_partition_size` field behaves unexpectedly and is required, not optional, in provider versions 3.90+, so we default to `null` as a workaround.

I was also thinking about adding a test, but I'm not sure how to approach it. The accelerator types supported by the machine and location in the example `node_pool` do not allow GPU partitioning (I specifically tried `nvidia-tesla-k80` and `nvidia-tesla-a100`). Would this require a new example?


morgante:
> I was also thinking about adding a test, but I'm not sure how to approach it. The accelerator types supported by the machine and location in the example node_pool do not allow GPU partitioning (I specifically tried nvidia-tesla-k80 and nvidia-tesla-a100). Would this require a new example?

It would be fine to change the existing machine/location to one which allows GPU partitioning (just need to adjust tests as well).

comment-bot-dev:

morgante (approved):

PR #1077, opened 2021-11-29
feat: Make auto_provisioning_defaults a non-beta feature

### LINKED ISSUE #1008: add service_account how parameter on variable cluster_autoscaling
Please add service_account how parameter on variable cluster_autoscaling

stanley98yu:
Fixes #1008 

This PR simply makes the auto-provisioning defaults block not beta-exclusive, but I also want to add the `min_cpu_platform` argument in the block as specified [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#nested_auto_provisioning_defaults) since it seems we added the feature for default node pools in #1057. However, I don't understand why we pull the value from the first node pool using `lookup(var.node_pools[0],...)` because, from what I understand, `min_cpu_platform` is a cluster-level variable, so it feels weird to read from a specific node pool's arguments, like `pool-01`, even though it will apply to other node pools as well. Shouldn't it be its own variable and read in like `var.min_cpu_platform` or something similar?

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -86,15 +86,18 @@ resource "google_container_cluster" "primary" {
 
   cluster_autoscaling {
     enabled = var.cluster_autoscaling.enabled
-{% if beta_cluster %}
     dynamic "auto_provisioning_defaults" {
       for_each = var.cluster_autoscaling.enabled ? [1] : []
 
       content {
         service_account = local.service_account
         oauth_scopes = local.node_pools_oauth_scopes["all"]
+{% if beta_cluster %}
+        min_cpu_platform = lookup(var.node_pools[0], "min_cpu_platform", "")
+{% endif %}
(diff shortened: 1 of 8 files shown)


morgante:
> Shouldn't it be its own variable and read in like var.min_cpu_platform or something similar?

It's actually not a cluster-level variable. `min_cpu_platform` is part of the node pool config for particular nodes.

The cluster-level setting is for *node pool autoprovisioning*. ie. what setting to use for any autoprovisioned node pools. We've made the decision to have your first node pool be your "default" node pool and the settings from it are mirrored to any autoprovisioned node pools.

comment-bot-dev:

stanley98yu:
That makes sense. Thanks for the explanation! The `min_cpu_platform` for `auto_provisioning_defaults` mirror the first node pool's value as expected then.

morgante (approved):

Issue #1118, opened 2022-01-07
Allow setting network tags on default node pool


tomasgareau:
I'm trying to bring up a [`safer-cluster`](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/modules/safer-cluster) in a Shared VPC network roughly based on the [`terraform-example-foundation/3-networks`](https://github.com/terraform-google-modules/terraform-example-foundation/tree/master/3-networks) approach.

In this VPC network, access to private Google APIs is configured with:
* [DNS entries](https://github.com/terraform-google-modules/terraform-example-foundation/blob/0ee39ceff2dce8c3a3b61d59bb1d45a4d0687e74/3-networks/modules/base_shared_vpc/dns.tf#L57-L84) that point `*.googleapis.com` to `private.googleapis.com` (for resources without an external IP address, i.e., my `safer-cluster` nodes)
* [a firewall rule](https://github.com/terraform-google-modules/terraform-example-foundation/blob/0ee39ceff2dce8c3a3b61d59bb1d45a4d0687e74/3-networks/modules/base_shared_vpc/firewall.tf#L46-L71) that allows egress to private.googleapis.com (i.e., 199.36.153.8/30) for any resources that have the network tag `allow-google-apis`

Like in the [`terraform-example-foundation-app`](https://github.com/GoogleCloudPlatform/terraform-example-foundation-app/blob/d549737be96de4660087e8210aab1f2b4f88e6f4/5-infrastructure/modules/base-env/boa-gke.tf#L130), I've added the `allow-google-apis` network tag to the `node_pools_tags` variable.

What I'd expect to happen is:

1. The `google_container_cluster.primary` resource gets created
1. This creates a default node pool, which immediately gets deleted (as configured by the `safer-cluster` module)
1. The node pool I configured now gets created (with the appropriate network tags)
1. My cluster is ready to go

Instead, however, this seems to be happening:

1. The `google_container_cluster.primary` resource gets created
1. The default node pool gets created and tries to register itself with the control plane
1. The default node pool cannot register itself with the control plane
1. The `create` operation finally times out and the cluster errors out with "Error waiting for creating GKE cluster: All cluster resources were brought up, but: only 0 nodes out of 1 have registered"

In my firewall logs, I can see that the default pool is trying to reach private.googleapis.com (i.e., DNS is working as expected). However, since I cannot add the `allow-google-apis` network tag to this pool, this egress gets denied:

[...]

PR #1123, opened 2022-01-12
Allow users to specify network tags for the default node pool

tomasgareau:
Addresses #1118.

This will apply the following tags to the default node pool:
* the [`cluster_network_tag`](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/c2dbc6f8aabbfc722a56e6a56e7121e3fbe6826a/autogen/main/main.tf.tmpl#L167) (`gke-${var.name}`)
* the pool variant of the `cluster_network_tag` (`gke-${var.name}-default-pool`)
* user-specified tags in `var.node_pools_tags["all"]`
* user-specified tags in `var.node_pools_tags["default-pool"]`

This may be needed for private clusters in VPC networks with a "default-deny-all" firewall rule that use network tags to define exceptions for e.g., egress to the `master_ipv4_cidr_block`.

Two quick notes/questions:
* I deviated slightly from @morgante's [suggestion](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1118#issuecomment-1009633635) in the related issue (adding the network tags from the first pool in the cluster config like for the [service account](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/14a0536bbcfeb89dc1af21f8fef0cb46affdc52e/autogen/main/cluster.tf.tmpl#L289)): `node_pools_tags` is a map-of-lists unlike `node_pools`, which is a list-of-maps. The "first" pool here wouldn't be well defined -- I think the approach I went with is more natural in any case.
* Are there any existing tests where we'd like to add coverage for this?
  * It's unfortunately a bit difficult for me to spin up tests as described in [CONTRIBUTING.md](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#test-environment) (we have an org policy that blocks the creation of service account keys), so I'm afraid the only testing I have of the above is manually spinning up an instance in our existing env and confirming that it contained the desired tags.
  * For some tests it seems like I could add an `it "has the expected network tags" do` block to the default pool checks (e.g., [simple_regional](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/14a0536bbcfeb89dc1af21f8fef0cb46affdc52e/test/integration/simple_regional/controls/gcloud.rb#L78))

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -288,6 +288,13 @@ resource "google_container_cluster" "primary" {
 
       service_account = lookup(var.node_pools[0], "service_account", local.service_account)
 
+      tags = concat(
+        lookup(local.node_pools_tags, "default_values", [true, true])[0] ? [local.cluster_network_tag] : [],
+        lookup(local.node_pools_tags, "default_values", [true, true])[1] ? ["${local.cluster_network_tag}-default-pool"] : [],
+        lookup(local.node_pools_tags, "all", []),
+        lookup(local.node_pools_tags, "default-pool", []),
+      )
+
       dynamic "workload_metadata_config" {
         for_each = local.cluster_node_metadata_config
 
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -292,7 +292,7 @@ resource "google_container_cluster" "primary" {
         lookup(local.node_pools_tags, "default_values", [true, true])[0] ? [local.cluster_network_tag] : [],
         lookup(local.node_pools_tags, "default_values", [true, true])[1] ? ["${local.cluster_network_tag}-default-pool"] : [],
         lookup(local.node_pools_tags, "all", []),
-        lookup(local.node_pools_tags, "default-pool", []),
+        lookup(local.node_pools_tags, var.node_pools[0].name, []),
       )
 
       dynamic "workload_metadata_config" {
--- a/cluster.tf
+++ b/cluster.tf
@@ -156,7 +156,7 @@ resource "google_container_cluster" "primary" {
         lookup(local.node_pools_tags, "default_values", [true, true])[0] ? [local.cluster_network_tag] : [],
         lookup(local.node_pools_tags, "default_values", [true, true])[1] ? ["${local.cluster_network_tag}-default-pool"] : [],
         lookup(local.node_pools_tags, "all", []),
-        lookup(local.node_pools_tags, "default-pool", []),
+        lookup(local.node_pools_tags, var.node_pools[0].name, []),
       )
 
       dynamic "workload_metadata_config" {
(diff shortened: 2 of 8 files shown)


comment-bot-dev:

morgante:
Thanks for working on this!

> I deviated slightly from @morgante's suggestion in the related issue (adding the network tags from the first pool in the cluster config like for the service account): node_pools_tags is a map-of-lists unlike node_pools, which is a list-of-maps. The "first" pool here wouldn't be well defined -- I think the approach I went with is more natural in any case.

I'd like to avoid introducing a new "default-pool" concept because we already have precedent for using the first node pool as the default pool ([for example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/beta-private-cluster/cluster.tf#L97). In this case, what we should do is retrieve the name of the first node pool (from `var.node_pools`) and use the network tags associated with that node pool (in `var. node_pools_tags`).

autogen/main/cluster.tf.tmpl
```diff
@@ -288,6 +288,13 @@ resource "google_container_cluster" "primary" {
 
       service_account = lookup(var.node_pools[0], "service_account", local.service_account)
 
+      tags = concat(
+        lookup(local.node_pools_tags, "default_values", [true, true])[0] ? [local.cluster_network_tag] : [],
+        lookup(local.node_pools_tags, "default_values", [true, true])[1] ? ["${local.cluster_network_tag}-default-pool"] : [],
+        lookup(local.node_pools_tags, "all", []),
+        lookup(local.node_pools_tags, "default-pool", []),
```
morgante, line 295: Thanks for working on this!

> I deviated slightly from @morgante's suggestion in the related issue (adding the network tags from the first pool in the cluster config like for the service account): node_pools_tags is a map-of-lists unlike node_pools, which is a list-of-maps. The "first" pool here wouldn't be well defined -- I think the approach I went with is more natural in any case.

I'd like to avoid introducing a new "default-pool" concept because we already have precedent for using the first node pool as the default pool ([example](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/beta-private-cluster/cluster.tf#L97)). In this case, what we should do is retrieve the name of the first node pool (from `var.node_pools`) and use the network tags associated with that node pool (in `var.node_pools_tags`).

morgante (changes requested):

autogen/main/cluster.tf.tmpl
tomasgareau: Ah, I see -- makes sense! I've changed the `"default-pool"` lookup to `var.node_pools[0].name` -- from what I can see
* node_pools[0] should always exist
* node_pools entries are assumed to have the attribute `name`

so I have no error checking for the cases where either of these aren't true. Let me know if either is not the case.

Did we want to keep the `"${local.cluster_network_tag}-default-pool"` tag?

morgante: Looks good, yeah you shouldn't need error checking.

> Did we want to keep the "${local.cluster_network_tag}-default-pool" tag?

Yes, I think it's helpful.

morgante (approved):

PR #1138, opened 2022-01-31
chore: switch region for gpu quota

bharathkkb:
tests seem to be failing due to a quota issue for GPUs


comment-bot-dev:

g-awmalik:
Can this be done in a vars files specific to this fixture instead of hardcoding in the fixture itself?

bharathkkb:
@g-awmalik We generally prefer to hardcode as much as possible rather than having var files

g-awmalik (approved):
LGTM

PR #1174, opened 2022-03-11
Add support for image streaming/GCFS

### LINKED ISSUE #1149: Support configuring GCFS / image streaming
### TL;DR

Support configuring GCFS / enabling image streaming by introducing a new variable that sets the required parameter in the node configuration.

### Terraform Resources

```markdown
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#gcfs_config
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#node_config
```

### Detailed design

```markdown
Image streaming can be enabled for clusters and/or node pools by adding a block

gcfs_config {
  enabled = true/false
}
```
to the node config. 

I propose to add a variable `enable_image_streaming`/`enable_gcfs` or similar to allow setting this block. I would default it to `false` as that's the current behaviour of the module.

Potentially, setting this variable to `true` could also automatically enable the `containerfilesystem.googleapis.com` API, though this might be out of scope.
```


### Additional information

Currently, there's no way to enable image streaming on a node pool that's managed by this module, as the absence of this block will force the node pool to be re-created without it even if image streaming is enabled manually.

Adding this block always as either true/false might also solve some other issues related to this block, e.g. https://github.com/hashicorp/terraform-provider-google/issues/11017.

If this is an acceptable feature request, I'm happy to provide a PR :)

### PR DESCRIPTION
**Pull request by sasha-s**
Fixes #1149

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ module "gke" {
       disk_size_gb              = 100
       disk_type                 = "pd-standard"
       image_type                = "COS_CONTAINERD"
+      enable_gcfs               = false
       auto_repair               = true
       auto_upgrade              = true
       service_account           = "project-service-account@<PROJECT ID>.iam.gserviceaccount.com"
@@ -235,6 +236,7 @@ The node_pools variable takes the following parameters:
 | disk_size_gb | Size of the disk attached to each node, specified in GB. The smallest allowed disk size is 10GB | 100 | Optional |
 | disk_type | Type of the disk attached to each node (e.g. 'pd-standard' or 'pd-ssd') | pd-standard | Optional |
 | effect | Effect for the taint | | Required |
+| enable_gcfs | Google Container File System (gcfs) has to be enabled for image streaming to be active. Needs image_type to be set to COS_CONTAINERD. | false | Optional |
 | enable_integrity_monitoring | Enables monitoring and attestation of the boot integrity of the instance. The attestation is performed against the integrity policy baseline. This baseline is initially derived from the implicitly trusted boot image when the instance is created. | true | Optional |
 | enable_secure_boot | Secure Boot helps ensure that the system only runs authentic software by verifying the digital signature of all boot components, and halting the boot process if signature verification fails. | false | Optional |
 | gpu_partition_size | Size of partitions to create on the GPU | null | Optional |
--- a/autogen/main/README.md
+++ b/autogen/main/README.md
@@ -108,6 +108,7 @@ module "gke" {
       disk_size_gb              = 100
       disk_type                 = "pd-standard"
       image_type                = "COS_CONTAINERD"
+      enable_gcfs               = false
       auto_repair               = true
       auto_upgrade              = true
       service_account           = "project-service-account@<PROJECT ID>.iam.gserviceaccount.com"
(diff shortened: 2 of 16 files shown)

### CODE REVIEW

**Comment by comment-bot-dev** [comment]
Thanks for the PR! 🚀
- ⚠️check_generate_modules
 The modules need to be regenerated. Please run `make_build`. 
 ```bash 
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=.git' /workspace/modules/private-cluster/cluster.tf /tmp/tmp.zh0eorg3D5/workspace/modules/private-cluster/cluster.tf
146c146
<       image_type       = lookup(var.node_pools[0], "enable_gcfs", false) ? "COS_CONTAINERD" : lookup(var.node_pools[0], "image_type", "COS_CONTAINERD")
---
>       image_type       = lookup(var.node_pools[0], "image_type", "COS_CONTAINERD")
152,155d151
<       gcfs_config {
<         enabled = lookup(var.node_pools[0], "enable_gcfs", false)
<       }
< 
283c279
<     image_type       = lookup(each.value, "enable_gcfs", false) ? "COS_CONTAINERD" : lookup(each.value, "image_type", "COS_CONTAINERD")
---
>     image_type       = lookup(each.value, "image_type", "COS_CONTAINERD")
285,289c281
<     min_cpu_platform = lookup(each.value, "min_cpu_platform", "")
<     gcfs_config {
<       enabled = lookup(each.value, "enable_gcfs", false)
<     }
< 
---
>     min_cpu_platform = lookup(var.node_pools[0], "min_cpu_platform", "")
Error: submodule's files generation has not been run, please run the
'make build' command and commit changes
 ```

**Review by bharathkkb — COMMENTED** [review]
Thanks for the PR @sasha-s 
Can we add this to the nodepool variable docs https://github.com/terraform-google-modules/terraform-google-kubernetes-engine#node_pools-variable

**Comment by sasha-s** [comment]
@bharathkkb Anything I should add to the PR?

**Comment by lauraseidler** [comment]
thanks for taking care of this (I opened the original issue) - if you have the time, maybe we can also add it to the safer cluster variants? or I can also do that in a follow up

**Comment by bharathkkb** [comment]
@lauraseidler I will leave it upto @sasha-s, otherwise happy to accept a follow up too.

**Comment by sasha-s** [comment]
I do not have access  to the CI build, what is going on there?
Also, it would be great if someone with better knowledge of terraform could add streaming to safer cluster variants.

BTW, should it be 
```
min_cpu_platform = lookup(each.value, "min_cpu_platform", "")
```
here
https://sourcegraph.com/github.com/terraform-google-modules/terraform-google-kubernetes-engine@bbd9b770987b6e8e502ce3747ccc6ce0a96e79c1/-/blob/cluster.tf?L268:5&subtree=true ?

(instead of `var.node_pools[0]`)

sasha-s:
Ping.

bharathkkb:
@sasha-s CI error looks like a flake- I have re triggered it.

> BTW, should it be min_cpu_platform = lookup(each.value, "min_cpu_platform", "")
here https://sourcegraph.com/github.com/terraform-google-modules/terraform-google-kubernetes-engine@bbd9b770987b6e8e502ce3747ccc6ce0a96e79c1/-/blob/cluster.tf?L268:5&subtree=true ?
(instead of var.node_pools[0])

Yeah we should switch that to each.value. Otherwise it will set the value for the first nodepool for all nodepools.

sasha-s:
Fixed the min_cpu_platform lookup in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1207

PR #1183, opened 2022-03-17
Apply applicable ASM_OPTS in config_map

frits-v:
[In autopilot clusters,] subsequent terraform runs overwrite the `ASM_OPTS: "CNI=on"` value in the asm_option config map. This prevents new or restating workloads from being scheduled as the proxy tries to cap add NET_ADMIN on start. This PR conditionally adds the expected setting.

### DIFF
--- a/modules/asm/main.tf
+++ b/modules/asm/main.tf
@@ -45,6 +45,7 @@ resource "kubernetes_config_map" "asm_options" {
 
   data = {
     multicluster_mode = var.multicluster_mode
+    ASM_OPTS          = var.enable_cni ? "CNI=on" : null
   }
 
   depends_on = [google_gke_hub_membership.membership, google_gke_hub_feature.mesh]


morgante:
@Monkeyanator PTAL

comment-bot-dev:
Thanks for the PR! 🚀
- ⚠️check_terraform
 ```bash 
Running terraform fmt
modules/asm/main.tf
--- old/modules/asm/main.tf
+++ new/modules/asm/main.tf
@@ -45,7 +45,7 @@
 
   data = {
     multicluster_mode = var.multicluster_mode
-    ASM_OPTS = var.enable_cni ? "CNI=on" : null
+    ASM_OPTS          = var.enable_cni ? "CNI=on" : null
   }
 
   depends_on = [google_gke_hub_membership.membership, google_gke_hub_feature.mesh]
 ```

modules/asm/main.tf
```diff
@@ -45,6 +45,7 @@ resource "kubernetes_config_map" "asm_options" {
 
   data = {
     multicluster_mode = var.multicluster_mode
+    ASM_OPTS          = var.enable_cni ? "CNI=on" : null
```
Monkeyanator, line 48: nit: should this be 

```suggestion
    ASM_OPTS = var.enable_cni ? "CNI=on" : ""
```

Monkeyanator (approved):

modules/asm/main.tf
morgante: Do we want it to be set to `""` or excluded entirely? Usually I think excluding is better.

frits-v: When set to null the value is not applied to the configMap. Empty value could trigger unintended behavior.

Issue #1181, opened 2022-03-17
ASM module not deferring data source read to apply time


bharathkkb:
### TL;DR

When GKE cluster and ASM module are created in the same config and GKE module inputs are passed to ASM module, [the data source read](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/774fc12df6cb9f801caf72cd0ac6dfdbe723da46/modules/asm/main.tf#L28-L32) for cluster is not deferred to apply time as project_id, cluster name and location are known at plan time. This results in an error as cluster does not exist yet.

```
 Error: Invalid template interpolation value
│ 
│   on ../../../modules/asm/hub.tf line 21, in resource "google_gke_hub_membership" "membership":
│   21:   membership_id = "${data.google_container_cluster.asm.name}-membership"
│     ├────────────────
│     │ data.google_container_cluster.asm.name is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
╵
╷
│ Error: Invalid template interpolation value
│ 
│   on ../../../modules/asm/hub.tf line 24, in resource "google_gke_hub_membership" "membership":
│   24:       resource_link = "//container.googleapis.com/${data.google_container_cluster.asm.id}"
│     ├────────────────
│     │ data.google_container_cluster.asm.id is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
```

Our current tests did not catch this as a random_string was passed as cluster name deferring the read to apply time 
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/774fc12df6cb9f801caf72cd0ac6dfdbe723da46/test/fixtures/simple_zonal_with_asm/example.tf#L25

Module depends on also throws an error due to gcloud module

```
│ Error: Invalid count argument
│ 
│   on .terraform/modules/example.asm.cpr/main.tf line 57, in resource "random_id" "cache":
│   57:   count = (!local.skip_download) ? 1 : 0
```

We can look into using a computed attribute to defer this to apply time. 

### Terraform Configuration

```hcl
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples/simple_zonal_with_asm with a static cluster name
```

### Terraform Version

```sh
v1.0.5
```

Monkeyanator:
[...]

PR #1188, opened 2022-03-25
[ASM module] Document ASM module data source workaround

Monkeyanator:
Documents workaround for https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1181

### DIFF
--- a/modules/asm/README.md
+++ b/modules/asm/README.md
@@ -7,8 +7,8 @@ This module installs [Anthos Service Mesh](https://cloud.google.com/service-mesh
 There are a few prerequisites to using this module that can be done either through Terraform or manually:
 
 1. Enable the `mesh.cloud.googleapis.com` service
-1. Enable the `servicemesh` feature on the cluster hub
-1. Register target cluster to the servicemesh-enabled hub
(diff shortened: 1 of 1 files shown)


morgante (commented):
Honestly I think we should just fix the module. It would be faster/simpler than this.

comment-bot-dev:

Monkeyanator:
> Documents workaround for #1181

Opened https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1189, if that approach looks good will close this.

bharathkkb:
@morgante I am thinking it could be worth documenting this (without the workaround code) incase the user is using plain resources or a different module? The error surfaced is not very actionable.

morgante:
> @morgante I am thinking it could be worth documenting this (without the workaround code) incase the user is using plain resources or a different module? The error surfaced is not very actionable.

Good point, but I think we should document it just by saying you must make the name dependent on the GKE cluster resource. I don't think we should use the `locals` hack as it's messy/unnecessary.

Monkeyanator:
Closing in favor of https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1189

PR #1243, opened 2022-05-11
fix: trim trailing dash from gcp SA name in WorkloadIdentity

sbienkow-ninja:
As per noted regexp, the service account name cannot end with a dash.

This can happen when the name is over 30 characters long
and so a substring is extracted, but the 30th character happens to be a
dash.

### DIFF
--- a/modules/workload-identity/main.tf
+++ b/modules/workload-identity/main.tf
@@ -15,9 +15,9 @@
  */
 
 locals {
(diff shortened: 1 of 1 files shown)


comment-bot-dev:

bharathkkb (approved):
Thanks for the PR @sbienkow-ninja 
Test failure looks unrelated

PR #1266, opened 2022-05-24
Fix: (ASM) Add skip_download pass through variable

frits-v:
In some terraform execution environments the `gcloud` binary is available but the `kubectl` component isn't. The `google-cloud-sdk` directory might not be writable by the terraform execution user, leading to a filesystem permission error.

This PR adds the following variables which are passed on to the kubectl-wrapper module:
* `skip_download` allows to force downloading all components in the module cache directory.
* `use_tf_google_credentials_env_var` allows to use the terraform service account


comment-bot-dev:

PR #1293, opened 2022-06-15
feat: source node pools' auto_upgrade configuration from the GKE default config

MrwanBaghdad:
Allows the module to use the `default_auto_upgrade` to infer if the auto_upgrade is enabled for the node pool. This is in line with how the `auto_upgrade` config is [calculated in the `node_pool` resource ](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/compare/master...MrwanBaghdad%3Amaster?body=&expand=1)

```
  management {
    auto_repair  = lookup(each.value, "auto_repair", true)
    auto_upgrade = lookup(each.value, "auto_upgrade", local.default_auto_upgrade)
  }
```

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -530,7 +530,7 @@ resource "google_container_node_pool" "pools" {
 
   cluster = google_container_cluster.primary.name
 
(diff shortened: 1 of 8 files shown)


comment-bot-dev:

MrwanBaghdad:
@bharathkkb Can you please take a look?

Issue #1290, opened 2022-06-13
missing `enable-managed-prometheus` for GKE cluster


hixichen:
### TL;DR

GKE setting:
https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed#config-mgd-collection

kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/prometheus-engine/v0.4.1/manifests/setup.yaml

kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/prometheus-engine/v0.4.1/manifests/operator.yaml

but, did not find configuration from terraform config.

JCPistell:
Possible duplicate of #1167

Issue #1167, opened 2022-03-05
Support Managed Service for Prometheus


mbarrien:
### TL;DR

Managed Service for Prometheus was recently [announced for general availability](https://cloud.google.com/blog/products/devops-sre/easy-managed-prometheus-metrics-service-for-kubernetes). Both the Terraform provider and this module should support it.

### Terraform Resources
Corresponding issue opened with Google provider to support this. 
https://github.com/hashicorp/terraform-provider-google/issues/11224

bharathkkb:
Thanks for the FR @mbarrien 
We will wait for support in the provider.

kriswuollett:
Not stale.

JCPistell:
FYI - As of the 4.25.0 release this is now available in the beta provider.

https://github.com/hashicorp/terraform-provider-google-beta/releases/tag/v4.25.0

bharathkkb:
fixed by https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1307

PR #1298, opened 2022-06-20
Support Managed Prometheus

abousias:
With the release of the `4.25.0` "beta" provider, it is now possible to enable Managed Prometheus. 
With this we can now enable this feature in the beta submodules.

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -89,16 +89,12 @@ resource "google_container_cluster" "primary" {
   }
   monitoring_service = local.cluster_telemetry_type_is_set || local.logmon_config_is_set ? null : var.monitoring_service
   dynamic "monitoring_config" {
-    for_each = length(var.monitoring_enabled_components) > 0 ? [1] : []
+    for_each = length(var.monitoring_enabled_components) > 0 || var.enable_managed_prometheus ? [1] : []
 
     content {
       enable_components = var.monitoring_enabled_components
-      dynamic "managed_prometheus" {
-        for_each = var.enable_managed_prometheus ? [1] : []
-
-        content {
-          enabled = var.enable_managed_prometheus
-        }
+      managed_prometheus {
+        enabled = var.enable_managed_prometheus
       }
     }
   }
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -99,7 +99,7 @@ resource "google_container_cluster" "primary" {
     for_each = var.enable_managed_prometheus ? [1] : []
 
     content {
-      managed_prometheus =   {
+      managed_prometheus {
         enabled = var.enable_managed_prometheus
       }
     }
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -95,6 +95,15 @@ resource "google_container_cluster" "primary" {
       enable_components = var.monitoring_enabled_components
     }
   }
+  dynamic "monitoring_config" {
+    for_each = var.enable_managed_prometheus ? [1] : []
+
+    content {
+      managed_prometheus =   {
+        enabled = var.enable_managed_prometheus
+      }
+    }
+  }
 {% else %}
   logging_service    = var.logging_service
   monitoring_service = var.monitoring_service
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -93,14 +93,12 @@ resource "google_container_cluster" "primary" {
 
     content {
       enable_components = var.monitoring_enabled_components
-    }
-  }
-  dynamic "monitoring_config" {
-    for_each = var.enable_managed_prometheus ? [1] : []
+      dynamic "managed_prometheus" {
+        for_each = var.enable_managed_prometheus ? [1] : []
 
-    content {
-      managed_prometheus {
-        enabled = var.enable_managed_prometheus
+        content {
+          enabled = var.enable_managed_prometheus
+        }
       }
     }
   }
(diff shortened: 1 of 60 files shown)


comment-bot-dev:

abousias:
Resolves #1290 and #1167

autogen/main/cluster.tf.tmpl
```diff
@@ -95,6 +95,15 @@ resource "google_container_cluster" "primary" {
       enable_components = var.monitoring_enabled_components
     }
   }
+  dynamic "monitoring_config" {
```
bharathkkb, line 98: There is an existing dynamic monitoring_config, I think we can nest `managed_prometheus` as a dynamic under that.

bharathkkb (changes requested):
Thanks for the PR @abousias

autogen/main/cluster.tf.tmpl
abousias: Thanks for quick response and the review comment @bharathkkb! I kept it separate to have a separate `for_each` check as in one case we have `var.monitoring_enabled_components=[]` but we still want managed prometheus enabled. I didn't think of nesting with another dynamic! I will try that!

abousias:
@bharathkkb the nested dynamic how the way I have it now, it seems to not be working as expected. With the `var.enable_managed_prometheus` set to `true`, the service didn't get enabled and the tfstate is like this:

```json
...
"monitoring_config": [
  {
    "enable_components": [
      "SYSTEM_COMPONENTS"
    ],
    "managed_prometheus": []
  }
],
...
```

I am looking into it.

autogen/main/cluster.tf.tmpl
abousias: Interestingly switching to a static block:

https://github.com/abousias/terraform-google-kubernetes-engine/blob/c5170bb58cab6706377baf24e58cadc28e685205/autogen/main/cluster.tf.tmpl#L91-L100

It gives an api error:

image

abousias: I am having the error, with the reverted code too. The issue didn't appear because in my test I was applying the change to an existing cluster. Now I am getting this with a brand new cluster.

https://github.com/abousias/terraform-google-kubernetes-engine/blob/c5170bb58cab6706377baf24e58cadc28e685205/autogen/main/cluster.tf.tmpl#L81

I am looking into it.

jroiseux: Hey @abousias I have also been working on this. I was able to fix that error by adding a check for the enable managed prometheus variable in the main.tf file.

```logmon_config_is_set      = length(var.logging_enabled_components) > 0 || length(var.monitoring_enabled_components) > 0 || var.monitoring_enable_managed_prometheus```

I was also able to get it working with nested dynamic blocks. Feel free to have a look at what I did if it helps.
https://github.com/jroiseux/terraform-google-kubernetes-engine/tree/managed-prometheus

abousias: Hey @jroiseux I was about to check exactly the same conditional 😄! I will take a look and copy! Many thanks for the help! Nice to know that it will work 😉!

abousias: @bharathkkb ready for re-review. It works fine now.

comment-bot-dev:
@abousias

autogen/main/cluster.tf.tmpl
abousias: ⚠️ I discovered that the current version is trying to disable `SYSTEM_COMPONENTS`, here is the plan against an existing cluster:

image

Whereas the separate block one leaves the `enabled_components` untouched:

image

This is because the `dynamic` is "forced" to produce the block but the `var.monitoring_enabled_components` is `[]`. My existing cluster is also based on the main module instead of the `beta` ones. I am checking if that is an issue for me.

@jroiseux this is most probably happening with your version, isn't it?

jroiseux: @abousias Yes, it happens on my version too.

This is what the statefile looks like when I use the current module for "beta-private-cluster" (without the changes from this PR) without setting `var.monitoring_enabled_components` :

Screenshot from 2022-06-23 09-16-34

I think it might be an issue with the default value. As `enabled_components` seems to end up as `["SYSTEM_COMPONENTS"]`, even though the default is `[]`.

abousias: @bharathkkb I am reverting back to the original implementation, leaving TF to merge the blocks in one well after the evaluation of the the `for_each`s.  As @jroiseux also points out, the other option would be to change the **default** value of `var.monitoring_enabled_components` to something like `[ "SYSTEM_COMPONENTS" ]` which I don't know if it is a good idea.

bharathkkb:
Thanks for working on this @abousias and collaborating with @jroiseux 
I merged https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1307

abousias:
> Thanks for working on this @abousias and collaborating with @jroiseux 
> I merged https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1307

Thank you @bharathkkb!

abousias:
> > Thanks for working on this @abousias and collaborating with @jroiseux 
> > I merged https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1307
> 
> Thank you @bharathkkb! 

@bharathkkb and @jroiseux was the issue, where the `SYSTEM_COMPONENTS` were getting disabled solved? Because I see now the code on #1307 similar to a previous version of this. I think this is a blocker!

jroiseux:
> > > Thanks for working on this @abousias and collaborating with @jroiseux
> > > I merged #1307
> > 
> > 
> > Thank you @bharathkkb!
> 
> @bharathkkb and @jroiseux was the issue, where the `SYSTEM_COMPONENTS` were getting disabled solved? Because I see now the code on #1307 similar to a previous version of this. I think this is a blocker!

Yes, it is no longer trying to disable it because of `enable_components = length(var.monitoring_enabled_components) > 0 ? var.monitoring_enabled_components : null`. So when it's set to null, it doesn't try to modify the existing value.

Issue #1305, opened 2022-06-23
cloud logging/monitoring for non-beta submodules


rwong2888:
### TL;DR

I see [1054](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1054) fixed via [1028](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1028) for beta clusters.

Can we promote the same in the other modules, particularly [private-cluster](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest/submodules/private-cluster)?

PR #1328, opened 2022-07-20
fix/1305 logging and monitoring config

rwong2888:
fix [1305](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1305)

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -79,7 +79,7 @@ resource "google_container_cluster" "primary" {
     }
   }
   # only one of logging/monitoring_service or logging/monitoring_config can be specified
-  logging_service    = local.cluster_telemetry_type_is_set || local.logmon_config_is_set ? null : var.logging_service
+  logging_service = local.cluster_telemetry_type_is_set || local.logmon_config_is_set ? null : var.logging_service
   dynamic "logging_config" {
     for_each = length(var.logging_enabled_components) > 0 ? [1] : []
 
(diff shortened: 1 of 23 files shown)


comment-bot-dev:
@rwong2888

rwong2888:
@Jberlinsky , @bharathkkb mind taking a look at this?

autogen/main/cluster.tf.tmpl
```diff
@@ -119,6 +121,25 @@ resource "google_container_cluster" "primary" {
   {% endif %}
 {% endif %}
   {% if autopilot_cluster != true %}
+  {% if beta_cluster != true %}
+  # only one of logging/monitoring_service or logging/monitoring_config can be specified
+  logging_service    = local.logmon_config_is_set ? null : var.logging_service
```
bharathkkb, line 126: We already have this in 
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/4bf0011202667b6efe8a0b5aea2910f1c3250968/autogen/main/cluster.tf.tmpl#L81-L105
We should just change the conditional.

bharathkkb (changes requested):
Thanks for the PR @rwong2888

autogen/main/cluster.tf.tmpl
rwong2888: I wasn't sure about "cluster_telemetry_type_is_set", "managed_prometheus", and "monitoring_enable_managed_prometheus" as I was primarily looking for "logging_config", but I can revise the PR to include those as well if you think they are ready to promote out of beta @bharathkkb

rwong2888: @bharathkkb, `cluster_telemetry` and `managed_prometheus` are still in beta as per below, so it does not look possible to simply modify the conditional. Please advise.
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster

bharathkkb: @rwong2888 Can we try doing it inline and guard the dynamic block for managed_prometheus with a conditional. Similarly for the `local.cluster_telemetry_type_is_set || local.logmon_config_is_set` conditional.

PR #1373, opened 2022-08-24
feat: enable PoCo referential_rules for ACM

### LINKED ISSUE #1369: ACM submodule missing variable
### TL;DR

The variable referential_rules_enabled is not listed in the acm submodule : https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/acm/feature.tf#L56

### Expected behavior

referential_rules_enabled is required to have 2 additional constrainttemplates installed. In the acm example, it is used: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/acm/feature.tf#L56
However in the sub-module it is missing.

### Observed behavior

Without this variable set to true, the constrainttemplates  asmpeerauthnmeshstrictmtls and asmauthzpolicydefaultdeny are missing.

### Terraform Configuration

```hcl
policy_controller {
      enabled                    = true
      template_library_installed = true
      referential_rules_enabled  = true
 }
```

### Terraform Version

```sh
Terraform v1.2.7
```

### Additional information

_No response_

apeabody:
Fixes #1369

### DIFF
--- a/modules/acm/feature.tf
+++ b/modules/acm/feature.tf
@@ -55,6 +55,7 @@ resource "google_gke_hub_feature_membership" "main" {
 
       content {
         enabled                    = true
+        referential_rules_enabled  = true
         template_library_installed = var.install_template_library
         log_denies_enabled         = var.enable_log_denies
       }


comment-bot-dev:
@apeabody

bharathkkb (approved):
Thanks @apeabody! We can move to var later if there is a request.

modules/acm/feature.tf
```diff
@@ -55,6 +55,7 @@ resource "google_gke_hub_feature_membership" "main" {
 
       content {
         enabled                    = true
+        referential_rules_enabled  = true
```
linde, line 58: wait, sorry for being late to this, but instead of hard coding this to true, shouldn't we create a variable that can be set but which defaults to `true`?

linde: at this point, i guess we can make this suggestion itself a feature request

linde: ha, i am late to the party -- this was [observed/suggested](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1373#pullrequestreview-1086219026)

apeabody: Hi Steve - I added the FR here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1393

PR #1390, opened 2022-09-07
feat: update default value to enable private ip

ericyz:


comment-bot-dev:
@ericyz

bharathkkb (changes requested):
Thanks for the PR @ericyz 
Any reason for this as it is a breaking change? We already have safer cluster modules where we enforce this as true

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/1fd7184c5be51d183908499d3cb2392b551031a4/modules/safer-cluster/main.tf#L126

ericyz:
As it aims to create private cluster, default to `enable_private_nodes = true` would make more logical sense.

ericyz:
I also acknowledge that it is a breaking change. Would you consider to incorporate it into next major version bump?

bharathkkb:
@ericyz considering we have safer cluster and this is exposed via a variable, I am leaning towards not to make this breaking change change.

ericyz:
I am for minimising the disruption, but having a public cluster with default value deviated the name of the private module. Hope this can get considered in future breaking change.

PR #1424, opened 2022-10-12
fix: incorrect `node_pools` variable type

michaellzc:
`node_pool` map is a mix of string, bool, and number. therefore, it is better to use `map(any)` to indicate its type

moreover, `list(map(string))` is breaking the use of this module in [cdktf](https://github.com/hashicorp/terraform-cdk). 

`cdktf` infers the type of module options based on the provided variable type. in the generated code, it will produce `[]map[string]string` for `node_pools` option,
but we want `[]map[string]interface{}` in order to pass correct value

e.g. `initial_node_count` is a `number` not `string`

```
│ Error: Incorrect attribute value type
│
│   on .terraform/modules/gke_self_15C14BFE/modules/beta-private-cluster/cluster.tf line 434, in resource "google_container_node_pool" "pools":
│  434:   initial_node_count = lookup(each.value, "autoscaling", true) ? lookup(
│  435:     each.value,
│  436:     "initial_node_count",
│  437:     lookup(each.value, "min_count", 1)
│  438:   ) : null
│     ├────────────────
│     │ each.value is map of string with 16 elements
│
│ Inappropriate value for attribute "initial_node_count": a number is required.
```

notes: this change is fairly safe and I did not run any integration test from this repo


comment-bot-dev:
@michaellzc

comment-bot-dev:
@michaellzc

bharathkkb (approved):
Thanks @michaellzc! This seems reasonable. We would likely move to a typed object with optional support at some point.

PR #1439, opened 2022-10-26
Make creation of istio-system namespace optional

pib:
This supports use-cases where the istio-system namespace is already created by existing terraform when adding this module. 

Terraform handles switching from a single resource to one with a `count` seamlessly, so that won't be an issue for anyone upgrading to this version.


comment-bot-dev:
@pib

bharathkkb (approved):

PR #1440, opened 2022-10-26
feat!: add service_externalips option, and add default for autoscaling location_policy

jpetrucciani:
This PR adds the ability to set `service_external_ips_config` on the underlying cluster upon creation, which otherwise can (in my experience) take up to 20-30 minutes to enable independently after the cluster settles after initial creation. I've also added a default value for autoscaling `location_policy` for now, as this is something that appears to have a default on clusters created with 1.24+, but appears to be removed after a second plan+apply with this module.

I added the changes to the templates in `./autogen/` and ran `make build` in order to generate the changes across the repo

I've been testing cluster creation with these new settings already via my fork (and it appears to be working well! shaves a good amount of time off of our new cluster creation process), but would love to see it included in the official module if possible! Please let me know if there are any changes you'd like, or things I should take into account!

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -620,8 +620,9 @@ resource "google_container_node_pool" "windows_pools" {
   dynamic "autoscaling" {
     for_each = lookup(each.value, "autoscaling", true) ? [each.value] : []
     content {
-      min_node_count = lookup(autoscaling.value, "min_count", 1)
-      max_node_count = lookup(autoscaling.value, "max_count", 100)
+      min_node_count  = lookup(autoscaling.value, "min_count", 1)
+      max_node_count  = lookup(autoscaling.value, "max_count", 100)
+      location_policy = "BALANCED"
     }
   }
 
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -202,6 +202,10 @@ resource "google_container_cluster" "primary" {
     }
   }
 
+  service_external_ips_config {
+    enabled = var.service_externalips
+  }
+
   addons_config {
     http_load_balancing {
       disabled = !var.http_load_balancing
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -96,6 +96,12 @@ variable "http_load_balancing" {
   default     = true
 }
 
+variable "service_externalips" {
+  type        = bool
+  description = "Whether external ips specified by a service will be allowed in this cluster"
+  default     = false
+}
+
 variable "datapath_provider" {
   type        = string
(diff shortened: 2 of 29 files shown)


comment-bot-dev:
@jpetrucciani

autogen/main/cluster.tf.tmpl
```diff
@@ -202,6 +202,10 @@ resource "google_container_cluster" "primary" {
     }
   }
 
+  service_external_ips_config {
```
bharathkkb, line 205: Can we make this a dynamic block if enabled?

autogen/main/variables.tf.tmpl
```diff
@@ -96,6 +96,12 @@ variable "http_load_balancing" {
   default     = true
 }
 
+variable "service_externalips" {
```
bharathkkb, line 99:
```suggestion
variable "service_external_ips" {
```

autogen/main/cluster.tf.tmpl
```diff
@@ -616,8 +620,9 @@ resource "google_container_node_pool" "windows_pools" {
   dynamic "autoscaling" {
     for_each = lookup(each.value, "autoscaling", true) ? [each.value] : []
     content {
-      min_node_count = lookup(autoscaling.value, "min_count", 1)
-      max_node_count = lookup(autoscaling.value, "max_count", 100)
+      min_node_count  = lookup(autoscaling.value, "min_count", 1)
+      max_node_count  = lookup(autoscaling.value, "max_count", 100)
+      location_policy = "BALANCED"
```
bharathkkb, line 625: This seems like an unrelated change?

bharathkkb (commented):
Thanks for the PR @jpetrucciani 
Looks like this was added in 4.35 so we need to bump the minimum version in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/84fbb9a465a0e304910519d444f2fea40869464b/autogen/main/versions.tf.tmpl#L27

https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#4350-september-6-2022

autogen/main/cluster.tf.tmpl
jpetrucciani: Unrelated to external ips yes, but this was something I've seen happen in clusters as of 1.24 - they get this default value from the provider, but this module will remove it on the next plan + apply. I can separate into second PR (and make it a var) if you'd like!

autogen/main/variables.tf.tmpl
jpetrucciani: sure, sounds good - I will make some changes locally and update!

autogen/main/cluster.tf.tmpl
jpetrucciani: Sure! let me include in the changes I'll make locally

jpetrucciani:
made a new branch to represent just the main feature #1441

comment-bot-dev:
@jpetrucciani

PR #1465, opened 2022-11-16
Add dependency on GKE hub membership in CPR creation in ASM submodule

### LINKED ISSUE #1451: ASM submodule not able to find CRD `controlplanerevisions.mesh.cloud.google.com`
### TL;DR

When using the `asm` module, the `cpr` call is not able to find the CRD `controlplanerevisions.mesh.cloud.google.com`. I think in my case this is happening because the cluster is not yet registered to the fleet when `asm` module runs the `cpr`.

### Expected behavior

I expect the module to first add the cluster to a given fleet (if registering is wanted via `enable_fleet_registration`) to make the CRD available inside the cluster and then run `cpr` module.

### Observed behavior

The module tries to run `cpr` because it is not dependent of the registration (which makes the CRD available in my understanding) and fails because the CRD is not available.

### Terraform Configuration

```hcl
module "asm" {
  source                    = "terraform-google-modules/kubernetes-engine/google//modules/asm"
  version                   = "23.2.0"
  project_id                = local.project_name
  cluster_name              = google_container_cluster.XXXXXXXXXXX.name
  cluster_location          = "europe-west1"
  multicluster_mode         = "connected"
  enable_cni                = true
  enable_mesh_feature       = true
  enable_fleet_registration = true
}
```

### Terraform Version

```sh
Terraform: 1.3.4
google: 4.42.1
Kubernetes: 2.14.0
```

### Additional information

I think that on the `depends` flag for the `cpr` you can just add the `google_gke_hub_membership.membership` as optional (if `enable_fleet_registration` is enabled) like so:

```hcl
module "cpr" {
  source  = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
  version = "~> 3.1"

  project_id       = var.project_id
  cluster_name     = var.cluster_name
  cluster_location = var.cluster_location
  internal_ip      = var.internal_ip

  kubectl_create_command  = "${path.module}/scripts/create_cpr.sh ${local.revision_name} ${local.channel} ${var.enable_cni} ${var.enable_vpc_sc}"
  kubectl_destroy_command = "${path.module}/scripts/destroy_cpr.sh ${local.revision_name}"

  module_depends_on = var.enable_fleet_registration ? [kubernetes_config_map.asm_options, google_gke_hub_membership.membership] : [kubernetes_config_map.asm_options]
}
```

Monkeyanator:
Fixes #1451

### DIFF
--- a/modules/asm/main.tf
+++ b/modules/asm/main.tf
@@ -67,5 +67,5 @@ module "cpr" {
   kubectl_create_command  = "${path.module}/scripts/create_cpr.sh ${local.revision_name} ${local.channel} ${var.enable_cni} ${var.enable_vpc_sc}"
   kubectl_destroy_command = "${path.module}/scripts/destroy_cpr.sh ${local.revision_name}"
 
-  module_depends_on = [kubernetes_config_map.asm_options]
+  module_depends_on = var.enable_fleet_registration ? [kubernetes_config_map.asm_options, google_gke_hub_membership.membership] : [kubernetes_config_map.asm_options]
 }
--- a/modules/asm/main.tf
+++ b/modules/asm/main.tf
@@ -67,5 +67,5 @@ module "cpr" {
   kubectl_create_command  = "${path.module}/scripts/create_cpr.sh ${local.revision_name} ${local.channel} ${var.enable_cni} ${var.enable_vpc_sc}"
   kubectl_destroy_command = "${path.module}/scripts/destroy_cpr.sh ${local.revision_name}"
 
-  module_depends_on = var.enable_fleet_registration ? [kubernetes_config_map.asm_options, google_gke_hub_membership.membership] : [kubernetes_config_map.asm_options]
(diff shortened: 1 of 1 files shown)


comment-bot-dev:
@Monkeyanator

Monkeyanator:
I think the conditional is not needed since if the `count` for the GKE Hub membership resource is 0, the dependency will be ignored.

comment-bot-dev:
@Monkeyanator

comment-bot-dev:
@Monkeyanator

Monkeyanator:
> Thanks for the PR Sam Naser Isn't `module.cpr` dependent on `kubernetes_config_map.asm_options` which is dependent on `google_gke_hub_membership.membership`. So wouldn't `module.cpr` wait for `google_gke_hub_membership.membership` anyway?

You're right, I just noticed this. So this fix doesn't make sense and the root cause must be different.

PR #1166, opened 2022-03-04
feat: add gcp_filestore_csi_driver_config to addons config

bharathkkb:
minor fixes from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1161

PR #1477, opened 2022-11-24
Enabled `filestore_csi_driver` option for autopilot-cluster

butterthon-dev:
This option appears to have been added in #1166.
However, for some reason it was not enabled on the autopilot cluster, so I have fixed it so that it can be enabled.
Currently it appears to be possible to use Filestore CSI in an Autopilot cluster.
see: https://cloud.google.com/kubernetes-engine/docs/tutorials/stateful-workload?hl=ja

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -224,6 +224,10 @@ resource "google_container_cluster" "primary" {
       disabled = !var.horizontal_pod_autoscaling
     }
 
+    gcp_filestore_csi_driver_config {
+      enabled = var.filestore_csi_driver
+    }
+
 
     {% if autopilot_cluster != true %}
     network_policy_config {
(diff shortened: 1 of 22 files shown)


comment-bot-dev:
@butterthon-dev

bharathkkb (changes requested):
Thanks for the PR @butterthon-dev 
It looks like the provider still does not support this. Could you file a provider issue to fix this first?

https://github.com/hashicorp/terraform-provider-google/blob/d2ea6580ef848edd94862d13ab7cddcb162df904/google/resource_container_cluster.go#L297-L304


From test:
```
       
       Error: Conflicting configuration arguments
       
         with module.example.module.gke.google_container_cluster.primary,
         on ../../../modules/beta-autopilot-private-cluster/cluster.tf line 113, in resource "google_container_cluster" "primary":
        113:     gcp_filestore_csi_driver_config {
       
       "addons_config.0.gcp_filestore_csi_driver_config": conflicts with
       enable_autopilot
       ```

Issue #1493, opened 2022-12-09
Unable to enable private nodes when creating nodes with specified pod ip range


Shelby-Jiao:
### TL;DR

We are trying to create private nodes with our specified pod ip range
So we set the variable enable_private_nodes to true.
After the change, the terraform plan is broken. It seems the provider kubernetes couldn't make a request with error
Error: error creating NodePool: googleapi: Error 400: EnablePrivateNodes must be enabled for private clusters with valid masterIpv4Cidr., badRequest

I find there's no variables for changing this in resource "google_container_node_pool" "pools"

### Expected behavior

a new node should be created as described below:
module.gke.google_container_node_pool.pools["default-node-pool"] will be created with such network config
resource "google_container_node_pool" "pools"

  network_config {
      enable_private_nodes = (known after apply)
      pod_ipv4_cidr_block  = (known after apply)
      pod_range            = "blablabla"
    }

### Observed behavior

Error: error creating NodePool: googleapi: Error 400: EnablePrivateNodes must be enabled for private clusters with valid masterIpv4Cidr., badRequest with module.gke.google_container_node_pool.pools["default-node-pool"], on .terraform/modules/gke/modules/beta-private-cluster/cluster.tf line 416, in resource "google_container_node_pool" "pools": 416: resource "google_container_node_pool" "pools" {

### Terraform Configuration

```hcl
We use the module beta-private-cluster just like the example in README.md.

# google_client_config and kubernetes provider must be explicitly specified like the following.
data "google_client_config" "default" {}

provider "kubernetes" {
  host                   = "https://${module.gke.endpoint}"    # Here we reference the endpoint of the cluster using the output of module gke
  token                  = data.google_client_config.default.access_token
  cluster_ca_certificate = base64decode(module.gke.ca_certificate)
}

module "gke" {
  source                     = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster"
  project_id                 = "<PROJECT ID>"
  name                       = "gke-test-1"
  region                     = "us-central1"
  zones                      = ["us-central1-a", "us-central1-b", "us-central1-f"]
  network                    = "vpc-01"
  subnetwork                 = "us-central1-01"
  ip_range_pods              = "us-central1-01-gke-01-pods"
  ip_range_services          = "us-central1-01-gke-01-services"
  http_load_balancing        = false
  network_policy             = false
  horizontal_pod_autoscaling = true
  filestore_csi_driver       = false
[...]

PR #1514, opened 2022-12-30
fix!: enable private nodes with specified pod ip range

splichy:
Fixes #1493 
enable_private_nodes attribute was introduced in google provider 4.45.0 and Google API now requires `enable_private_nodes` to be set on both cluster and node_pool when pod_ip_range is specified on the node_pool level.

PR #1516, opened 2023-01-03
fix: enable private nodes with specified pod ip range

splichy:
_The second try, as there is something rotten in #1514 where terraform-google-kubernetes-engine-int-trigger not passing while not receiving any email what is wrong :man_shrugging:_ 
Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1493
enable_private_nodes attribute was introduced in google provider 4.45.0 and Google API now requires enable_private_nodes to be set on both cluster and node_pool when pod_ip_range is specified on the node_pool level.


comment-bot-dev:
@splichy

Issue #1545, opened 2023-01-25
Update notification_config block in google_container_cluster to support filtering notification types


joys021:
### TL;DR

Update the notification_config block to support the option to Filter notification types. GKE Docs reference here: [Filtering notifications](https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-notifications?_ga=2.72140563.-1888638151.1620840363#filtering)

### Terraform Resources

```markdown
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#filter
https://cloud.google.com/kubernetes-engine/docs/concepts/cluster-notifications?_ga=2.72140563.-1888638151.1620840363#filtering
```

### Detailed design

```markdown
notification_config {
  pubsub {
    enabled = true
    topic = google_pubsub_topic.notifications.id
    filter {
     event_type = ["UPGRADE_AVAILABLE_EVENT", "UPGRADE_EVENT", "SECURITY_BULLETIN_EVENT"]
    }
  }
}
```

PR #1546, opened 2023-01-26
feat: add support for notification filtering

joys021:
Issue https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1545


autogen/main/variables.tf.tmpl
```diff
@@ -524,10 +524,16 @@ variable "disable_default_snat" {
   default     = false
 }
 
-variable "notification_config_topic" {
-  type = string
-  description = "The desired Pub/Sub topic to which notifications will be sent by GKE. Format is projects/{project}/topics/{topic}."
-  default = ""
+variable "notification_config" {
```
bharathkkb, line 527: nit: since this would be a breaking change due to var type change, could we add a new `notification_config_topic_types` top level var instead?

bharathkkb (commented):
Thanks for the PR @joys021

PR #1554, opened 2023-02-02
fix: nodepool autoscaling vars avail in GKE 1.24.1 result in conflicts. Change default behavior to null

Aurelian-Shuttleworth:
Changes introduced in [feat: add nodepool autoscaling vars avail in GKE 1.24.1 ](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1415) result in conflicts when using total limits.

Currently, [google_container_node_pool](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool) resource states that total and per zone limits can't be used together.

This change removes the default behaviour where min and max are provided by default if autoscaling is enabled.

PR #1555, opened 2023-02-02
fix: nodepool autoscaling vars avail in GKE 1.24.1 result in conflicts. Preserve default behavior

Aurelian-Shuttleworth:
Changes introduced in [feat: add nodepool autoscaling vars avail in GKE 1.24.1 ](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1415) result in conflicts when using total limits.

Currently, [google_container_node_pool](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool) resource states that total and per zone limits can't be used together.

To address this issue and maintain the current behaviour where min and max values for a node pool are set by default, this fix strays from standard behaviour where the value is set to null. Please advise if this is acceptable.
If this is not, I have prepared a second fix where the values are nulled. [fix: nodepool autoscaling vars avail in GKE 1.24.1 result in conflicts. Change default behavior to null](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1554)


comment-bot-dev:
@Aurelian-Shuttleworth

comment-bot-dev:
@Aurelian-Shuttleworth

Issue #1461, opened 2022-11-15
Autopilot clusters require enable_vertical_pod_autoscaling to be set to true.


Gagrio:
Please fix the readme to reflect this, as it's currently set as a not required value.

PR #1564, opened 2023-02-13
fix: Autopilot vertical pod autoscaling

andrewspiers-csiro:
Autopilot clusters require vertical pod autoscaling to be switched on, without this change clusters generated using the autopilot modules are rejected by Google's API. See
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1461


bharathkkb (approved):
Thanks for the PR @andrewspiers-csiro

Issue #1065, opened 2021-11-16
Using exisiting gcp sa and k8s sa the null object command to anontate fails.


xophere:
### TL;DR

When attempting to use this module with and existing SA's the module's null object command for kubectl anontate fails the command succeeds on the command line.  Claim annotate command not found.

### Expected behavior

k8s SA is annotated with gcp SA.

### Observed behavior

t@prj-iac-cicd-87b6.iam.gserviceaccount.com/roles/iam.workloadIdentityUser/serviceAccount:PROJECT.svc.id.goog[iac-cicd/iac-cicd-gitlab-runner]]
╷
│ Error: local-exec provisioner error
│ 
│   with module.cicd_bootstrap.module.my-app-workload-identity.module.annotate-sa.module.gcloud_kubectl.null_resource.run_command[0],
│   on .terraform/modules/cicd_bootstrap.my-app-workload-identity.annotate-sa/main.tf line 231, in resource "null_resource" "run_command":
│  231:   provisioner "local-exec" {
│ 
│ Error running command 'PATH=/google-cloud-sdk/bin:$PATH
│ .terraform/modules/cicd_bootstrap.my-app-workload-identity.annotate-sa/modules/kubectl-wrapper/scripts/kubectl_wrapper.sh   PROJECT false false kubectl annotate
│ --overwrite sa -n iac-cicd iac-cicd-gitlab-runner iam.gke.io/gcp-service-account=org-cicd-account@PROJECT.iam.gserviceaccount.com
│ ': exit status 1. Output: + '[' 11 -lt 5 ']'
│ + CLUSTER_NAME=CLUSTERNAME
│ + LOCATION=false
│ + PROJECT_ID=false
│ + INTERNAL=kubectl
│ + USE_EXISTING_CONTEXT=annotate
│ + ENABLE_IMPERSONATE_SERVICE_ACCOUNT=--overwrite
│ + IMPERSONATE_SERVICE_ACCOUNT=sa
│ + shift 5
│ + annotate
│ .terraform/modules/cicd_bootstrap.my-app-workload-identity.annotate-sa/modules/kubectl-wrapper/scripts/kubectl_wrapper.sh: line 34: annotate: command not found
│ + RANDOM_ID=24686_25245
│ + export TMPDIR=/tmp/kubectl_wrapper_24686_25245
│ + TMPDIR=/tmp/kubectl_wrapper_24686_25245
│ + trap cleanup EXIT
[...]

PR #1143, opened 2022-02-03
Update README to prevent KSA misconfiguration

MoskitoHero:
Update the documentation of the workload identity module to address issue #1065

PR #1567, opened 2023-02-18
feat: add validation checks when using existing KSA

loremaps:
See #1065

This PR adds checks for `cluster_name ` and `location` variables, when using an existing KSA.

The error message is not great:

```
╷
│ Error: Invalid function argument
│
│   on ../../../../../personal/terraform-google-kubernetes-engine/modules/workload-identity/main.tf line 33, in locals:
│   33:   validate_cluster = (var.use_existing_k8s_sa && var.cluster_name == "") ? tobool("Variable 'cluster_name' is required when using existing KSA") : true
│     ├────────────────
│     │ while calling tobool(v)
│
│ Invalid value for "v" parameter: cannot convert "Variable 'cluster_name' is required when using existing KSA" to bool; only the strings "true" or "false" are allowed.
╵
```

But I was bitten by this issue two times already. I believe that this does give some guidance. But if you feel the error is too misleading, please feel free to close this PR 😅 

This was the simplest solution I could find until hashicorp/terraform#25609 or hashicorp/terraform#31122 are addressed. Taken from [here](https://github.com/hashicorp/terraform/issues/25609#issuecomment-1057614400) 

Thank you for your time reviewing!


bharathkkb (commented):
Thanks for the PR @loremaps 
I think the `tobool` workaround is a bit confusing. Could we update the documentation instead?

loremaps:
Hey @bharathkkb thanks for reviewing.

Indeed it is and please feel free to close this PR. 

The documentation was updated in #1143 AFAIK to help us avoid this misconfiguration. 

But if you miss the docs, please compare the error message shown in #1065 and that of the `tobool` workaround. And then you have to manually remove from state to recover (see https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1065#issuecomment-989003568)

Once again, thanks for your time and feel free to close this, np!

bharathkkb:
Thanks @loremaps

PR #1574, opened 2023-02-27
feat: Workload Identity module, to bind roles in various projects for the service account created

SudharsaneSivamany:
It will help users to bind the created service account account with roles in multiple projects.


modules/workload-identity/main.tf
```diff
@@ -85,7 +85,7 @@ resource "google_service_account_iam_member" "main" {
 resource "google_project_iam_member" "workload_identity_sa_bindings" {
   for_each = toset(var.roles)
 
-  project = var.project_id
-  role    = each.value
+  project = element(split("=>", each.value), 0)
+  role    = element(split("=>", each.value), 1)
```
apeabody, line 89: I would recommend using a `map(list(string))` rather than string parsing.

modules/workload-identity/variables.tf
```diff
@@ -85,7 +85,7 @@ variable "automount_service_account_token" {
 }
 
 variable "roles" {
-  description = "A list of roles to be added to the created service account"
+  description = "A list of roles to be added to the created service account for specific projects"
```
apeabody, line 88: Currently this would be a breaking change, if possible can we make this backward compatible.  

A quick thought might be add a new (optional) `map(list(string))` of additional projects=>[roles], and leave the existing `roles` for just `var.project_id`?

apeabody (changes requested):
Thanks for the contribution @SudharsaneSivamany!  Please see the comments as ideally we can avoid a breaking change.

modules/workload-identity/main.tf
SudharsaneSivamany: Agreed with your point.

modules/workload-identity/variables.tf
SudharsaneSivamany: Changes are made as mentioned. Pls review

SudharsaneSivamany (commented):
Changes are made. Pls review

modules/workload-identity/main.tf
```diff
@@ -89,3 +89,11 @@ resource "google_project_iam_member" "workload_identity_sa_bindings" {
   role    = each.value
   member  = local.gcp_sa_fqn
 }
+
+resource "google_project_iam_member" "workload_identity_sa_bindings_additional_projects" {
+  for_each = toset(distinct(flatten([for project, roles in var.additional_projects : [for role in roles : join("=>", [project, role])]])))
```
apeabody, line 94: Thanks @SudharsaneSivamany 

If feasible, I suggest we try to avoid the join/split paradigm.

If you create a set of objects containing project and role, you should be able to pass those directly and avoid the join/split.  I have not tested/debugged this exact code, but something like this:

```suggestion
  for_each = flatten([for project, roles in var.additional_projects : [for role in roles : {project = project, role = role}]])
```

project = each.value.project
role = each.value.role

SudharsaneSivamany: Sure @apeabody . Changes are made as suggested. Please review.

apeabody: Thanks! Removes the risk of an unexpected value causing an issue with the split.  I've triggered the CI.

apeabody (approved):
Thanks @SudharsaneSivamany!

PR #1576, opened 2023-03-02
fix(acm)!: prevent conflicts in IAM binding

evenh:
When using both config sync and policy controller the existing binding resources conflicts with each other and causes a constant diff.

This PR will cause a diff in existing configurations, but when applying it should actually apply the the binding to both members resulting in the intended outcome and no further diffs.

### DIFF
--- a/modules/acm/creds.tf
+++ b/modules/acm/creds.tf
@@ -17,6 +17,11 @@
 locals {
   # GCP service account ids must be <= 30 chars matching regex ^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$
   service_account_name = trimsuffix(substr(var.metrics_gcp_sa_name, 0, 30), "-")
+
+  iam_ksa_binding_members = var.create_metrics_gcp_sa ? [
+    var.enable_config_sync ? "config-management-monitoring/default" : null,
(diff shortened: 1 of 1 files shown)


apeabody:
Thanks for the contribution @evenh, relevant output from LINT:

```
Running terraform fmt
modules/acm/creds.tf
--- old/modules/acm/creds.tf
+++ new/modules/acm/creds.tf
@@ -44,7 +44,7 @@
   role               = "roles/iam.workloadIdentityUser"
 
   members = [
-    for ksa in local.iam_ksa_binding_members :"serviceAccount:${var.project_id}.svc.id.goog[${ksa}]"
+    for ksa in local.iam_ksa_binding_members : "serviceAccount:${var.project_id}.svc.id.goog[${ksa}]"
   ]
 
   depends_on = [google_gke_hub_feature_membership.main]
```

apeabody:
My suspicion is the failing INT is unrelated.

evenh:
I don't have insight into the integration tests but I hope it is unrelated

apeabody:
re-triggering CI

apeabody (approved):
Thanks for the contribution @evenh!

Issue #1566, opened 2023-02-17
Can not use the module when organization policy constraints/gcp.restrictNonCmekServices is active


felix-lecuyer:
### TL;DR

When the organization policy constraints/gcp.restrictNonCmekServices apply on services compute.googleapis.com and container.googleapis.com, there is no configuration that enables the deployment of a kubernetes cluster using the module. 

### Expected behavior

Given:
- a project where the organization policy constraints/gcp.restrictNonCmekServices apply on services compute.googleapis.com and container.googleapis.com
-  a terraform configuration using the kubernetes engine module and providing a boot_disk_kms_key pointing to a CMEK for nodepools and autoscale configuration

the configuration should result in the deployment of a kubernetes cluster, which nodes have disk encrypted using the provided CMEK. 

### Observed behavior

When running terraform apply, the following error message is returned:
Error: googleapi: Error 400: Failed precondition: Constraint `constraints/gcp.restrictNonCmekServices` violated for `projects/{project_id}` attempting to create a resource without specifying a KMS CryptoKey., failedPrecondition 

### Terraform Configuration

```hcl
module "gcp-network" {
  source       = "terraform-google-modules/network/google"
  version      = "6.0.1"
  project_id   = var.project_id
  network_name = var.network

  subnets = [
    {
      subnet_name           = var.subnetwork
      subnet_ip             = "10.128.0.0/20"
      subnet_region         = var.region
      subnet_private_access = true
    },
  ]

  secondary_ranges = {
    (var.subnetwork) = [
      {
        range_name    = var.ip_range_pods_name
        ip_cidr_range = "10.129.0.0/20"
      },
      {
        range_name    = var.ip_range_services_name
        ip_cidr_range = "10.130.0.0/20"
      },
    ]
  }
}

module "gke" {
  source                   = "terraform-google-modules/kubernetes-engine/google"
  version                  = "25.0.0"
  project_id               = var.project_id
  database_encryption      = [{ "key_name" : var.key_name, "state" : "ENCRYPTED" }]
  name                     = "main"
  region                   = var.region
  network                  = module.gcp-network.network_name
  subnetwork               = module.gcp-network.subnets_names[0]
  ip_range_pods            = var.ip_range_pods_name
  ip_range_services        = var.ip_range_services_name
  remove_default_node_pool = true
  create_service_account   = false
  service_account          = var.service_account
  node_metadata            = "metadata"
  kubernetes_version       = "latest"
[...]

PR #1577, opened 2023-03-02
fix: add variable for customer managed encryption key for cluster autoscaling defaults

felix-lecuyer:
Fix issue #1566 . 

Add a new input variable for beta modules to input a customer managed encryption key used in cluster_autoscaling defaults. 

Add some examples on how to configure an autopilot and a node pool cluster with customer managed encryption key. 

Purpose of this change: be able to use the module in a project where the organization policy constraints/gcp.restrictNonCmekServices is active (e.g. in a sovereign folder).


autogen/main/variables.tf.tmpl
```diff
@@ -753,3 +753,11 @@ variable "enable_identity_service" {
 }
   {% endif %}
 {% endif %}
+
+{% if beta_cluster %}
```
bharathkkb, line 757: Any reason for restricting to beta? I see it available in the GA provider

https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#4310-aug-1-2022

bharathkkb (commented):
Thanks for the PR @felix-lecuyer 
Could we showcase this feature in an existing example rather than adding a new example? We also try to keep the examples self contained so it would be good to spin up a sample key in the example itself rather than getting it as an input.

autogen/main/variables.tf.tmpl
felix-lecuyer: Hi Barath, 
Thanks for the feedback. No particular reason so I removed the restriction to beta in my last commit..

felix-lecuyer:
Hi @bharathkkb, 
I managed to take into account your comments - sorry for the delay.
Looking forward to a new review. 

Many thanks.
Felix

dangarthwaite:
Hi @bharathkkb, we really need this to comply with gcp.restrictNonCmekServices.  How can we get this merged?

PR #1594, opened 2023-03-23
fix: Workload Identity Module - Add required parameters to example snippet

slashr:
Update example code snippet demonstrating use of existing Kubernetes SA to include the required parameters `cluster` and `location`. #1065

### DIFF
--- a/modules/workload-identity/README.md
+++ b/modules/workload-identity/README.md
@@ -83,6 +83,8 @@ resource "kubernetes_service_account" "preexisting" {
 module "my-app-workload-identity" {
   source              = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
   use_existing_k8s_sa = true
+  cluster             = "my-k8s-cluster-name"
+  location            = "my-k8s-cluster-location"
   name                = kubernetes_service_account.preexisting.metadata[0].name
   namespace           = kubernetes_service_account.preexisting.metadata[0].namespace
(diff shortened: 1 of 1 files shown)


apeabody (changes requested):
Thanks for the contribution @slashr!

We can add something to the example, however it appears these are only required if using existing KSA:
- https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/workload-identity/variables.tf#L39
- https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/workload-identity/main.tf#L67

Let's clarify in the example, as these values are not used if not using an existing KSA.

slashr:
Hi @apeabody 

Thanks for your feedback. The variables are already added to the specific code snippet that's under the "Using an existing Kubernetes Service Account" section.

If that's not explicit enough, do you suggest adding an additional comment next to the variables reemphasising that those variables are only meant for existing k8s service account usage?

apeabody:
> Hi @apeabody
> 
> Thanks for your feedback. The variables are already added to the specific code snippet that's under the "Using an existing Kubernetes Service Account" section.
> 
> If that's not explicit enough, do you suggest adding an additional comment next to the variables reemphasising that those variables are only meant for existing k8s service account usage?

Thanks @slashr, yes, I see now they are under just the "Using an existing Kubernetes Service Account" section.

Issue #1582, opened 2023-03-07
workload-identity compatibility


DerEinePete:
Hi,

In the workload-identity module is stated "Note: This module currently supports Kubernetes <= 1.23." Is this still the case?

https://cloud.google.com/kubernetes-engine/docs/release-schedule

The release schedule of the Google Cloud states, that it is already upgrading kubernetes to 1.24.
But I haven't experienced any issue.

nicolaihald:
@DerEinePete : 
Just found this ticket and thought i'd post a comment. I'm not a maintainer, but it's likely related to https://github.com/hashicorp/terraform-provider-kubernetes/issues/1724  

`v1.24.0` includes a [breaking change](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#no-really-you-must-read-this-before-you-upgrade) as the default secret that is/was generated when creating a service account no longer will be generated. So, depending on how you manage your kubernetes service accounts, you may run into a warning similar to the one below. The fix is simple though. If i have time I may create a PR to fix it. That being said, we don't use this module so it's way down on the priority list. 


> _Warning: "default_secret_name" is no longer applicable for Kubernetes v1.24.0 and above_



 [v1.24 Release Notes - Breaking Change](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#no-really-you-must-read-this-before-you-upgrade):
> _"The `LegacyServiceAccountTokenNoAutoGeneration`eature gate is beta, and enabled by default. When enabled, Secret API objects containing service account tokens are no longer auto-generated for every ServiceAccount. Use the [TokenRequest](https://kubernetes.io/docs/reference/kubernetes-api/authentication-resources/token-request-v1/) API to acquire service account tokens, or if a non-expiring token is required, create a Secret API object for the token controller to populate with a service account token by following this [guide](https://kubernetes.io/docs/concepts/configuration/secret/#service-account-token-secrets). (https://github.com/kubernetes/kubernetes/pull/108309, [@zshihang](https://github.com/zshihang))"_


You can see an example of how to fix it here: 
- [Was fixed in https://github.com/oracle-quickstart/oci-micronaut/pull/31](https://github.com/oracle-quickstart/oci-micronaut/pull/31/files)

henrybell:
[...]

Issue #1313, opened 2022-07-01
CI failure waiting for k8s objects 


bharathkkb:
### TL;DR

CI seems to be failing with `Waiting for default secret of "default/name" to appear`.  Maybe related to https://github.com/hashicorp/terraform-provider-kubernetes/issues/1724, test clusters are 1.24.1-x.

### Terraform Configuration

```hcl
N/A
```

### Terraform Version

```sh
N/A
```

PR #1315, opened 2022-07-08
chore: pin workload_identity example to k8s 1.23

apeabody:
* A temporary fix for #1313.  With k8s 1.24 (no channel), the k8s provider resource "kubernetes_service_account" doesn't currently work as expected.  This temporarily sets the workload-identity example/module to k8s 1.23: https://cloud.google.com/kubernetes-engine/docs/release-schedule

PR #1595, opened 2023-03-25
Fix: Remove 1.23 restriction on workload identity module

liggitt:
The issue with hashicorp/kubernetes hanging creating service accounts was resolved in https://github.com/hashicorp/terraform-provider-kubernetes/pull/1792 in https://github.com/hashicorp/terraform-provider-kubernetes/releases/tag/v2.13.0

This reverts the temporary limitation of the workload identity module to 1.23 added in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1315 to resolve https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1313, and updates the minimum required version of hashicorp/kubernetes to 2.13.0 for the workload-identity module

Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1582

cc @bharathkkb @apeabody

### DIFF
--- a/examples/workload_identity/main.tf
+++ b/examples/workload_identity/main.tf
@@ -38,8 +38,6 @@ module "gke" {
   remove_default_node_pool = true
   service_account          = "create"
   node_metadata            = "GKE_METADATA"
-  # https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1313
-  kubernetes_version = "1.23"
   node_pools = [
     {
       name         = "wi-pool"
--- a/modules/workload-identity/README.md
+++ b/modules/workload-identity/README.md
@@ -14,8 +14,6 @@ The `terraform-google-workload-identity` can create service accounts for you,
 or you can use existing accounts; this applies for both the Google and
 Kubernetes accounts.
 
-Note: This module currently supports Kubernetes <= 1.23.
-
 ### Creating a Workload Identity
 
 ```hcl
--- a/modules/workload-identity/versions.tf
+++ b/modules/workload-identity/versions.tf
@@ -25,7 +25,7 @@ terraform {
     }
     kubernetes = {
       source  = "hashicorp/kubernetes"
-      version = "~> 2.0"
+      version = "~> 2.13"
     }
   }
 


liggitt:
looks like the CI failure is pre-existing (same failure in https://pantheon.corp.google.com/cloud-build/builds/0065125b-5404-4ce8-8d16-c5eee9f04a41?project=cloud-foundation-cicd&e=-13802955&jsmode=o&mods=logs_tg_prod from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1584):

```
Finished Step #94 - "init simple-autopilot-private-non-default-sa"
ERROR
ERROR: build step 94 "gcr.io/cloud-foundation-cicd/cft/developer-tools:1.8" failed: step exited with non-zero status: 1
```

apeabody:
Thanks @liggitt!

@bharathkkb, I opened https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1597

liggitt:
looks like the rerun passed, and the `workload-identity-local` steps succeeded

apeabody:
> looks like the rerun passed, and the `workload-identity-local` steps succeeded

Yes, looks like some CI flakiness as both this and my test PR just passed.

apeabody (approved):
Thanks @liggitt! LGTM!

liggitt:
thanks! any chance of pulling this into a 25.1.0 patch release ahead of 26.0.0?

apeabody:
> thanks! any chance of pulling this into a 25.1.0 patch release ahead of 26.0.0?

Let me discuss with @bharathkkb, but I think we might be able to make the next release `v25.1.0`.

bharathkkb:
@liggitt @apeabody IIUC The `1.23` k8s version  was just hardcoded in the example and not in the module. Since the previously module supported `> 2.0, < 3.0` anyway, users could technically use 2.13+ providers with the fix. 

We have now also bumped the min constraint to `> 2.13, < 3.0` which would need a breaking release incase users were explicitly pinned between `> 2.0, < 2.13`.

liggitt:
Interesting, thanks for that clarification. It's surprising a minor version dependency bump for a downstream bugfix requires a major version change here, but I'm glad existing consumers can use against 1.24+ successfully without any changes here

apeabody:
> @liggitt @apeabody IIUC The `1.23` k8s version was just hardcoded in the example and not in the module. Since the previously module supported `> 2.0, < 3.0` anyway, users could technically use 2.13+ providers with the fix.
> 
> We have now also bumped the min constraint to `> 2.13, < 3.0` which would need a breaking release incase users were explicitly pinned between `> 2.0, < 2.13`.

Thanks @bharathkkb!  That was going to be my question with regards to it being a non-Google provider.

apeabody:
> Interesting, thanks for that clarification. It's surprising a minor version dependency bump for a downstream bugfix requires a major version change here, but I'm glad existing consumers can use against 1.24+ successfully without any changes here

Hi @liggitt, yes good call by @bharathkkb that it was the module README.md and example used for CI which were pinned to `kubernetes_version = "1.23"`, nothing in the `v25.0.0` `terraform-google-kubernetes-engine` module actually limited the kubernetes version.  As long as a user doesn't have it pinned elsewhere, terraform init -upgrade` should update the `kubernetes` provider to the most recent 2.x release and everything should work as intended.

PR #1598, opened 2023-03-27
fix: fixes for tflint and dev-tools 1.10

apeabody:


autogen/safer-cluster/variables.tf.tmpl
```diff
@@ -26,12 +26,6 @@ variable "name" {
   description = "The name of the cluster"
 }
 
-variable "description" {
```
bharathkkb, line 29: nit: It would be good to keep this and plumb it through

bharathkkb (approved):

autogen/safer-cluster/variables.tf.tmpl
apeabody: Done - That was probably less work than removing it in the first place.

PR #1603, opened 2023-04-06
feat: add enable_private_nodes options to node_pool network_config

verdel:
This PR adds the ability to set the enable_private_notes option in network_config for a pool of public cluster nodes.

We already had a cluster in which the nodes had public addresses. We wanted to change this property of the node without recreating the cluster. We would have to recreate the cluster, since the private_cluster_config property is immutable and we cannot use the module for a private cluster.


jcortejoso:
HI @verdel! Any reason for closing this PR?
We are in a similar need (having private nodes in public cluster), and it would be nice to have support from the module, considering it is an option supported by GKE.
Thanks!

PR #1610, opened 2023-04-17
fix: use provided service_account_name if available

### LINKED ISSUE #1609: [BUG] Service account created doesn't have the correct name
### TL;DR

When specifying a service_account_name, with create_service_account = True the created service account doesn't have the correct name, and the node pool is waiting for the service_account_name that you provided. This results in the inability to set a service_account_name.

### Expected behavior

When specifying a service_account_name, the created service_account has the correct name, and is the one used by the node pool

### Observed behavior

Node pool is using correct service_account_name, but the service account created is not the correct one.

### Terraform Configuration

```hcl
module "gke" {
  source  = "terraform-google-modules/kubernetes-engine/google//modules/private-cluster"
  version = "25.0.0"
  depends_on = [
    google_project_iam_member.roles
  ]
  name                       = var.cluster_name
  project_id                 = var.project_id
  network                    = var.cluster_network
  subnetwork                 = var.cluster_subnetwork
  ip_range_pods              = var.cluster_ip_range_pods
  ip_range_services          = var.cluster_ip_range_services
  network_project_id         = var.cluster_network_project_id
  region                     = var.default_region
  remove_default_node_pool   = true
  enable_private_nodes       = true
  enable_private_endpoint    = true
  master_ipv4_cidr_block     = "172.16.255.240/28"
  master_authorized_networks = [{ cidr_block = "172.16.255.240/28", display_name = "placeholder to not have a public IP" }]
  service_account            = "gke-${var.cluster_name}@${var.project_id}.iam.gserviceaccount.com"
}
```

### Terraform Version

```sh
Terraform v1.4.0
```

### Additional information

Service account name defined here is not using the correct name : https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/sa.tf#L42

griseau:
Fixes #1609

### DIFF
--- a/autogen/main/sa.tf.tmpl
+++ b/autogen/main/sa.tf.tmpl
@@ -17,14 +17,9 @@
 {{ autogeneration_note }}
 
 locals {
-  service_account_list = compact(
-    concat(
-      google_service_account.cluster_service_account.*.email,
-      ["dummy"],
-    ),
-  )
+  service_account_default_name = "tf-gke-${substr(var.name, 0, min(15, length(var.name)))}-${random_string.cluster_service_account_suffix.result}"
   // if user set var.service_account it will be used even if var.create_service_account==true, so service account will be created but not used
-  service_account = (var.service_account == "" || var.service_account == "create") && var.create_service_account ? local.service_account_list[0] : var.service_account
+  service_account = (var.service_account == "" || var.service_account == "create") && var.create_service_account ? local.service_account_default_name : var.service_account
 
   registry_projects_list = length(var.registry_project_ids) == 0 ? [var.project_id] : var.registry_project_ids
 }
@@ -39,7 +34,7 @@ resource "random_string" "cluster_service_account_suffix" {
 resource "google_service_account" "cluster_service_account" {
   count        = var.create_service_account ? 1 : 0
   project      = var.project_id
-  account_id   = "tf-gke-${substr(var.name, 0, min(15, length(var.name)))}-${random_string.cluster_service_account_suffix.result}"
+  account_id   = local.service_account
   display_name = "Terraform-managed service account for cluster ${var.name}"
 }
 


griseau:
Hey @apeabody, 

I see that terraform-google-kubernetes-engine-int-trigger action failed but I don't have the possibility to see the reason it's failing, so can't fix it :/

apeabody:
> Hey @apeabody,
> 
> I see that terraform-google-kubernetes-engine-int-trigger action failed but I don't have the possibility to see the reason it's failing, so can't fix it :/

Thanks @griseau - It appeared to be a flaky test run, no issues on the second run.

autogen/main/sa.tf.tmpl
```diff
@@ -39,7 +42,7 @@ resource "random_string" "cluster_service_account_suffix" {
 resource "google_service_account" "cluster_service_account" {
   count        = var.create_service_account ? 1 : 0
   project      = var.project_id
-  account_id   = "tf-gke-${substr(var.name, 0, min(15, length(var.name)))}-${random_string.cluster_service_account_suffix.result}"
+  account_id   = local.service_account_name
```
apeabody, line 45: Do we need `local.service_account_name`, or would this be simpler?
```suggestion
  account_id   = var.service_account_name == "" ? local.service_account_default_name : var.service_account_name
```

apeabody (changes requested):
Thanks for the contribution @griseau!

Can you please run `make build` to build the terraform and docs.

autogen/main/sa.tf.tmpl
griseau: Indeed I'm removing it ! 

I wanted to use var.service_account at the beginning but this variable contains the full email of the service account, whereas here we just want the name. Maybe we should make the doc clearer here ?

apeabody: Sure, a code comment or something else might be nice.  I'll re-trigger the CI.

/gcbrun

griseau: I've just pushed an update of the doc, you might need to retrigger the CI, sorry

apeabody (approved):
Thanks for the contribution @griseau!

apeabody:
> Hey @apeabody, Anything else required here ? Would we be able to have it in the next release ? :)

Yup, will be included in the next release.

PR #1614, opened 2023-04-23
chore: Attach KMS Key in Safer IAP GKE cluster

avinashkumar1289:
This request have changes
1) KMS.tf file : Generating a new KMS Key in the GKE Project
2) Attaching the key to GKE cluster for application encryption


Jberlinsky:
To be clear -- this PR does not create new support for KMS, it just adds an example of providing a KMS key to the existing module. Is that the intent?

avinashkumar1289:
Yes so it provide an example of creating a KMS Key and attach the key to GKE that's the intent.

examples/safer_cluster_iap_bastion/kms.tf
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+data "google_project" "project" {}
+
+locals {
+  gke_sa = "serviceAccount:service-${data.google_project.project.number}@container-engine-robot.iam.gserviceaccount.com"
+}
+
```
bharathkkb, line 22: We can generate this using [service identities](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/project_services/variables.tf#L34) from the project services module in apis.tf and use the `enabled_api_identities` output to retrieve the sa email.

examples/safer_cluster_iap_bastion/variables.tf
```diff
@@ -69,3 +69,15 @@ variable "bastion_members" {
...
+  description = "Keyring name."
+  type        = string
+  default     = "gke-keyring"
+}
+
+variable "keys" {
+  description = "Key names."
+  type        = list(string)
+  default     = ["gke-key"]
+}
```
bharathkkb, line 83: We are transitioning to inline these variables so you can directly provide them in kms.tf

bharathkkb (commented):
Thanks for the PR @avinashkumar1289

examples/safer_cluster_iap_bastion/kms.tf
avinashkumar1289: Hi @bharathkkb Thanks for the response. (activate_api_identities)[https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/project_services/variables.tf#L34] required the api and the role. I can pass the api but what should be the role I need to pass ? Because this service account needs the KMS encrypter/Decrypter role which is taken care at the KMS Module
Something like. this

```
 encrypters = [
   "serviceAccount:${module.enabled_google_apis.enabled_api_identities["container.googleapis.com"]}",
  ]
  decrypters = [
  "serviceAccount:${module.enabled_google_apis.enabled_api_identities["container.googleapis.com"]}",
  ]
  
  activate_api_identities     = [{
    api = "container.googleapis.com",
    roles = [????],
  }]
```

examples/safer_cluster_iap_bastion/variables.tf
avinashkumar1289: @bharathkkb updated the comment

examples/safer_cluster_iap_bastion/kms.tf
avinashkumar1289: Updating the role at the time of api_identities and removed thee role from KMS module

bharathkkb (approved):

PR #1615, opened 2023-04-23
chore: Use `google_service_account#member` attribute

j8r:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account#member can be used, which is the same as `serviceAccount:{email}`.


bharathkkb (approved):
Thanks for the PR @j8r

Issue #1586, opened 2023-03-09
Add support for GKE Protect/Security Posture


jawnsy:
### TL;DR

Add support for GKE Security Posture, formerly known as GKE Protect

### Terraform Resources

>The protect_config block supports:
>
>workload_config - (Optional, Beta) WorkloadConfig defines which actions are enabled for a cluster's workload configurations. Structure is documented below
>
>workload_vulnerability_mode - (Optional, Beta) Sets which mode to use for Protect workload vulnerability scanning feature. Accepted values are WORKLOAD_VULNERABILITY_MODE_UNSPECIFIED, DISABLED, BASIC.
>
>The protect_config.workload_config block supports:
>
>auditMode - (Optional, Beta) WorkloadConfig defines the flags to enable or disable the workload configurations for the cluster. Accepted values are MODE_UNSPECIFIED, DISABLED, BASIC.

Documentation link: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#protect_config

### Additional information

Requires TPG 4.40.0 and was introduced inhttps://github.com/hashicorp/terraform-provider-google/issues/12778

NissesSenap:
Work ongoing

PR #1617, opened 2023-04-25
feat!: Add protect_config beta feature

NissesSenap:
fix: #1586

### DIFF
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -530,6 +530,17 @@ variable "enable_confidential_nodes" {
   description = "An optional flag to enable confidential node config."
   default = false
 }
+variable "workload_vulnerability_mode" {
+  description = "(beta) Vulnerability mode."
+  type        = string
+  default     = ""
+}
+
+variable "workload_config_audit_mode" {
+  description = "(beta) Worload config audit mode."
+  type        = string
+  default     = "DISABLED"
+}
 {% endif %}
 
 variable "disable_default_snat" {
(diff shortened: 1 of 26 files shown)


autogen/main/variables.tf.tmpl
```diff
@@ -766,5 +766,23 @@ variable "enable_identity_service" {
   description = "Enable the Identity Service component, which allows customers to use external identity providers with the K8S API."
   default     = false
 }
+
+variable "protect_config" {
```
ericyz, line 770: Suggest to simplify the variable for this configuration. Do you reckon we could `workload_config_audit_mode` and 
`workload_vulnerability_mode` variables to toggle the feature?

ericyz (changes requested):
minor suggestion for review

autogen/main/variables.tf.tmpl
NissesSenap: My main reasoning behind use an object is that I don't know if GCP will add any other API that we might want to add in the future, and then I thought it would be easier to do that change with an object without making any breaking change.
But a string is definitely easier. I have updated the PR to match your suggestion.

ericyz:
LGTM and will approve after the test

ericyz:
LGTM and will approve after the int test

bharathkkb:
@NissesSenap looks like there are some validate failures

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/actions/runs/4943651436/jobs/8838343712

NissesSenap:
I have fixed that now @bharathkkb 
I tried running `make docker_test_integration` but there are a number of things that didn't pass. But I don't think they have anything to do with my changes.

So hopefully everything should pass now in the CI

NissesSenap:
ping @ericyz @bharathkkb

ericyz (approved):

PR #1618, opened 2023-04-26
fix!: set release_channel and auto_upgrade, drop meshtelemetry

apeabody:
* Set `release_channel` default to `REGULAR` for GKE behavior, `REGULAR` is already default for safe_cluster
* Set `auto_upgrade` to force true when a `release_channel` other than `UNSPECIFIED` is used
* CI `prepare` stage is silently failing (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1619) as `meshtelemetry.googleapis.com` is deprecated: `SERVICE_CONFIG_NOT_FOUND_OR_PERMISSION_DENIED`

Fixes: #1620


bharathkkb (approved):
Thanks for adding the upgrade guide too!

PR #1621, opened 2023-04-28
chore: remove deprecated hashicorp/template from module examples

jleloup:
Using the workload-identity module from an Apple M2 laptop triggers an error related to hashicorp/template being deprecated and not providing any builds for `darwin_arm64`.

Since the only references I can see for this provider are in some examples: I'm updating them to use the supported syntax instead.


apeabody (approved):
Thanks for the contribution @jleloup!

apeabody:
Marking as `chore` as this is under examples.

PR #1637, opened 2023-05-16
fix: avoid TPG 4.65.0 and 4.65.1

apeabody:
https://github.com/hashicorp/terraform-provider-google/issues/14602#issuecomment-1549689089

### DIFF
--- a/autogen/main/versions.tf.tmpl
+++ b/autogen/main/versions.tf.tmpl
@@ -24,11 +24,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 4.51.0, < 5.0"
+      version = ">= 4.51.0, < 4.65.0, !=4.65.0, !=4.65.1"
     }
     google-beta = {
       source  = "hashicorp/google-beta"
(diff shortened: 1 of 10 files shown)


bharathkkb (approved):

wyardley:
q: maybe I'm parsing this wrong, but doesn't `<4.65.0` already prevent `4.65.0` and `4.65.1` in `version = ">= 4.51.0, < 4.65.0, !=4.65.0, !=4.65.1"`

Is this to say that `4.65.2`, for example, is invalid? I'm getting an error with it.

apeabody:
Hi @wyardley - Thanks for reaching out!

We needed to avoid usage of TPG `4.65.0` and `4.65.1`, please use one of the following which includes #1640 and will enable `4.65.2`, etc:
* https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/releases/tag/v26.1.1
* https://github.com/terraform-google-modules/terraform-google-kubernetes-engine

wyardley:
@apeabody it's in trying to update to v26.1.1 exactly where I get the error, because you're requiring the version to be < 4.65

Issue #1527, opened 2023-01-11
Firewall rule name conflicts if `var.name` is too long since it'll be truncated.


FuyuanChu:
### TL;DR

Firewall rule name conflicts if `var.name` is too long since it'll be truncated.

`"gke-${substr(var.name, 0, min(25, length(var.name)))}-webhooks"`

### Expected behavior

Generate unique firewall rule names regardless of the length of `var.name`

### Observed behavior

If I have two clusters with a long name variable such as
`var.name = a-cluster-running-in-region-hk`
`var.name = a-cluster-running-in-region-sg`

The firewall rule name generated for webhook will be the same since `var.name` will be truncated.

### Terraform Configuration

Nothing special. 
Just create two clusters with long names that share the same 25 characters.
And set `add_master_webhook_firewall_rules` to  `true`

```hcl

# google_client_config and kubernetes provider must be explicitly specified like the following.
data "google_client_config" "default" {}

provider "kubernetes" {
  host                   = "https://${module.gke.endpoint}"
  token                  = data.google_client_config.default.access_token
  cluster_ca_certificate = base64decode(module.gke.ca_certificate)
}

module "gke" {
  source                     = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster"
  project_id                 = "<PROJECT ID>"
  name                       = "a-cluster-running-in-region-sg"
  add_master_webhook_firewall_rules    = true
  region                     = "us-central1"
  zones                      = ["us-central1-a", "us-central1-b", "us-central1-f"]
  network                    = "vpc-01"
  subnetwork                 = "us-central1-01"
  ip_range_pods              = "us-central1-01-gke-01-pods"
  ip_range_services          = "us-central1-01-gke-01-services"
  http_load_balancing        = false
  network_policy             = false
  horizontal_pod_autoscaling = true
  filestore_csi_driver       = false
  enable_private_endpoint    = true
  enable_private_nodes       = true
  master_ipv4_cidr_block     = "10.0.0.0/28"
  istio                      = true
  cloudrun                   = true
  dns_cache                  = false

  node_pools = [
    {
      name                      = "default-node-pool"
      machine_type              = "e2-medium"
      node_locations            = "us-central1-b,us-central1-c"
      min_count                 = 1
      max_count                 = 100
      local_ssd_count           = 0
      spot                      = false
      local_ssd_ephemeral_count = 0
      disk_size_gb              = 100
[...]

PR #1645, opened 2023-05-25
fix: set max firewall name to 36

NissesSenap:
This, to be able to have longer and unique names.
The firewall API supports 63 charters
Solves #1527

https://cloud.google.com/compute/docs/reference/rest/v1/firewalls

```shell
# The longest name with the extra chars added by the module
echo gke--intra-cluster-egress |wc -c
26
# the max ammount of char - the number of chars the module creates
expr 63 - 26
37
# I picked 36 since it looks better then 37 :)
```

### DIFF
--- a/autogen/main/firewall.tf.tmpl
+++ b/autogen/main/firewall.tf.tmpl
@@ -26,7 +26,7 @@
  *****************************************/
 resource "google_compute_firewall" "intra_egress" {
   count       = var.add_cluster_firewall_rules ? 1 : 0
-  name        = "gke-${substr(var.name, 0, min(25, length(var.name)))}-intra-cluster-egress"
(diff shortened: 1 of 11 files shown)


bharathkkb (commented):
Thanks for the PR @NissesSenap 

Looks like some of our tests may need updating

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d252579613ca44e11ad28a59f930cdb42dea2c29/test/integration/safer_cluster/safer_cluster_test.go#L66-L67

NissesSenap:
@bharathkkb I updated the tests as you pointed out

bharathkkb:
Looks like 36 is OOB since the test cluster name is smaller but I think we can just skip truncating since it's a static size from example.

bharathkkb (approved):

Issue #1646, opened 2023-05-25
Variable identity_namespace has no effect on beta-autopilot clusters


tedh-slalom:
### TL;DR

The modules `beta-autopilot-private-cluster` and `beta-autopilot-public-cluster` expose an `identity_namespace` to support non-default identity pools. The cluster resources do not respect these values, causing issues with multi-project Anthos configurations.

### Expected behavior

In project `project-a`, setting `identity_namespace` = `project-b.svc.id.goog` should result in a cluster with:

```
    workload_identity_config {
      workload_pool = "project-b.svc.id.goog"
    }
```

### Observed behavior

The cluster is provisioned with the default `project-a.svc.id.goog` identity pool.

### Terraform Configuration

```hcl
module "gke" {
  source = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster"
  
  # other fields omitted 

  project_id                      = var.project_id
  identity_namespace              = "${var.fleet_project_id}.svc.id.goog"
}
```

### Terraform Version

```sh
1.4.6
```

PR #1647, opened 2023-05-25
fix: beta-autopilot cluster modules respect identity_namespace

tbh1:
This change addresses #1646 by applying the existing `identity_namespace` variable on the `beta-autopilot-public-cluster` and `beta-autopilot-private-cluster` modules. This variable currently has no effect.


tbh1:
Testing revealed that `workload_identity_config` is currently marked as conflicting with `enable_autopilot` in the provider.

Closing this PR for now.

PR #1654, opened 2023-05-31
fix: allow using k8s current context

rochana-atapattu:
by allowing this we could use this module with private clusters with connect gateway to connect to the cluster


NissesSenap:
@rochana-atapattu you need to do this by changing in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/autogen/main
Also think if it's something that could be useful for the other submodules.

Then run `make build` and commit the generated changes, also run `make docker_test_lint` to validate the linting of your PR.

PR #1657, opened 2023-06-01
fix: policy bundles now use idx as resource name (bundles will be re-applied)

apeabody:
The module name is used by the gcloud module in paths, so long uls or special characters could cause issues depending on the local filesystem.

### DIFF
--- a/modules/acm/policy_bundles.tf
+++ b/modules/acm/policy_bundles.tf
@@ -18,12 +18,13 @@ module "policy_bundles" {
   source  = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
   version = "~> 3.1"
 
-  for_each                = toset(var.policy_bundles)
+  # Use index as name to avoid long url or special filesystem chars
+  for_each                = { for i, v in var.policy_bundles : i => v }
   project_id              = var.project_id
   cluster_name            = var.cluster_name
   cluster_location        = var.location
-  kubectl_create_command  = "kubectl apply -k ${each.key}"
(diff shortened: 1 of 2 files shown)


modules/acm/policy_bundles.tf
```diff
@@ -18,12 +18,13 @@ module "policy_bundles" {
   source  = "terraform-google-modules/gcloud/google//modules/kubectl-wrapper"
   version = "~> 3.1"
 
-  for_each                = toset(var.policy_bundles)
+  # Use index as name to avoid long url or special filesystem chars
+  for_each                = { for i, v in var.policy_bundles : i => v }
```
bharathkkb, line 22: Would this be okay for patch release since the module key is changing which could result in recreate?

bharathkkb (approved):
Thanks @apeabody, one question regarding upgrade

modules/acm/policy_bundles.tf
apeabody: My initial thought is this is declarative, so while it should result in a change (delete/apply) it would result in the same end cluster state.  I also considered using a `moved` block, but even if we weren't using `idx`, I didn't see a clear way to make it work with arbitrary resources.

@bharathkkb WDYT?

bharathkkb: Yeah it would be tricky without stable ID. I think we can just call this out in the readme as it would re apply.

apeabody: Updated the subject for the change log to clarify.

PR #1660, opened 2023-06-05
fix: add subnetwork as an output 

NissesSenap:
This way I can get rid of a data since the subnet name is looked up inside the module

### DIFF
--- a/autogen/main/outputs.tf.tmpl
+++ b/autogen/main/outputs.tf.tmpl
@@ -146,9 +146,9 @@ output "service_account" {
   value       = local.service_account
 }
 
-output "netowrk" {
+output "subnetwork" {
   description = "The subnetwork to host the cluster in"
-  value       = data.google_compute_network.network
+  value       = data.google_compute_subnetwork.gke_subnetwork
 }
 
 {% if autopilot_cluster != true %}
--- a/autogen/main/outputs.tf.tmpl
+++ b/autogen/main/outputs.tf.tmpl
@@ -146,6 +146,11 @@ output "service_account" {
   value       = local.service_account
 }
 
+output "netowrk" {
+  description = "The subnetwork to host the cluster in"
+  value       = data.google_compute_network.network
+}
+
 {% if autopilot_cluster != true %}
 output "instance_group_urls" {
   description = "List of GKE generated instance groups"
--- a/autogen/main/outputs.tf.tmpl
+++ b/autogen/main/outputs.tf.tmpl
@@ -147,7 +147,7 @@ output "service_account" {
 }
 
 output "subnetwork" {
-  description = "The subnetwork to host the cluster in"
+  description = "The subnetwork hosting the cluster"
   value       = data.google_compute_subnetwork.gke_subnetwork
 }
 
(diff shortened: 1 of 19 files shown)


autogen/main/outputs.tf.tmpl
```diff
@@ -146,6 +146,11 @@ output "service_account" {
   value       = local.service_account
 }
 
+output "netowrk" {
```
apeabody, line 149: typo

```suggestion
output "network" {
```

autogen/main/outputs.tf.tmpl
```diff
@@ -146,6 +146,11 @@ output "service_account" {
...
+  description = "The subnetwork to host the cluster in"
```
apeabody, line 150: I assume this should be network instead of subnetwork?

```suggestion
  description = "The cluster's network"
```

apeabody (changes requested):
Thanks for the contribution @NissesSenap!  A few notes:

autogen/main/outputs.tf.tmpl
NissesSenap: I did one to many things when I created this PR.... It should actually be subnetwork.
I'm running the linting as well now to make sure that it passes. 
I will push the fix as soon as it's done.

NissesSenap:
@apeabody it should now be fixed, the linting passed on my local host as well. I will remember to do that in the future.

autogen/main/outputs.tf.tmpl
```diff
@@ -146,6 +146,11 @@ output "service_account" {
   value       = local.service_account
 }
 
+output "subnetwork" {
+  description = "The subnetwork to host the cluster in"
```
apeabody, line 150: nit: would this be more clear?

```suggestion
  description = "The subnetwork hosting the cluster"
```

NissesSenap:
@apeabody committed and regenerated.
Thanks for taking the time.

bharathkkb (commented):
Thanks for the PR @NissesSenap 

I am not sure if there is much value in adding this because the data lookup is conditional and that var is false by default.

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d252579613ca44e11ad28a59f930cdb42dea2c29/networks.tf#L22

NissesSenap:
There are many useful settings in this module that are off by default. 

It's no criticism, i understand why the module is like it is, but I'm just trying to slim down all the extra config i need to do outside the module as much as i can.

I guess it will just be a null value for the users that don't have this setting enabled.

Else i will just have to live with my extra data resource.

autogen/main/outputs.tf.tmpl
```diff
@@ -146,6 +146,11 @@ output "service_account" {
   value       = local.service_account
 }
 
+output "subnetwork" {
+  description = "The subnetwork hosting the cluster"
```
bharathkkb, line 150: @NissesSenap could you clarify it will be null value if fw option is not set

NissesSenap: I assume that if the `data.google_compute_subnetwork.gke_subnetwork` isn't set due to https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d252579613ca44e11ad28a59f930cdb42dea2c29/networks.tf#L22 not being set the output value will be null.

But I'm a bit unsure how terraform works here. I don't think we have a simple if statement that we can set to template the output away.
Can you do a ` count   = var.add_cluster_firewall_rules ? 1 : 0` on an output?
It seems a bit strange since it's the interface we provide to the end user.

PR #1665, opened 2023-06-14
feat(TPG>=4.32.0)!: Support enabling Policy Controller mutations

ferrarimarco:
This PR exposes the `mutation_enabled` option. Without this, users are forced to manually edit the configuration descriptor ([example](https://github.com/GoogleCloudPlatform/gke-securing-third-party-apps-blueprint/blob/e54cafe9456d515d09d56f5c0205f4e6a61ae2d5/terraform/scripts/enablePolicyControllerMutations.sh)).

I also updated the minimum google-beta provider version to [4.32.0](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/CHANGELOG.md#4320-august-15-2022) because that's when the `mutation_enabled` option was implemented there.


apeabody (approved):
Thanks for the contribution @ferrarimarco!

PR #1696, opened 2023-07-23
feat!: support for enabling image streaming at cluster level

### LINKED ISSUE #1360: [beta-private-cluster-update-variant] Enable image streaming at cluster level
### TL;DR

The documentation describes the `enable_gcfs` flag for node pools, but this does not seem to enable the cluster-level image streaming option

### Terraform Resources

_No response_

### Detailed design

_No response_

### Additional information

We are defining our node pools like this, [according to the documentation](https://registry.terraform.io/modules/terraform-google-modules/kubernetes-engine/google/latest#node_pools-variable), with the `enable_gcfs = true` flag:

```terraform
    {
      name               = "pool"
      preemptible        = false
      spot               = true
      enable_secure_boot = true
      enable_gcfs        = true
      machine_type       = "t2d-standard-4"
      min_count          = 0
      max_count          = 10
      max_surge          = 4
      image_type         = "COS_CONTAINERD"
    },
```

However, after the cluster is created, the Image Streaming option is not enabled in the UI:

image

lauraseidler:
Could previously be set manually via console/`gcloud` - if this was enabled, `enable_gcfs` must be set to `true` with this update, otherwise image streaming will be disabled.

Fixes #1360


apeabody (approved):
Thanks for the contribution @lauraseidler!

Good call on the breaking change `!` flag.

Jojoooo1:
Any reason it is not supported for private cluster ?

Issue #1442, opened 2022-10-30
GKE Cluster gets destroyed and reconfigured on reapply


jlevertov:
### TL;DR

Because of the zones random shuffle, when the cluster is regional no matter what, when I reapply the module (even if no change was made) terraform reads the cluster as needs to be recreated (destroyed and created)

### Expected behavior

When no change is made in the module configuration, nothing should happen.

### Observed behavior

when I reapply the module (even if no change was made) the cluster is being recreated (destroyed and created)
Plan example:
```
 # module.cloud-infra.module.gke_configuration.module.gke.random_shuffle.available_zones must be replaced
-/+ resource "random_shuffle" "available_zones" {
      ~ id           = "-" -> (known after apply)
      ~ input        = [
          - "zone-a",
          - "zone-b",
          - "zone-c",
        ] -> (known after apply) # forces replacement
      ~ result       = [
          - "zone-b",
          - "zone-a",
          - "zone-c",
        ] -> (known after apply)
        # (1 unchanged attribute hidden)
    }
```
Logs Example:
```
module.cloud-infra.module.gke_configuration.module.gke.random_shuffle.available_zones: Destroying... [id=-]
module.cloud-infra.module.gke_configuration.module.gke.random_shuffle.available_zones: Destruction complete after 0s
.
.
.
module.cloud-infra.module.gke_configuration.module.gke.data.google_compute_zones.available: Read complete after 0s [id=<redacted>]
module.cloud-infra.module.gke_configuration.module.gke.data.google_container_engine_versions.zone: Reading...
module.cloud-infra.module.gke_configuration.module.gke.random_shuffle.available_zones: Creating...
module.cloud-infra.module.gke_configuration.module.gke.random_shuffle.available_zones: Creation complete after 0s [id=-]
module.cloud-infra.module.gke_configuration.module.gke.data.google_container_engine_versions.zone: Read complete after 1s [id=2022-10-30 15:07:05.538703735 +0000 UTC]
```

### Terraform Configuration

```hcl
module "gke" {
  source                            = "terraform-google-modules/kubernetes-engine/google//modules/private-cluster"
  version                           = "23.0.0"
  project_id                        = var.project_id
  name                              = var.config.name
  kubernetes_version                = var.config.kubernetes_version
  regional                          = true
  region                            = var.config.region
  zones                             = [zone-a, zone-b, zone-c]
  network                           = var.config.network_config.name
[...]

PR #1707, opened 2023-08-10
Stop random_shuffle for available_zones

NissesSenap:
This way we will still get random zones, but only the first time the cluster is created. Solves #1442


ericyz:
Cloud Build Error:

 Inappropriate value for attribute "keepers": element "zone_names": string

ericyz:
Thanks for your contribution. I would suggest to conditionally query the data source only when the var.zones are not provided.

Raised the fix (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1709/files) as per comment in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1442

I will close this for now.

Issue #1706, opened 2023-08-09
monitoring_enable_managed_prometheus field not working in v27.0.0


prateekn:
### TL;DR

Inspite of setting  "monitoring_enable_managed_prometheus" variable equal to false on version v27.0.0 of this terraform module
"Managed Service for Prometheus" in gcp still **enabled**

gke version 1.27.4-gke.900

### Expected behavior

Managed Service for Prometheus should be disabled.

### Observed behavior

Managed Service for Prometheus is enabled.

![error](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/assets/1693166/5dc228a3-1a76-4a54-a06b-94111659cca1)

### Terraform Configuration

```hcl
module "gke" {
  source                 = "terraform-google-modules/kubernetes-engine/google/modules/beta-private-cluster-update-variant"
  version                = "v27.0.0"
  description                   = "GKE Prod Cluster"
  kubernetes_version            = var.kubernetes_version
  regional                      = true
  region                        = var.region
  network                       = var.network
  network_project_id            = var.network_project_id
  subnetwork                    = var.subnetwork
  ip_range_pods                 = var.ip_range_pods
  ip_range_services             = var.ip_range_services
  create_service_account        = false
  service_account               = var.default_cluster_sa
  add_cluster_firewall_rules    = true
  firewall_inbound_ports        = ["8443", "9443", "15017"]
  default_max_pods_per_node     = 32
  monitoring_enable_managed_prometheus = false
  http_load_balancing           = true
  network_policy                = false
  horizontal_pod_autoscaling    = true
  filestore_csi_driver          = false
  enable_private_endpoint       = true
  enable_private_nodes          = true
  master_ipv4_cidr_block        = "10.10.20.0/28"
  remove_default_node_pool      = true
  gce_pd_csi_driver             = true
  enable_intranode_visibility   = false
  enable_binary_authorization   = true
  database_encryption           = [{
    state = "ENCRYPTED"
    key_name = data.google_kms_crypto_key.gke_key_1.id
  }]
  release_channel               = "UNSPECIFIED"
  grant_registry_access         = false
  node_metadata                 = "GKE_METADATA"
  logging_enabled_components    = ["SYSTEM_COMPONENTS"]
  monitoring_enabled_components = ["SYSTEM_COMPONENTS"]
  gateway_api_channel           = "CHANNEL_STANDARD"
  master_authorized_networks = [    
    {
[...]

PR #1715, opened 2023-08-22
fix: make promethus configurable

ericyz:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1706


ericyz:
Hi @apeabody , could you please help to review the change? It passed the all checks.

autogen/main/cluster.tf.tmpl
```diff
@@ -113,17 +113,15 @@ resource "google_container_cluster" "primary" {
...
-          enabled = var.monitoring_enable_managed_prometheus
-        }
+    {% if beta_cluster %}
+    for_each = local.cluster_telemetry_type_is_set || local.logmon_config_is_set ? [1] : []
+    {% else %}
+    for_each = local.logmon_config_is_set || local.logmon_config_is_set ? [1] : []
+    {% endif %}
+    content{
+      enable_components = var.monitoring_enabled_components
+      managed_prometheus {
```
apeabody, line 123: Thanks @ericyz - As this is change from a dynamic to static block, likely some existing clusters will see config churn/change?  As as the gating variables are different.  Should this be a breaking change?

ericyz: Good point. It should be a break change

ericyz: Good point. It should be a breaking change

apeabody (approved):
Thank @ericyz!

PR #1694, opened 2023-07-18
feat: Support for Security Posture for safer cluster variants

ebeltramo96:
followup to: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1658 (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1617)

PR #1719, opened 2023-08-24
feat: Add security posture

dave-ledgy:
Adds support for security posture configuration - https://cloud.google.com/kubernetes-engine/docs/concepts/about-security-posture-dashboard


jon-walton:
Hi @dave-ledgy . Thanks for adding this, I was also looking to use the same features. I'm assuming that before the PR will get reviewed, you'll need to sign the CLA. Is that something you're looking to do?

ebeltramo96:
duplicated https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1694

PR #1749, opened 2023-09-27
feat!: enabling vulnerability and audit modes for workloads

mmorejon:
Enable vulnerability and audit modes for workloads.

### DIFF
--- a/autogen/safer-cluster/variables.tf.tmpl
+++ b/autogen/safer-cluster/variables.tf.tmpl
@@ -500,7 +500,7 @@ variable "enable_mesh_certificates" {
 variable "workload_vulnerability_mode" {
   description = "(beta) Vulnerability mode."
   type        = string
-  default     = ""
+  default     = null
 }
 
 variable "workload_config_audit_mode" {
--- a/autogen/safer-cluster/variables.tf.tmpl
+++ b/autogen/safer-cluster/variables.tf.tmpl
@@ -500,7 +500,7 @@ variable "enable_mesh_certificates" {
 variable "workload_vulnerability_mode" {
   description = "(beta) Vulnerability mode."
   type        = string
-  default     = null
+  default     = ""
 }
 
 variable "workload_config_audit_mode" {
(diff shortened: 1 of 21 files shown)


autogen/safer-cluster/variables.tf.tmpl
```diff
@@ -496,3 +496,15 @@ variable "enable_mesh_certificates" {
   default = false
   description = "Controls the issuance of workload mTLS certificates. When enabled the GKE Workload Identity Certificates controller and node agent will be deployed in the cluster. Requires Workload Identity."
 }
+
+variable "workload_vulnerability_mode" {
+  description = "(beta) Vulnerability mode."
+  type        = string
+  default     = ""
```
apeabody, line 503: Are these new default values going to cause a replacement for existing clusters? Could we use `null`?

mmorejon: `workload_config_audit_mode` cannot be `null`.  This value is required. I got this message setting the value as `null`.
```
The argument "protect_config.0.workload_config.0.audit_mode" is required, but no definition was found.
```

`workload_vulnerability_mode` has the same behavior using an empty string or null value. I got a change in the tfstate only when `DISABLED` or `BASIC` is used.
```
~ protect_config {
   ~ workload_vulnerability_mode = "WORKLOAD_VULNERABILITY_MODE_UNSPECIFIED" -> "DISABLED"
}
```
and
```
~ protect_config {
   ~ workload_vulnerability_mode = "WORKLOAD_VULNERABILITY_MODE_UNSPECIFIED" -> "BASIC"
}
```

mmorejon: @apeabody , what do you think?

mmorejon:
Do you have any other suggestions? @apeabody

autogen/safer-cluster/variables.tf.tmpl
mmorejon: Done. I set back the default value to empty.

apeabody (approved):
Thanks for the contribution @mmorejon!

PR #1752, opened 2023-09-28
fix(acm): Allow to enable config_sync or policy_controller standalone

foosinn:
Currently enabling only one of `enable_config_sync` or `enable_policy_controller` will cause a string template containing `null`. See:

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/acm/creds.tf#L47

This removes the null item and creates instead an array of 0 to 2 elements.


apeabody (approved):
Thanks for the contribution @foosinn!

Issue #1632, opened 2023-05-10
Parametrise `enable_binary_authorization` for `safer-cluster`


thevops:
### TL;DR

Binary Authorization costs and it's useful when used. But when it's unused then it only takes money.

Please, could you add a variable for `enable_binary_authorization` (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/safer-cluster/main.tf#L161) to make possible disabling this feature for a cluster?

PR #1760, opened 2023-10-06
feat: make `enable_binary_authorization` optional

thevops:
By default `enable_binary_authorization` is set to `true` which means that the Binary Authorization feature is enabled. This is the current value of the variable in the `safer-cluster` [module](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/safer-cluster/main.tf#L161).

That feature makes sense only when it is used. Until it is not used, it only takes money for nothing.

Fixes: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1632


JosefBocan:
@Jberlinsky @ericyz - To resolve this issue will solve my problem with using your source terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster due the failing on set of `enable_binary_authorization`
Thanks guys ❤️

PR #1796, opened 2023-11-07
fix: alpha option for cluster creation

prudnitskiy:
This MR fixes cluster creation if `enable_kubernetes_alpha` is set to true. The problem is default pool created with `auto_repair = true` and `auto_upgrade = true` which prohibits cluster creation even if `remove_default_node_pool` set to `true`


apeabody (approved):
Thanks for the contribution @prudnitskiy

Issue #730, opened 2020-11-02
IAP Bastion Host example Tinyproxy improvements


bharathkkb:
Currently the [Safer Cluster Access with IAP Bastion Host](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples/safer_cluster_iap_bastion#safer-cluster-access-with-iap-bastion-host) example does not monitor the status of Tinyproxy. We should consider adding a custom conf file which writes to a logfile and additionally look into monitoring options for surfacing any errors. Alternatively we could include some troubleshooting documentation differentiating a Tinyproxy issue from IAP/Bastion host/ K8s connectivity issues.

karlkfi:
Does tinyproxy even get started? It seems like the startup script just installs it and doesn't configure or start it.

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/examples/safer_cluster_iap_bastion/bastion.tf#L25

karlkfi:
Hmm. I had to install and configure tiny proxy manually:

```
# SSH into the bastion-host VM
gcloud beta compute ssh bastion-host --tunnel-through-iap --project ${PLATFORM_PROJECT_ID} --zone ${BASTION_ZONE}

# Become root
sudo su

# Create Tinyproxy user
useradd -M -U -s /bin/false tinyproxy

# Create Tinyproxy log file
mkdir -p /usr/local/var/log/tinyproxy
touch /usr/local/var/log/tinyproxy/tinyproxy.log
chown tinyproxy:root /usr/local/var/log/tinyproxy/tinyproxy.log

# Configure Tinyproxy
mkdir -p /usr/local/etc/tinyproxy
cat > /usr/local/etc/tinyproxy/tinyproxy.conf << EOF
##User/Group to use after dropping root
User tinyproxy
Group tinyproxy

##Port and address to bind to
Port 8888
Bind 0.0.0.0

##File locations
DefaultErrorFile "/usr/local/share/tinyproxy/default.html"
StatFile "/usr/local/share/tinyproxy/stats.html"
LogFile "/usr/local/var/log/tinyproxy/tinyproxy.log"
LogLevel Info
PidFile "/var/run/tinyproxy.pid"

##Authentication
Allow 127.0.0.1

##HTTP Headers
ViaProxyName "server-hostname"
DisableViaHeader No

##Threading
StartServers 5
MinSpareServers 5
MaxSpareServers 10 
[...]

PR #1798, opened 2023-11-08
chore: Configure tinyproxy to always restart if killed.

grugnog:
See https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/730#issuecomment-770431179

This uses an override rather than modifying the package owned configuration, which is cleaner and won't be affected by package upgrades.


examples/safer_cluster_iap_bastion/templates/startup-script.tftpl
```diff
@@ -1,2 +1,6 @@
 sudo apt-get update -y
 sudo apt-get install -y tinyproxy
+sudo mkdir -p /etc/systemd/system/tinyproxy.service.d/
+echo -e '[Service]\nRestart=always' | sudo tee /etc/systemd/system/tinyproxy.service.d/override.conf
+sudo systemctl daemon-reload
+sudo systemctl restart tinyproxy
```
apeabody, line 6: nit: add end of file line
```suggestion
sudo systemctl restart tinyproxy

```

Issue #1808, opened 2023-12-01
Random suffix for service account generated even when `create_service_account` is false


lucetre:
### TL;DR

When setting `create_service_account` to false in the private-cluster module, the `random_string.cluster_service_account_suffix` resource is still being evaluated and generating a random suffix even though it should not be created.

### Expected behavior

When `create_service_account` is set to false, the `random_string.cluster_service_account_suffix` resource should not be evaluated or generate a random suffix.

<img width="673" alt="image" src="https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/assets/26702243/b214ce3a-5afb-47ce-8fef-e66770118de1">

### Observed behavior

The `random_string.cluster_service_account_suffix` resource is being evaluated and generating a random suffix even when `create_service_account` is set to false.

<img width="775" alt="image" src="https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/assets/26702243/b6a2b403-032c-4958-be78-41fc003899c8">

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/24ccc6ba3a938d1e2ef88a0173c7849d6fa26d0c/sa.tf#L34-L39

### Terraform Configuration

```hcl
module "private_cluster_eventfnt_asia_south1_1" {
  source  = "terraform-google-modules/kubernetes-engine/google//modules/private-cluster"
  version = "~> 29.0.0"

  project_id                = local.project_id
  name                      = local.eventfnt_asia_south1_1_name
  regional                  = true
  region                    = local.eventfnt_asia_south1_1_config.region
  zones                     = local.eventfnt_asia_south1_1_config.zones
  release_channel           = local.eventfnt_asia_south1_1_config.release_channel
  default_max_pods_per_node = local.cluster_preset.default_max_pods_per_node
  enable_cost_allocation    = local.enable_cost_allocation

  network_project_id     = local.eventfnt_asia_south1_1_config.network_project_id
  network                = local.eventfnt_asia_south1_1_config.network_name
  subnetwork             = local.eventfnt_asia_south1_1_config.subnetwork_name
  master_ipv4_cidr_block = local.eventfnt_asia_south1_1_config.master_ipv4_cidr_block
  ip_range_pods          = local.eventfnt_asia_south1_1_config.ip_range_pods_name
  ip_range_services      = local.eventfnt_asia_south1_1_config.ip_range_services_name

  identity_namespace           = "enabled"
[...]

PR #1809, opened 2023-12-01
fix: handle random suffix only when `create_service_account` is false - Issue #1808

lucetre:
### Description
This pull request addresses issue #1808, which pertains to the generation of random suffixes for service account names even when create_service_account is set to false. The changes ensure that the random suffix logic is conditioned on the `create_service_account `setting.

### Changes Made

1. Updated the `service_account_default_name` local variable in sa.tf to conditionally include the random suffix only when `create_service_account` is true.
2. Modified the `random_string` resource to have a count condition, ensuring it is created only when `create_service_account` is true.

### Testing
- [x] Ensure the `create_service_account` variable is set to `false`.
- [x] Verify that the `random_string.cluster_service_account_suffix` resource is not created when `create_service_account` is `false`.
- [x] Confirm that the service account name is generated correctly without a random suffix in the absence of `random_string.cluster_service_account_suffix`.

<details>
<summary>Testing codes</summary>

```
variable "name" {
  type    = string
  default = ""
}

variable "create_service_account" {
  type    = bool
  default = false
}

variable "service_account_name" {
  type    = string
  default = ""
}

locals {
  service_account_default_name = "tf-gke-${substr(var.name, 0, min(15, length(var.name)))}-${var.create_service_account ? random_string.cluster_service_account_suffix[0].result : ""}"
}

resource "random_string" "cluster_service_account_suffix" {
  count   = var.create_service_account ? 1 : 0
  upper   = false
  lower   = true
  special = false
  length  = 4
}

  count        = var.create_service_account ? 1 : 0
  project      = var.project_id
  account_id   = var.service_account_name == "" ? local.service_account_default_name : var.service_account_name
  display_name = "Terraform-managed service account for cluster ${var.name}"
}
```
</details>

### Related Issue
- #1808


Bo98:
Thanks - I was looking for the same thing.

Though I think you need to modify `autogen/main/sa.tf.impl` instead and then run `make build`.

PR #1827, opened 2023-12-21
fix!: Revert create least privilege default service account (#1757)

apeabody:
This reverts commit 350faa747659202de58bef64de67c94845b223ff.

### DIFF
--- a/autogen/main/sa.tf.tmpl
+++ b/autogen/main/sa.tf.tmpl
@@ -45,10 +45,31 @@ resource "google_service_account" "cluster_service_account" {
   display_name = "Terraform-managed service account for cluster ${var.name}"
 }
 
-resource "google_project_iam_member" "cluster_service_account-nodeService_account" {
+resource "google_project_iam_member" "cluster_service_account-log_writer" {
   count   = var.create_service_account ? 1 : 0
   project = google_service_account.cluster_service_account[0].project
-  role    = "roles/container.nodeServiceAccount"
+  role    = "roles/logging.logWriter"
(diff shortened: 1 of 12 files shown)


gorge511:
Hi @apeabody, why this is reverted to old solution and not to the new generally available role `roles/container.nodeServiceAgent`? Is there any reasoning I'm not aware of?

apeabody:
> roles/container.nodeServiceAgent

Hi @gorge511 - I believe at the time of the revert that `roles/container.nodeServiceAgent` wasn't yet generally available. Would be happy to review a PR to switch it to now. We can include in the upcoming release.

apeabody:
> > roles/container.nodeServiceAgent
> 
> Hi @gorge511 - I believe at the time of the revert that `roles/container.nodeServiceAgent` wasn't yet generally available. Would be happy to review a PR to switch it to now. We can include in the upcoming release.

I haven't had a chance to dig into the details, but looks like use of an *Agent* role isn't recommended for this propose: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1757#issuecomment-1889614996

gorge511:
> I haven't had a chance to dig into the details, but looks like use of an _Agent_ role isn't recommended for this propose: [#1757 (comment)](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1757#issuecomment-1889614996)

Thanks for the quick response. Yes, it seems that the Agent role is not the right one, even if it looks really good for this use case 🙂 
I guess we should continue with the discussion in #1757, there are more people in the discussion.

PR #1828, opened 2023-12-22
fix: readme

teeheehs:
- Had to dig around for the terraform references when using this, thought I'd add as links to module :)


apeabody (approved):
Thanks for the contribution @teeheehs!

PR #1721, opened 2023-08-25
feat(multi nic): support enabling multi networking

sebastiandero:

PR #1829, opened 2023-12-23
feat: add beta multi networking options

dzoeteman:
Add beta multi-networking options.
This is lightly based on the work in #1721, but that PR was closed due to inactivity.

- Implemented in autogen template
- Had to always include `network_config` block on the node pool due to otherwise too complicated if statement depending on beta cluster or not; this should be fine anyway with the defaults of the option in the block.
- Tested on GCP.


dzoeteman:
@apeabody Hi, could I ask for a `/gcbrun `? Happy to update branch as well 🙇

PR #1830, opened 2023-12-23
feat: add master_global_access_enabled for public clusters

dzoeteman:
Add master_global_access_enabled flag to public clusters too.
The underlying option is available in GCP APIs and UI for public clusters as well.
Also removed 'Beta' label from the other private cluster config options since they're (no longer) beta.


dzoeteman:
@apeabody Could I ask for a `/gcbrun ` here as well? 🙇

Issue #1671, opened 2023-06-14
asm labelling docs unclear due to ignored var


zeemy23:
### TL;DR

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/asm/README.md

These docs guide the user to add the mesh_id label in a way that this PR from 2021 discontinued support for. 

 https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1061

I can add a PR that clarifies the readme for future users. We were pulling our hair out trying to figure out why the label was getting ignored.

### Expected behavior

following asm docs results in asm functionality

### Observed behavior

ignored labels causes much strife if following docs

### Terraform Configuration

```hcl
module "gke" {
...
  cluster_resource_labels = { "mesh_id" : "proj-${data.google_project.project.number}" }
...
}
```

### Terraform Version

```sh
all versions.
```

legal90:
That issue is still actual. Should it be reopened? cc @apeabody , @ferrarimarco 

Both ASM tf module documentation, as wells as the official Google docs are saying that `mesh_id` label has to be added explicitly to the cluster:
- https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/3c9a634ceef3fe8fc073de161f17f0064f424c04/modules/asm/README.md
- https://cloud.google.com/service-mesh/docs/managed/provision-managed-anthos-service-mesh#apply_the_mesh_id_label

But, as @zeemy23 said above, this label is ignored on gke tf modules:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/eeebd782df42bc197638f4fd291ee99a680002fc/autogen/main/cluster.tf.tmpl#L248-L250
It was added by https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1061.

apeabody:
> That issue is still actual. Should it be reopened? cc @apeabody , @ferrarimarco
> 
> Both ASM tf module documentation, as wells as the official Google docs are saying that `mesh_id` label has to be added explicitly to the cluster:
> 
> * https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/3c9a634ceef3fe8fc073de161f17f0064f424c04/modules/asm/README.md
> * https://cloud.google.com/service-mesh/docs/managed/provision-managed-anthos-service-mesh#apply_the_mesh_id_label
> 
[...]

PR #1836, opened 2024-01-04
fix: Do not ignore "mesh_id" label on "google_container_cluster" resource

legal90:
Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1671
Relates to https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1061

According to the documentation, `mesh_id` label is supposed to be set explicitly on the GKE cluster:
- [asm module doc] https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/d82c50c9cd059fd581f7ab5fe943dfada3cce0c3/modules/asm/README.md#L36-L43
- [Google doc for managed ASM] https://cloud.google.com/service-mesh/docs/managed/provision-managed-anthos-service-mesh#configure_your_fleet
- [Google doc for in-cluster ASM] https://cloud.google.com/service-mesh/docs/unified-install/project-cluster-setup#set_up_your_cluster

However, right now this label is ignored by terraform, so it's not possible to set it using IaC. 
This PR fixes that issue and removes `mesh_id` label from the `ignore_changes` list

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -399,7 +399,7 @@ resource "google_container_cluster" "primary" {
 
   {% if autopilot_cluster != true %}
   lifecycle {
-    ignore_changes = [node_pool, initial_node_count, resource_labels["asmv"], resource_labels["mesh_id"]]
+    ignore_changes = [node_pool, initial_node_count, resource_labels["asmv"]]
   }
   {% endif %}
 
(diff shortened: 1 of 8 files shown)


ferrarimarco (approved):

apeabody (approved):

Thanks for the review @ferrarimarco!

PR #1839, opened 2024-01-10
feat(TPG>=5.9)!: cluster autoscaling profile is GA

### LINKED ISSUE #1820: Feature: Add autoscaling profiles to non beta submodules
### TL;DR

Since https://github.com/hashicorp/terraform-provider-google/pull/16653 was merged and https://github.com/hashicorp/terraform-provider-google/releases/tag/v5.9.0 released, now the `autoscaling_profile` is GA and available out of BETA.

I think it should be basically to do the same as https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/456 but in the non beta modules.

This week I don't have that much time, but if no one has done it in the next weeks, I will probably open a PR with the changes.

### Terraform Resources

```markdown
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#autoscaling_profile
```

### Detailed design

_No response_

### Additional information

_No response_

florianmutter:
GKE autoscaling profile is GA now

Closes #1820

### DIFF
--- a/test/integration/go.mod
+++ b/test/integration/go.mod
@@ -6,7 +6,7 @@ toolchain go1.21.5
 
 require (
 	github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/blueprint-test v0.11.1
-	github.com/gruntwork-io/terratest v0.46.9
+	github.com/gruntwork-io/terratest v0.46.11
 	github.com/stretchr/testify v1.8.4
 )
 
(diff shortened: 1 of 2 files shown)


florianmutter:
@apeabody could you review maybe?

apeabody (changes requested):
Thanks for the contribution @florianmutter!

We'll also need to bump the minimum provider in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/versions.tf.tmp

Ref: https://github.com/hashicorp/terraform-provider-google/releases/tag/v5.9.0

florianmutter:
Update the required provider version for the regular provider. Should the beta provider be aligned? I don't think it is necessary but might prevent confusion on why the version requirements differ.

florianmutter:
Updated both versions

florianmutter:
@apeabody ping :)

apeabody:
re-trigger

```
STDERR: Error: Provider produced inconsistent result after apply

When applying changes to
module.example.module.workload_identity.google_service_account.cluster_service_account[0],
provider "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
unexpected new value: Root object was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
---- End output 
```

apeabody (approved):
Thanks for the contribution @florianmutter!

florianmutter:
@apeabody any chance we can get a new release of the module? Seems like an ok list of changes for a new release to me :D https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/compare/v29.0.0...master

PR #1848, opened 2024-01-17
feat: transition policy controller

apeabody:


evenh:
Looking forward to this PR as I'm currently getting an infinite diff when disabling `policy_controller` using the following properties:

```
  enable_fleet_registration = false
  enable_fleet_feature      = false
  enable_config_sync        = false
  enable_policy_controller  = false

PR #1860, opened 2024-01-31
feat: add optional membership_location to fleet-membership

apeabody:


modules/fleet-membership/membership.tf
```diff
@@ -21,6 +21,7 @@ resource "google_gke_hub_membership" "primary" {
 
   project       = local.hub_project_id
   membership_id = local.gke_hub_membership_name
+  location      = var.membership_location
```
bharathkkb, line 24: Do we need to bump tpb/b version for this?

bharathkkb (approved):

modules/fleet-membership/membership.tf
bharathkkb: Yup looks like we are past 5.1.0 
https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#510-oct-9-2023

apeabody: Yeah - https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1824 already bumped past what this required

PR #1864, opened 2024-02-02
feat: add missing dns_cache output in non beta

DrFaust92:
no functional change


apeabody (approved):

PR #1867, opened 2024-02-05
feat: Enabling backups add-on for GKE autopilot clusters.

kamal-shaker:
Enabling the GKE backup option for auto-pilot clusters, which is now GA.


apeabody (changes requested):
Thanks for the contribution @kamal-shaker!

Can you please build your changes from the source template as described here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

DrFaust92:
it looks like its already supported for autopilot clusters in module. cc @apeabody can be closed

apeabody:
Thanks @DrFaust92 , yes, this is currently available for Autopilot clusters: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/beta-autopilot-private-cluster/cluster.tf#L139

PR #1776, opened 2023-10-20
feat: add advanced datapath observability config option

TheKangaroo:
fixes #1755 

I'm not sure about the variable names though. Any ideas for shorter names?

PR #1873, opened 2024-02-09
feat: Allow enabling observability metrics in the safer_cluster module

marcusramberg:
Allow using #1776 in safer_cluster module.


marcusramberg:
I'm closing this as it seems like it's not gonna get any attention and we've stopped using the safer_cluster module.

PR #1878, opened 2024-02-15
feat: add fleet block

apeabody:


apeabody:
need to add test coverage

modules/private-cluster/variables.tf
```diff
@@ -721,3 +721,15 @@ variable "config_connector" {
   description = "Whether ConfigConnector is enabled for this cluster."
   default     = false
 }
+
+variable "fleet_register" {
```
bharathkkb, line 725: Do we need a bool val for intent or could we just use presence of `fleet_project`?

bharathkkb (approved):
Thanks @apeabody, minor question

modules/private-cluster/variables.tf
apeabody: Hmm - Technically `fleet_project` can be optional and use the default provider project for fleet registration, but perhaps it would be better to require the fleet_project for clarity and use it for both.  I'll make that change. 👍

apeabody: @bharathkkb also due to: https://github.com/hashicorp/terraform-provider-google/issues/16680

PR #1883, opened 2024-02-21
docs: add multicluster configuration note

patrickblackjr:
This PR adds a section about multiple clusters using the Kubernetes provider.

This is more general Terraform knowledge, rather than module-specific, but this was still the first time I'd encountered this so it required some extra research. Thus, I thought it would be handy for other people to have available.


patrickblackjr:
This is still awaiting review.

apeabody (approved):
Thanks for the contribution @patrickblackjr!

PR #1917, opened 2024-03-28
feat: added option for external identity providers

ssergeby:
feat: added option for external identity providers to authenticate to GKE "enable_identity_service" and a new version of google provider for private-cluster module


apeabody (changes requested):
Thanks for the contribution @ssergeby!

Can you please make your change in the source template as described here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

DrFaust92:
this is already a supported flag, this can be closed @apeabody

apeabody:
Thanks, added in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1997

PR #1920, opened 2024-04-03
feat: network tags for autoprovisioned node pools

### LINKED ISSUE #1649: No ability to use node_pool_auto_config outside of auto pilot cluster.
There is currently no way to specify node_pool_auto_config settings for other cluster types such as private clusters. We have enabled node auto provisioning, but cannot add NAP tags to the automatically provisioned node pools.

This just needs the same logic as is used in the auto pilot module adding to other modules, such as private clusters.

g-awmalik:
Fixes #1649

### DIFF
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -306,13 +306,12 @@ variable "node_pools_oauth_scopes" {
 }
 {% endif %}
 
-{% if autopilot_cluster %}
 variable "network_tags" {
   description = "(Optional, Beta) - List of network tags applied to auto-provisioned node pools."
   type = list(string)
   default = []
 }
-{% endif %}
+
 variable "stub_domains" {
   type        = map(list(string))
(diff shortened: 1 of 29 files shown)


autogen/main/variables.tf.tmpl
```diff
@@ -306,13 +306,12 @@ variable "node_pools_oauth_scopes" {
 }
 {% endif %}
 
-{% if autopilot_cluster %}
 variable "network_tags" {
   description = "(Optional, Beta) - List of network tags applied to auto-provisioned node pools."
```
apeabody, line 310: Looks like this no longer needs to be marked as beta?

```suggestion
  description = "(Optional) - List of network tags applied to auto-provisioned node pools."
```

apeabody (approved):
Thanks @g-awmalik!

autogen/main/variables.tf.tmpl
g-awmalik: Ack.

PR #1921, opened 2024-04-04
Adding pod_pids_limit for kubelet_config in all modules, examples and autogen

djsuszi:
it's one of option for kubelet configuration
    https://cloud.google.com/kubernetes-engine/docs/how-to/node-system-config
    default none
    This setting sets the maximum number of process IDs (PIDs) that each Pod can use.
    Value must be must be between 1024 and 4194304


djsuszi:
created new valid one
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1922

PR #1922, opened 2024-04-04
feat: add pod_pids_limit for kubelet_config in all modules, example and autogen

djsuszi:
it's one of option for kubelet configuration
https://cloud.google.com/kubernetes-engine/docs/how-to/node-system-config
default none
This setting sets the maximum number of process IDs (PIDs) that each Pod can use.
Value must be must be between 1024 and 4194304


apeabody (approved):
Thanks for the contribution @djsuszi!

apeabody (approved):
Thanks for the contribution @djsuszi!

PR #1937, opened 2024-05-11
fix: local nvme ssd count

florianMalbranque:
allow more than 1 nvme ssd disk to be attached to a given compute engine


florianMalbranque:
👋 @apeabody, PTAL 😄

apeabody (approved):
Thanks for the contribution @florianMalbranque!

PR #1943, opened 2024-05-14
Adding extra permissions to the cluster's default service account

jullianow:
Recently, I started testing a component suggested by Google to consume GMP metrics (Google Manager Prometheus).
Ref: https://cloud.google.com/stackdriver/docs/managed-prometheus/hpa

This component, until the version of this module in 28.0.0, worked normally, and after I updated to version 30.2.0, it stopped working.
Trying to understand better, I realized that one of the main changes in the module version was due to this [PR](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1757).

What happens is that the role `roles/container.nodeServiceAccount` does not have all the necessary permissions for the Adapter to work, especially the `monitoring.metricDescriptors.get` permission.
This caused me to get this error when I updated the service account permissions.

image

Therefore, the objective of this change is to revert at least two roles that were removed (roles/monitoring.metricWriter and roles/stackdriver.resourceMetadata.writer) so that the GMP Adapter works normally again.


apeabody (changes requested):
Thanks for the contribution @jullianow!

From the Integration Test:
```
Error: Reference to undeclared resource

  on ../../../sa.tf line 57, in resource "google_project_iam_member" "cluster_service_account-metric_writer":
  57:   project = google_project_iam_member.cluster_service_account-log_writer[0].project

A managed resource "google_project_iam_member"
"cluster_service_account-log_writer" has not been declared in
module.example.module.gke.

Error: Reference to undeclared resource

  on ../../../sa.tf line 64, in resource "google_project_iam_member" "cluster_service_account-resourceMetadata-writer":
  64:   project = google_project_iam_member.cluster_service_account-monitoring_viewer[0].project

A managed resource "google_project_iam_member"
"cluster_service_account-monitoring_viewer" has not been declared in
module.example.module.gke.}
```

jullianow:
@apeabody Adjusted.

apeabody (approved):
Thanks for the contribution @jullianow!

Confirmed description of these roles here: https://cloud.google.com/iam/docs/understanding-roles

oliverboehme-ida:
Hm I just saw this change when trying to upgrade to `v31.1.0`.

Wouldn't it have been sufficient for solving your problem to just assign `roles/monitoring.viewer` (a read-only role).
This would also be a less invasive way in my eyes.

Also I don't see at all why it needs `roles/stackdriver.resourceMetadata.writer`.

PR #1951, opened 2024-05-23
fix: dynamic block ephemeral_storage_local_ssd_config

alvicsam:
Fixes changes implemented in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1944
During renaming of variable I forgot to change second variable name in dynamic block. This PR fixes it.

cc @apeabody


apeabody (approved):
Thanks for the contribution @alvicsam!

PR #1962, opened 2024-05-31
feat: add support for `logging_variant` in Autopilot clusters

abdel:
Leverage the existing `logging_variant` [[ref](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#logging_variant)] option at a cluster-level in `google_container_cluster` module to (optionally) allow enabling `MAX_THROUGHPUT` logging in Autopilot clusters.


apeabody (changes requested):
Thanks for the contribution @abdel!

Can you please make your change in the source template as described here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating. Thanks!

abdel:
@apeabody Sorry for the late response, didn't get a notification for some reason! I made the changes to the `autogen` templates, and did the re-build for the modules and the docs.

Let me know if that looks OK!

autogen/main/variables.tf.tmpl
```diff
@@ -889,3 +889,11 @@ variable "fleet_project_grant_service_agent" {
   default = false
 }
 {% endif %}
+{% if beta_cluster and autopilot_cluster %}
+
+variable "logging_variant" {
+  description = "(Optional) The type of logging agent that is deployed by default for newly created node pools in the cluster. Valid values include DEFAULT and MAX_THROUGHPUT."
+  type        = string
+  default     = "DEFAULT"
```
apeabody, line 897: As this will override the cluster's value, we should probably default to `null`.  This will also avoid config churn on initial apply.

apeabody (changes requested):
Thanks for the contribution @abdel!

autogen/main/variables.tf.tmpl
abdel: Good point -- I have changed the default to `null`

apeabody (approved):
Thanks for the contribution @abdel!

PR #1969, opened 2024-06-06
feat: Add enable_l4_ilb_subsetting for autopilot clusters

kliu47:
setting already exists for other submodules. Adding functionality to autopilot submodule


apeabody (changes requested):
Thanks for the contribution @kliu47!

Can we confirm this is supported in Autopilot? `GKE subsetting is disabled in Autopilot clusters.`
https://cloud.google.com/kubernetes-engine/docs/how-to/internal-load-balancing#subsetting

At the very least we should enable in at least one autopilot example so it's exercised by the CI tests: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/examples/simple_autopilot_public/main.tf#L35

kliu47:
@apeabody , Yes, this feature is supported in autopilot clusters, however the default behavior is to have it disabled
image


I've also gone ahead and added the setting to true in the public autopilot example:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1969/files#diff-229618ff3e7246fb9ddf90bc00eea2b8f399b4412e8d42d7783d2ecf04bece48R51

apeabody (approved):
Thanks for the contribution @kliu47!

PR #1142, opened 2022-02-02
Add Identity Service config to cluster

stanley98yu:
Fixes #1137

PR #1982, opened 2024-06-21
fix: Fix the value of output "identity_service_enabled" in beta modules

legal90:
It should return the value of "Identity Service" feature flag (`local.cluster_identity_service_enabled`), not "Pod Security Policy"(`local.cluster_pod_security_policy_enabled`)

It seems that there was a typo in the initial implementation of this feature in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1142.

Note: the local variable `local.cluster_output_identity_service_enabled ` was already defined earlier, but it was deleted in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1598 because it wasn't used.


apeabody (approved):

PR #1989, opened 2024-07-04
change: ip_range_services to optional value (#1949)

### LINKED ISSUE #1949: Make service range optional
### TL;DR

There is a new default service IP range which should be used for new 1.29.3+ clusters:

GKE versions up to 1.29.2 will create a random /20 range from 10.0.0.0/8 if ip_range_services is omitted. From 1.29.3 this will set the new default service range 34.118.224.0/20 on each cluster. ( See https://cloud.google.com/kubernetes-engine/docs/release-notes#May_10_2024 )

Currently it is possible to use the new default service range by explicitly setting `var.ip_range_services` to `null`.
This works for our use case, but I was wondering if you wanted this to be the default for the module in the future. I'm happy to provide a PR for this, but wanted to make sure it was something that was wanted. 

### Terraform Resources

```markdown
# something like this:
 
 variable "ip_range_services" {
   type        = string
-  description = "The _name_ of the secondary subnet range to use for services"
+  description = "The _name_ of the secondary subnet range to use for services. Omit to use default range."
+  default     = null
 }
```

### Detailed design

_No response_

### Additional information

_No response_

0Delta:
resolve #1949 

As of GKE version 1.29 and Autopilot 1.27, the service ip range is given a default of 34.118.224.0/20 per cluster.
Versions earlier than the specified version may be omitted, but will be rejected by the validator.

This change has a big impact, but has not been thoroughly discussed yet.
We hope to merge it after sufficient discussion and making the necessary corrections.

### DIFF
--- a/autogen/main/variables.tf.tmpl
+++ b/autogen/main/variables.tf.tmpl
@@ -147,7 +147,8 @@ variable "additional_ip_range_pods" {
 
 variable "ip_range_services" {
   type        = string
-  description = "The _name_ of the secondary subnet range to use for services"
+  description = "The _name_ of the secondary subnet range to use for services. Omit to use default range."
+  default     = null
 }
 
 variable "stack_type" {
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ Then perform the following commands on the root folder:
 | ip\_masq\_link\_local | Whether to masquerade traffic to the link-local prefix (169.254.0.0/16). | `bool` | `false` | no |
 | ip\_masq\_resync\_interval | The interval at which the agent attempts to sync its ConfigMap file from the disk. | `string` | `"60s"` | no |
 | ip\_range\_pods | The _name_ of the secondary subnet ip range to use for pods | `string` | n/a | yes |
-| ip\_range\_services | The _name_ of the secondary subnet range to use for services | `string` | n/a | yes |
+| ip\_range\_services | The _name_ of the secondary subnet range to use for services. Omit to use default range. | `string` | `null` | no |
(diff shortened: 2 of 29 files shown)


0Delta:
Hi @apeabody , sorry for the sudden mention.
I'd like to gather opinions from others about this change. 
Do you know of a good place to ask questions or users who can give their opinions?

apeabody:
> Hi @apeabody , sorry for the sudden mention. I'd like to gather opinions from others about this change. Do you know of a good place to ask questions or users who can give their opinions?

Hi @0Delta! - Generally the related issue (e.g. https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1949) or if implementation specific the PR itself are good options.  However this PR is currently marked as `draft`, so won't be reviewed and less likely to get comments.

0Delta:
Thanks for the replies @apeabody.
This issue is now mostly a discussion about what to change, not the code implementation.
I'd like someone to provide advice on the original PR - do you know who would be the best person to do so?

autogen/main/variables.tf.tmpl
```diff
@@ -147,7 +147,8 @@ variable "additional_ip_range_pods" {
 
 variable "ip_range_services" {
   type        = string
-  description = "The _name_ of the secondary subnet range to use for services"
+  description = "The _name_ of the secondary subnet range to use for services. Omit to use default range."
```
apeabody, line 150: Can we clarify the new section to something similar to: If not provided the default `34.118.224.0/20` range will be used.

apeabody:
> Thanks for the replies @apeabody. This issue is now mostly a discussion about what to change, not the code implementation. I'd like someone to provide advice on the original PR - do you know who would be the best person to do so?

Thanks @0Delta - Given the scope of the change, we'll need to include full test coverage.  This can be done by updating at least one standard cluster and one autopilot cluster example to omit `ip_range_services`: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/examples, and then validate the actual cluster in the associated example's test verify: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/tree/master/test/integration

PR #2011, opened 2024-07-26
feat: support workload_identity_config on autopilot

DrFaust92:
supported via https://github.com/hashicorp/terraform-provider-google/pull/18166


apeabody (changes requested):

I believe this will require bumping the minimum provider to `[5.31](https://github.com/hashicorp/terraform-provider-google/releases/tag/v5.31.0)`

DrFaust92:
[apeabody](https://github.com/apeabody) oops, saw the ref for beta for 5.33+ and missed the GA one, bumped GA as well

apeabody:
> [apeabody](https://github.com/apeabody) oops, saw the ref for beta for 5.33+ and missed the GA one, bumped GA as well

Thanks @DrFaust92 - Actually looks like this is only being added to the `beta` autopilot clusters?  Probably want to enable to the GA variants as well?

DrFaust92:
[apeabody](https://github.com/apeabody) i dont see a non beta auto pilot dir at all. is that even an option?
Screenshot 2024-07-30 at 23 12 05

apeabody:
> [apeabody](https://github.com/apeabody) i dont see a non beta auto pilot dir at all. is that even an option? Screenshot 2024-07-30 at 23 12 05

You are correct @DrFaust92, I saw the updates to the over module's `version.tf`, but didn't check that they weren't Autopilot clusters.  If we don't have a non-beta Autopilot clusters, we can actually revert https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2011/files#diff-5ccc4bcabc51d0aa62a45f4a5cfe9868fb5f29fc613afa1f5a16e5c263b16d99L49, and then this should be ready to commit.  Thanks!

autogen/main/versions.tf.tmpl
```diff
@@ -46,7 +46,7 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 5.25.0, < 6"
+      version = ">= 5.31.0, < 6"
```
apeabody, line 49: @DrFaust92 - Let's revert and build, and then this should be ready to merge.

apeabody (approved):

PR #2012, opened 2024-07-26
feat: support enable_nested_virtualization

DrFaust92:


DrFaust92:
[apeabody](https://github.com/apeabody) tried to add a test, hope this is well

autogen/main/cluster.tf.tmpl
```diff
@@ -992,9 +992,10 @@ resource "google_container_node_pool" "windows_pools" {
     }
 
     dynamic "advanced_machine_features" {
-      for_each = lookup(each.value, "threads_per_core", 0) > 0 ? [1] : []
+      for_each = lookup(each.value, "threads_per_core", 0) > 0 || lookup(each.value, "enable_nested_virtualization", false) ? [1] : []
       content {
-        threads_per_core = lookup(each.value, "threads_per_core", 0)
+        threads_per_core             = lookup(each.value, "threads_per_core", 0)
```
apeabody, line 997: @DrFaust92 Given this block will now be created for either condition, I think we might need this to default to null rather than 0?

```suggestion
        threads_per_core             = lookup(each.value, "threads_per_core", null)
```

DrFaust92:
Hi [apeabody](https://github.com/apeabody), can you share the error?

apeabody:
> Hi [apeabody](https://github.com/apeabody), can you share the error?

Hmm, unfortunately it didn't accept the `null` for `threads_per_core`.  Perhaps the origin `"0"` or `""` are considered "unset"?  I'll need to investigate further and perhaps get the documentation updated.


"The number of threads per physical core. To disable simultaneous multithreading (SMT) set this to 1. If unset, the maximum number of threads supported per core by the underlying processor is assumed."


```
Step #67 - "converge node-pool-local": STDERR: Error: Missing required argument
Step #67 - "converge node-pool-local": 
Step #67 - "converge node-pool-local":   with module.example.module.gke.google_container_node_pool.pools["pool-05"],
Step #67 - "converge node-pool-local":   on ../../../modules/beta-public-cluster/cluster.tf line 637, in resource "google_container_node_pool" "pools":
Step #67 - "converge node-pool-local":  637:   node_config {
Step #67 - "converge node-pool-local": 
Step #67 - "converge node-pool-local": The argument "node_config.0.advanced_machine_features.0.threads_per_core" is
Step #67 - "converge node-pool-local": required, but no definition was found.
```

DrFaust92:
[apeabody](https://github.com/apeabody) probably zero us unset as its an int and in golang many times its as good as unset

apeabody:
> [apeabody](https://github.com/apeabody) probably zero us unset as its an int and in golang many times its as good as unset

Agreed, `"0"` is my educated guess as well.  We can go forward with that value for now if you like.

apeabody (approved):

PR #2003, opened 2024-07-21
feat: add support for confidential storage + docs fixes

DrFaust92:

PR #2018, opened 2024-07-31
fix: enable_confidential_storage fix

DrFaust92:
fix copy paste mistake from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2003


apeabody (approved):

PR #2033, opened 2024-08-08
feat: add in sole tenant selectors

aaahrens:
The goal of this PR is to add sole_tenant support in the node_config of private nodes, similar to how taints current work. It works additive only, and out add the config if someone inputted something like below:

```
...
node_pools_sole_tenant_selectors = {
    "my-node-pool" = {
      key = "compute.googleapis.com/node-group-name"
      operator = "IN"
      values = ["sole-pool-host"]
    }
  }
...
```


apeabody (changes requested):
Thanks for the contribution @aaahrens!

Please update the template with your change as detailed here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

PR #2040, opened 2024-08-15
fix: remove duplicte enable_secure_boot

DrFaust92:
should be cosmetic only


apeabody (approved):

PR #2052, opened 2024-08-22
fix(autopilot): option to configure gcfs

apeabody:
This currently plumbs `enable_gcfs` with a default of `null`, however in a future breaking change it will be `true`.

### DIFF
--- a/modules/beta-autopilot-private-cluster/variables.tf
+++ b/modules/beta-autopilot-private-cluster/variables.tf
@@ -524,6 +524,12 @@ variable "enable_l4_ilb_subsetting" {
   default     = false
 }
 
+variable "enable_gcfs" {
+  type        = bool
+  description = "(Beta) Enable image streaming on cluster level."
+  default     = null
+}
+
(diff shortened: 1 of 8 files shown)


modules/beta-autopilot-private-cluster/variables.tf
```diff
@@ -524,6 +524,12 @@ variable "enable_l4_ilb_subsetting" {
   default     = false
 }
 
+variable "enable_gcfs" {
+  type        = bool
+  description = "(Beta) Enable image streaming on cluster level."
+  default     = null
```
bharathkkb, line 530: With this approach do we still need ignore_changes or can we revert that?

bharathkkb (approved):

modules/beta-autopilot-private-cluster/variables.tf
apeabody: I think it should be fine to revert it now as it doesn't address the desired issue anything.

PR #2074, opened 2024-09-09
fix(TPG>=5.41)!: add config_sync.enabled

apeabody:
- Add `config_sync.enabled` introduced in TPG v5.41

### DIFF
--- a/modules/acm/feature.tf
+++ b/modules/acm/feature.tf
@@ -42,6 +42,7 @@ resource "google_gke_hub_feature_membership" "main" {
       for_each = var.enable_config_sync ? [{ enabled = true }] : []
 
       content {
+        enabled       = var.enable_config_sync
         source_format = var.source_format != "" ? var.source_format : null
 
         git {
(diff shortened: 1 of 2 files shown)


bharathkkb (approved):

modules/acm/feature.tf
```diff
@@ -42,6 +42,7 @@ resource "google_gke_hub_feature_membership" "main" {
       for_each = var.enable_config_sync ? [{ enabled = true }] : []
 
       content {
+        enabled       = var.enable_config_sync
```
haiyanmeng, line 45: Do we still need line 42 `for_each = var.enable_config_sync ? [{ enabled = true }] : []`?

apeabody: Yes, as this is currently a dynamic block and the block content only populated based on the for_each.  However the actual value of `[{ enabled = true }]` isn't important other than it evaluates to length `1`.

haiyanmeng (approved):

janetkuo (approved):

apeabody:
Depends on https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2076

PR #2082, opened 2024-09-11
feat(TPG>=5.44.2)!: add standard cluster support for `insecureKubeletReadonlyPortEnabled`

### LINKED ISSUE #2013: insecureKubeletReadonlyPortEnabled support
### TL;DR

Is the `insecureKubeletReadonlyPortEnabled` flag in the kubelet config supported? With the recent announcements
https://cloud.google.com/kubernetes-engine/docs/how-to/disable-kubelet-readonly-port#migrate-apps, I updated the cluster configs.

Probably depends on https://github.com/hashicorp/terraform-provider-google/issues/15208

After running:
```
% gcloud container clusters update xxx \
    --location=us-central1 --project xxx \
    --no-enable-insecure-kubelet-readonly-port
% gcloud container node-pools update primary \     
    --cluster=cluster --project xxx \
    --location=us-central1 \                  
    --no-enable-insecure-kubelet-readonly-port
```

I seem to be getting some unresolvable diffs with the kubelet config. I might be just running into something related to https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1922 but I believe I did check the configs after updating to the latest module version, and I didn't have any luck just by setting `pod_pids_limit` and `cpu_cfs_quota` in the pod config.

diffs:
```
          - kubelet_config {
              - cpu_cfs_quota  = false -> null
              - pod_pids_limit = 0 -> null
            }
```

I can see this in the debug logs
```
2024-07-26T11:40:22.622-0700 [DEBUG] provider.terraform-provider-google:   "kubeletConfig": {
2024-07-26T11:40:22.622-0700 [DEBUG] provider.terraform-provider-google:    "insecureKubeletReadonlyPortEnabled": false
2024-07-26T11:40:22.622-0700 [DEBUG] provider.terraform-provider-google:   },
```

### Terraform Resources

_No response_

### Detailed design

_No response_

### Additional information

_No response_

wyardley:
This adds support for `insecure_kubelet_readonly_port_enabled`, now that it's supported at the provider level.
This will need either > 5.44.2 (cherrypicked update) or > 6.7.0, so will be a breaking change; happy to tweak the title and / or commit type as desired.

Fixes #2013


autogen/main/versions.tf.tmpl
```diff
@@ -24,11 +24,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 5.40.0, < 7"
+      version = ">= 5.44.0, < 7, !=6.0.0, !=6.0.1, !=6.1.0"
```
wyardley, line 27: Let me know if there's a better way to specify > 5.44.0 or > 6.2.0.

apeabody: Unfortunately looking at the underlying library it appears the constraints are AND, there isn't a way to group/OR them.

https://github.com/hashicorp/go-version

wyardley:
Thanks for triggering @apeabody 
Would you mind letting me know what the failed tests are? If it gets super complicated, I can get the test setup working locally, but if it's something simple, I can try to fix.

wyardley:
Thanks, yeah, passed that time. Any other adjustments you all want? presumably this will have to wait for another major release?

wyardley:
Not sure if there's anything special that needs to be done to support the situation where the default changes down the road. I think leaving the default `null` should already handle that Ok?
When I applied 33.x version module without this change and 6.2.0 provider, it behaved somewhat as I'd expect, reflecting the new setting in the state, but not triggering any changes in the plan.

apeabody:
> Thanks, yeah, passed that time. Any other adjustments you all want? presumably this will have to wait for another major release?

Correct - This is a breaking change, so will need to wait till we start assembling the next major version.

apeabody:
> Not sure if there's anything special that needs to be done to support the situation where the default changes down the road. I think leaving the default `null` should already handle that Ok? When I applied 33.x version module without this change and 6.2.0 provider, it behaved somewhat as I'd expect, reflecting the new setting in the state, but not triggering any changes in the plan.

It's hard to predict, but as this is a field (and not a block), leaving as `null` is probably a good option.

wyardley (commented):
@apeabody I realized this probably needs to also be tested in a couple more spots (and probably should use `"FALSE"` as the default, since even though `"TRUE"` may be a safer bet, we probably don't want to encourage people to set that by putting it in examples) , and probably setup for autopilot clusters too? Does it make sense to use a single parameter to handle
* `node_pool_auto_config.kubelet_config.insecure_kubelet_readonly_port_enabled` for autopilot clusters (currently a bit stuck on how to handle that block)
* `node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` (I think `node_pool_defaults.node_config_defaults` needs to have its beta gate removed as well?
* node pool settings (`node_config.kubelet_config`) at the node-pool level

apeabody:
> @apeabody I realized this probably needs to also be tested in a couple more spots (and probably should use `"FALSE"` as the default, since even though `"TRUE"` may be a safer bet, we probably don't want to encourage people to set that by putting it in examples) , and probably setup for autopilot clusters too? Does it make sense to use a single parameter to handle
> 
> * `node_pool_auto_config.kubelet_config.insecure_kubelet_readonly_port_enabled` for autopilot clusters (currently a bit stuck on how to handle that block)
> * `node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` (I think `node_pool_defaults.node_config_defaults` needs to have its beta gate removed as well?
> * node pool settings (`node_config.kubelet_config`) at the node-pool level
> 
> I guess maybe we need to differentiate between the setting for a given node pool vs. the autopilot or `node_pool_defaults` cases, where there'd just be a parameter set at the top level, so maybe don't try to solve that problem as part of this PR?
> 
> Does it make sense to have the same parameter name (`insecure_kubelet_readonly_port_enabled`) as a module variable and as a value within the `node_pools` var as well?

I haven't looked into these specifically, but generally we want to retain the ability for a separate cluster vs nodepool configuration - same parameter name is fine.  Looking into `node_pool_defaults.node_config_defaults`, I would suggest it be best to do an separate/initial PR to promote that from beta for clarity of the change.  Fortunately `node_config_defaults` is already GA in `5.40`, so it can be a normal `Feat:` PR. :)

wyardley:
> but generally we want to retain the ability for a separate cluster vs nodepool configuration - same parameter name is fine

Ok, I may convert this to a draft, but just pushed up something for you to look at here.

While it would be nice to have it be a true bool, I'm thinking match the upstream behavior of `"TRUE"` / `"FALSE"` (look at that PR if you want more detail about why it got implemented that way), which also lets us have it be optional (vs. a boolean?)

apeabody:
> > but generally we want to retain the ability for a separate cluster vs nodepool configuration - same parameter name is fine
> 
> Ok, I may convert this to a draft, but just pushed up something for you to look at here.
> 
> While it would be nice to have it be a true bool, I'm thinking match the upstream behavior of `"TRUE"` / `"FALSE"` (look at that PR if you want more detail about why it got implemented that way), which also lets us have it be optional (vs. a boolean?)

I'll take a look next week, but in Terraform bools can be `null` as well (behave as if omitted), which should probably work for you use case?

wyardley:
@apeabody Per discussion [here](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2098#issuecomment-2353532876), just mentioning that I got this communication today:

> Due to this security issue, we will disable port 10255 in the following phases:

>    - GKE version 1.32 and higher: Port 10255 will be disabled by default on new clusters. You can still enable it, if necessary, but it is not recommended.
>      Note: There is no change to port 10255 on existing clusters, even on clusters upgraded to 1.32.
>    - Future GKE versions: Port 10255 will be completely disabled with no option to enable it.

autogen/main/cluster.tf.tmpl
```diff
@@ -677,12 +680,12 @@ resource "google_container_cluster" "primary" {
       gcfs_config {
```
apeabody, line 680: I believe that `gcfs_config` is still beta, so will need to still gate this block.

wyardley: Yeah was going to rebase after #2098 originally, but now they may need to go together?

apeabody: My quick read is that even in [v6.3](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster) while `node_config_defaults` is GA, `node_config_defaults.gcfs_config` is still beta?

wyardley: correct. The narrower fix I ended up with there still keeps `gcfs_config` gated (however breaks because then `node_config_defaults` is empty, which I think will be a potential issue either way unless we do the dynamic block?)

wyardley:
@apeabody another issue here w/r/t autopilot clusters:

autpilot doesn't allow 5.44, and 5.44 is the min version that will support this. It also doesn't support 6.2 either (#2076 and #2086). So unless some additional fixes get cherrypicked to 5.x, this may have to be breaking for >= 6.N where N is some future release where some of these issues are fixed.

I guess we could potentially revert the autopilot related examples and code if you want to have some stab at the non-autopilot stuff working with some 5.x version, and then do a different breaking change for autopilot, but this one is already breaking, so maybe go big or go home?

apeabody:
> @apeabody another issue here w/r/t autopilot clusters:
> 
> autpilot doesn't allow 5.44, and 5.44 is the min version that will support this. It also doesn't support 6.2 either (#2076 and #2086). So unless some additional fixes get cherrypicked to 5.x, this may have to be breaking for >= 6.N where N is some future release where some of these issues are fixed.
> 
> I guess we could potentially revert the autopilot related examples and code if you want to have some stab at the non-autopilot stuff working with some 5.x version, and then do a different breaking change for autopilot, but this one is already breaking, so maybe go big or go home?

Yeah - I'm open to requiring v6, however https://github.com/hashicorp/terraform-provider-google/issues/19428 will need to be resolved.  Otherwise it might be possible to workaround the conflict in our Autopilot sub-modules (perhaps more dynamic blocks, etc), but that might not be ultimately possible and retain the current options.

wyardley:
@apeabody looks like it'll go in 5.44.1: https://github.com/hashicorp/terraform-provider-google/commits/release-5.44.1

wyardley:
@apeabody 6664523d1c902f13a725b6dbef614605cbe03b03 isn't intended to work yet, and may have some mistakes, but just for you to take a look at.
You could try running it after 5.44.1 comes out maybe. Not sure if autopilot needs 6.3.0 excluded as well.

apeabody:
> @apeabody [6664523](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/commit/6664523d1c902f13a725b6dbef614605cbe03b03) isn't intended to work yet, and may have some mistakes, but just for you to take a look at. You could try running it after 5.44.1 comes out maybe. Not sure if autopilot needs 6.3.0 excluded as well.

Yeah, waiting to try Autopilot with v5.44.1 and v6.4.0.

I can run the tests for this change once v5.44.1 is published, as the constraint isn't currently possible to satisfy.

wyardley:
6.4.0 and 5.44.1 are out now.
Running the lint checks locally now.

wyardley:
```
terraform_validate ./examples/simple_autopilot_private
╷
│ Error: Unsupported argument
│ 
│   on ../../modules/beta-autopilot-private-cluster/cluster.tf line 125, in resource "google_container_cluster" "primary":
│  125:         insecure_kubelet_readonly_port_enabled = var.insecure_kubelet_readonly_port_enabled
```

Seeing this in the CI lint error - I think this is the issue we already knew with autopilot support fixes not getting cherry-picked back to 5.x?

I could drop the autopilot stuff in here to cut a non-breaking release with regular support and then add that later in a separate PR, to make the changelog and versioning easier, or we could leave the version constraints as-is but increase the required version to 6.4 for autopilot (or some other option if you'd prefer?)

autogen/main/versions.tf.tmpl
```diff
@@ -46,13 +46,11 @@ terraform {
   required_providers {
     google = {
       source = "hashicorp/google"
-      # Workaround for https://github.com/hashicorp/terraform-provider-google/issues/19428
-      version = ">= 5.40.0, != 5.44.0, < 6.2.0, < 7"
+      version = ">= 5.44.1, < 7, !=6.0.0, !=6.0.1, !=6.1.0, !=6.2.0, !=6.3.0"
```
wyardley, line 49: could update these two to just be `> 6.4.0, <7` and then I _think_ this should work (but at the cost of making this (more) breaking)

apeabody:
> ```
> terraform_validate ./examples/simple_autopilot_private
> ╷
> │ Error: Unsupported argument
> │ 
> │   on ../../modules/beta-autopilot-private-cluster/cluster.tf line 125, in resource "google_container_cluster" "primary":
> │  125:         insecure_kubelet_readonly_port_enabled = var.insecure_kubelet_readonly_port_enabled
> ```
> 
> Seeing this in the CI lint error - I think this is the issue we already knew with autopilot support fixes not getting cherry-picked back to 5.x?
> 
> I could drop the autopilot stuff in here to cut a non-breaking release with regular support and then add that later in a separate PR, to make the changelog and versioning easier, or we could leave the version constraints as-is but increase the required version to 6.4 for autopilot (or some other option if you'd prefer?)

Hey @wyardley!

I'm thinking let's split out the Autopilot for now, especially as it has it's own set of challenges with some of the provider versions.

I was hoping to not require TPG v6 as the minimum till after the next breaking release, but it's always an option.

wyardley:
> I'm thinking let's split out the Autopilot for now,

Can do.
Do you think I need to call out in the parameter docs that it's not yet supported for autopilot?

wyardley:
Also, do you prefer atomic commits? I'm leaving it that way for now for easier review, but let me know if you want me to squash at some point.

apeabody:
> > I'm thinking let's split out the Autopilot for now,
> 
> Can do. Do you think I need to call out in the parameter docs that it's not yet supported for autopilot?

It should be self documenting, with the parameter simply not included in the autopilot modules.  If you are thinking for a PR title, you can just say something like standard clusters.

apeabody:
> Also, do you prefer atomic commits? I'm leaving it that way for now for easier review, but let me know if you want me to squash at some point.

Whatever you prefer, we squash all commits on final merge.

autogen/main/versions.tf.tmpl
```diff
@@ -89,7 +87,7 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 5.40.0, < 7"
+      version = ">= 5.44.1, < 7, !=6.0.0, !=6.0.1, !=6.1.0, !=6.2.0, !=6.3.0"
```
wyardley, line 90: had forgotten to update this one.

LMK if you want to leave 6.3.0 or any other 6.x versions allowed; I think with the GCFS issue, this is best?

wyardley:
Sorry, had to fix the two CI linting issues (and running into some issues locally with lint -- maybe due to parallel being enabled and / or platform specific stuff?), but passing now, at least, if you want to run again.

wyardley:
@apeabody was that a legit failure that time, or just needs to be re-run?

apeabody:
> @apeabody was that a legit failure that time, or just needs to be re-run?

Let's try a re-run.

wyardley:
Rebased and resolved conflicts.

autogen/main/cluster.tf.tmpl
```diff
@@ -674,15 +673,17 @@ resource "google_container_cluster" "primary" {
       }
       {% endif %}
       {% if autopilot_cluster != true %}
+      {% if beta_cluster %}
       gcfs_config {
         enabled = var.enable_gcfs
       }
       {% endif %}
+      insecure_kubelet_readonly_port_enabled = var.insecure_kubelet_readonly_port_enabled != "" ? var.insecure_kubelet_readonly_port_enabled : null
```
apeabody, line 681: Looks like `insecure_kubelet_readonly_port_enabled` is a bool, not string.  Should this be `!= null`?

autogen/main/cluster.tf.tmpl
```diff
@@ -1040,14 +1041,15 @@ resource "google_container_node_pool" "windows_pools" {
...
 
       content {
-        cpu_manager_policy   = lookup(each.value, "cpu_manager_policy", "static")
-        cpu_cfs_quota        = lookup(each.value, "cpu_cfs_quota", null)
-        cpu_cfs_quota_period = lookup(each.value, "cpu_cfs_quota_period", null)
-        pod_pids_limit       = lookup(each.value, "pod_pids_limit", null)
+        cpu_manager_policy                     = lookup(each.value, "cpu_manager_policy", "static")
+        cpu_cfs_quota                          = lookup(each.value, "cpu_cfs_quota", null)
+        cpu_cfs_quota_period                   = lookup(each.value, "cpu_cfs_quota_period", null)
+        insecure_kubelet_readonly_port_enabled = lookup(each.value, "insecure_kubelet_readonly_port_enabled", var.insecure_kubelet_readonly_port_enabled != "" ? var.insecure_kubelet_readonly_port_enabled : null)
```
apeabody, line 1051: same here?

```suggestion
        insecure_kubelet_readonly_port_enabled = lookup(each.value, "insecure_kubelet_readonly_port_enabled", var.insecure_kubelet_readonly_port_enabled != null ? var.insecure_kubelet_readonly_port_enabled : null)
```

autogen/main/cluster.tf.tmpl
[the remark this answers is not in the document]
apeabody: Actually I see now the cluster `insecure_kubelet_readonly_port_enabled` is a string, but the nodepool `insecure_kubelet_readonly_port_enabled` is a bool.

autogen/main/variables.tf.tmpl
```diff
@@ -103,6 +103,17 @@ variable "service_external_ips" {
 }
 
 {% if autopilot_cluster != true %}
+variable "insecure_kubelet_readonly_port_enabled" {
+  type = string
```
apeabody, line 107: I might have already asked, but can we use a `bool` with a default of `null` for this?

wyardley: Yes, I think that's a good idea (for this module). Will make that update.

autogen/main/cluster.tf.tmpl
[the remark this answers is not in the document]
wyardley: At the provider level, it's all implemented as a string enum `"TRUE"` / `"FALSE"` (for [Reasons](https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#discussion_r1714500935) -- see discussion starting there if you want more details).

But within the module, I think it makes sense to implement it as an optional bool as you've suggested.

wyardley:
@apeabody also, do we need to add anything into `safer_cluster` for this?

apeabody:
> @apeabody also, do we need to add anything into `safer_cluster` for this?

Only if you would like, it can always be added in a later PR.  And with the `null` default, should automatically follow the API default.

wyardley:
> Only if you would like, it can always be added in a later PR. And with the null default, should automatically follow the API default.

In the short term, a "safer" cluster would probably force it to be false, but OTOH, the default seems planned to be changed in the near future to the secure default anyway.

begarco:
Is there any plan to merge this in master ? Or is there something which is mising ? Maybe we can help ?

prune998:
> @begarco i believe they’ll wait to send this out as a breaking release, as it will drop support for some versions of the provider that are currently supported. Other than that, I think it’s ready to go.

In fact, it's all broken already even without this change !

I opened https://github.com/hashicorp/terraform-provider-google/issues/19792 to demonstrate how the provider badly handle data from Google API. 
The outcome right now is that I just can't create/update new nodepools (olders are OK as they don't have the `kube-config` values set).

Unless if i'm mistaken, this needs to go out ASAP

wyardley:
@prune998 I realize this is probably not ideal, but a couple things you could try in the short term:
1) Since the nodepool can be created initially, use the CLI to set the value to false manually on the nodepool and see if this resolves the issue with an unapplyable plan
2) Try setting `node_config.kubelet_config.cpu_manager_policy` to `""` (not sure if that will solve the issue, or still cause the same issue because it will essentially set an empty value)

apeabody:
FYI @wyardley

```
Step #101 - "verify private-zonal-with-networking":         	Error Trace:	/workspace/test/integration/private_zonal_with_networking/private_zonal_with_networking_test.go:81
Step #101 - "verify private-zonal-with-networking":         	            				/builder/home/go/pkg/mod/github.com/!google!cloud!platform/cloud-foundation-toolkit/infra/blueprint-test@v0.16.2/pkg/tft/terraform.go:638
Step #101 - "verify private-zonal-with-networking":         	            				/builder/home/go/pkg/mod/github.com/!google!cloud!platform/cloud-foundation-toolkit/infra/blueprint-test@v0.16.2/pkg/tft/terraform.go:670
Step #101 - "verify private-zonal-with-networking":         	            				/builder/home/go/pkg/mod/github.com/!google!cloud!platform/cloud-foundation-toolkit/infra/blueprint-test@v0.16.2/pkg/utils/stages.go:31
Step #101 - "verify private-zonal-with-networking":         	            				/builder/home/go/pkg/mod/github.com/!google!cloud!platform/cloud-foundation-toolkit/infra/blueprint-test@v0.16.2/pkg/tft/terraform.go:670
Step #101 - "verify private-zonal-with-networking":         	Error:      	Not equal: 
Step #101 - "verify private-zonal-with-networking":         	            	expected: map[string]interface {}{"diskSizeGb":100, "diskType":"pd-standard", "effectiveCgroupMode":"EFFECTIVE_CGROUP_MODE_V2", "imageType":"COS_CONTAINERD", "labels":map[string]interface {}{"cluster_name":"CLUSTER_NAME", "node_pool":"default-node-pool"}, "loggingConfig":map[string]interface {}{"variantConfig":map[string]interface {}{"variant":"DEFAULT"}}, "machineType":"e2-medium", "metadata":map[string]interface {}{"cluster_name":"CLUSTER_NAME", "disable-legacy-endpoints":"true", "node_pool":"default-node-pool"}, "oauthScopes":[]interface {}{"https://www.googleapis.com/auth/cloud-platform"}, "serviceAccount":"SERVICE_ACCOUNT", "shieldedInstanceConfig":map[string]interface {}{"enableIntegrityMonitoring":true}, "tags":[]interface {}{"gke-CLUSTER_NAME", "gke-CLUSTER_NAME-default-node-pool"}, "windowsNodeConfig":map[string]interface {}{}, "workloadMetadataConfig":map[string]interface {}{"mode":"GKE_METADATA"}}
Step #101 - "verify private-zonal-with-networking":         	            	actual  : map[string]interface {}{"diskSizeGb":100, "diskType":"pd-standard", "effectiveCgroupMode":"EFFECTIVE_CGROUP_MODE_V2", "imageType":"COS_CONTAINERD", "kubeletConfig":map[string]interface {}{"insecureKubeletReadonlyPortEnabled":true}, "labels":map[string]interface {}{"cluster_name":"CLUSTER_NAME", "node_pool":"default-node-pool"}, "loggingConfig":map[string]interface {}{"variantConfig":map[string]interface {}{"variant":"DEFAULT"}}, "machineType":"e2-medium", "metadata":map[string]interface {}{"cluster_name":"CLUSTER_NAME", "disable-legacy-endpoints":"true", "node_pool":"default-node-pool"}, "oauthScopes":[]interface {}{"https://www.googleapis.com/auth/cloud-platform"}, "serviceAccount":"SERVICE_ACCOUNT", "shieldedInstanceConfig":map[string]interface {}{"enableIntegrityMonitoring":true}, "tags":[]interface {}{"gke-CLUSTER_NAME", "gke-CLUSTER_NAME-default-node-pool"}, "windowsNodeConfig":map[string]interface {}{}, "workloadMetadataConfig":map[string]interface {}{"mode":"GKE_METADATA"}}
Step #101 - "verify private-zonal-with-networking":         	            	
Step #101 - "verify private-zonal-with-networking":         	            	Diff:
Step #101 - "verify private-zonal-with-networking":         	            	--- Expected
Step #101 - "verify private-zonal-with-networking":         	            	+++ Actual
Step #101 - "verify private-zonal-with-networking":         	            	@@ -1,2 +1,2 @@
Step #101 - "verify private-zonal-with-networking":         	            	-(map[string]interface {}) (len=14) {
Step #101 - "verify private-zonal-with-networking":         	            	+(map[string]interface {}) (len=15) {
Step #101 - "verify private-zonal-with-networking":         	            	  (string) (len=10) "diskSizeGb": (float64) 100,
Step #101 - "verify private-zonal-with-networking":         	            	@@ -5,2 +5,5 @@
Step #101 - "verify private-zonal-with-networking":         	            	  (string) (len=9) "imageType": (string) (len=14) "COS_CONTAINERD",
Step #101 - "verify private-zonal-with-networking":         	            	+ (string) (len=13) "kubeletConfig": (map[string]interface {}) (len=1) {
Step #101 - "verify private-zonal-with-networking":         	            	+  (string) (len=34) "insecureKubeletReadonlyPortEnabled": (bool) true
Step #101 - "verify private-zonal-with-networking":         	            	+ },
Step #101 - "verify private-zonal-with-networking":         	            	  (string) (len=6) "labels": (map[string]interface {}) (len=2) {
Step #101 - "verify private-zonal-with-networking":         	Test:       	TestPrivateZonalWithNetworking
```

wyardley:
Maybe this is due to the default changing, as mentioned above?

apeabody:
> Maybe this is due to the default changing, as mentioned above?

Yup, just found it in the other fresh CI runs as well.

prune998:
@wyardley I did more testing... trying to config things manually and finding a TF config that would NOT re-create all the nodes pools and all the nodes inside of them.

I started by setting `cpu_manager_policy=""`  (empty value instead of null)

```
   # google_container_node_pool.pools["spot-n2dst8"] will be updated in-place
   ~ resource "google_container_node_pool" "pools" {
         id                          = "projects/my-project/locations/us-central1/clusters/qa-cluster/nodePools/spot-n2dst8"
         name                        = "spot-n2dst8"
         # (11 unchanged attributes hidden)
       ~ node_config {
           ~ kubelet_config {
               - cpu_manager_policy                     = "none" -> null
                 # (4 unchanged attributes hidden)
             }
             # (5 unchanged blocks hidden)
         }
         # (5 unchanged blocks hidden)
     }
 Plan: 0 to add, 1 to change, 0 to destroy.
```

That applied fine and had the nodepool modified with the whole `kubectlConfig` removed (with `gcloud container node-pools describe`)

Without touching nothing, I re-plan, and this time it's showing:

```
  # google_container_node_pool.pools["spot-n2dst8"] will be updated in-place
  ~ resource "google_container_node_pool" "pools" {
      id                          = "projects/my-project/locations/us-central1/clusters/qa-cluster/nodePools/spot-n2dst8"
      name                        = "spot-n2dst8"
      # (11 unchanged attributes hidden)
    ~ node_config {
        + kubelet_config {
              # (1 unchanged attribute hidden)
          }
          # (5 unchanged blocks hidden)
      }
      # (5 unchanged blocks hidden)
  }
Plan: 0 to add, 1 to change, 0 to destroy.
```

Which is not what I expected... but... apply failed with a PANIC error

```
│ Error: Plugin did not respond
│ 
│   with google_container_node_pool.pools["spot-n2dst8"],
│   on cluster.tf line 527, in resource "google_container_node_pool" "pools":
│  527: resource "google_container_node_pool" "pools" {
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may
│ contain more details.
╵
Stack trace from the terraform-provider-google_v6.6.0_x5 plugin:
panic: interface conversion: interface {} is nil, not map[string]interface {}
goroutine 75 [running]:
github.com/hashicorp/terraform-provider-google/google/services/container.expandKubeletConfig({0x107a37320?, 0x1400118b0b0?})
 github.com/hashicorp/terraform-provider-google/google/services/container/node_config.go:1151 +0x3c8
github.com/hashicorp/terraform-provider-google/google/services/container.nodePoolUpdate(0x140014a5b00, {0x108377c60?, 0x14001336a80}, 0x14001284840, {0x0, 0x0}, 0x274a48a7800)
 github.com/hashicorp/terraform-provider-google/google/services/container/resource_container_node_pool.go:1821 +0x239c
github.com/hashicorp/terraform-provider-google/google/services/container.resourceContainerNodePoolUpdate(0x140014a5b00, {0x108377c60?, 0x14001336a80?})
 github.com/hashicorp/terraform-provider-google/google/services/container/resource_container_node_pool.go:752 +0x290
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).update(0x1083ddcb8?, {0x1083ddcb8?, 0x140014d09f0?}, 0xd?, {0x108377c60?, 0x14001336a80?})
 github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/schema/resource.go:800 +0x134
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0x1400108ddc0, {0x1083ddcb8, 0x140014d09f0}, 0x14000b4e340, 0x140014a5880, {0x108377c60, 0x14001336a80})
 github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/schema/resource.go:919 +0x658
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*GRPCProviderServer).ApplyResourceChange(0x1400111f8d8, {0x1083ddcb8?, 0x140014d0900?}, 0x14001befe50)
 github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/schema/grpc_provider.go:1078 +0xb08
github.com/hashicorp/terraform-plugin-mux/tf5muxserver.(*muxServer).ApplyResourceChange(0x1083ddcf0?, {0x1083ddcb8?, 0x140014d0600?}, 0x14001befe50)
 github.com/hashicorp/terraform-plugin-mux@v0.15.0/tf5muxserver/mux_server_ApplyResourceChange.go:36 +0x184
github.com/hashicorp/terraform-plugin-go/tfprotov5/tf5server.(*server).ApplyResourceChange(0x1400051d360, {0x1083ddcb8?, 0x140010bfbf0?}, 0x140010b16c0)
 github.com/hashicorp/terraform-plugin-go@v0.23.0/tfprotov5/tf5server/server.go:865 +0x2b0
github.com/hashicorp/terraform-plugin-go/tfprotov5/internal/tfplugin5._Provider_ApplyResourceChange_Handler({0x108314f40?, 0x1400051d360}, {0x1083ddcb8, 0x140010bfbf0}, 0x140014a4200, 0x0)
 github.com/hashicorp/terraform-plugin-go@v0.23.0/tfprotov5/internal/tfplugin5/tfplugin5_grpc.pb.go:518 +0x164
google.golang.org/grpc.(*Server).processUnaryRPC(0x140010f6400, {0x1083ddcb8, 0x140010bfb60}, {0x1083e8f88, 0x140006c6180}, 0x140014a0480, 0x14000c06000, 0x10a13df98, 0x0)
 google.golang.org/grpc@v1.65.0/server.go:1379 +0xba0
google.golang.org/grpc.(*Server).handleStream(0x140010f6400, {0x1083e8f88, 0x140006c6180}, 0x140014a0480)
 google.golang.org/grpc@v1.65.0/server.go:1790 +0xc80
google.golang.org/grpc.(*Server).serveStreams.func2.1()
 google.golang.org/grpc@v1.65.0/server.go:1029 +0x8c
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 41
 google.golang.org/grpc@v1.65.0/server.go:1040 +0x150
Error: The terraform-provider-google_v6.6.0_x5 plugin crashed!
This is always indicative of a bug within the plugin. It would be immensely
helpful if you could report the crash with the plugin's maintainers so that it
can be fixed. The output above should help diagnose the issue.
```

prune998:
also, old nodepools will have the whole `kubeletConfig` section removed,
but newer nodepools will keep "insecureKubeletReadonlyPortEnabled" defined

```
    "kubeletConfig": {
      "insecureKubeletReadonlyPortEnabled": false
    },
```

But using gcloud, it gets rid of it:

```
❯ cat /tmp/kubelet-config
kubeletConfig:
  cpuManagerPolicy: ""

gcloud container node-pools update  --project my-project --cluster qa-cluster --location us-central1  gke-shared-spot-n2dst32  --system-config-from-file=/tmp/kubelet-config
```

wyardley:
```
$ gcloud container node-pools update xx --cluster=CLUSTER_NAME --location=LOCATION --no-enable-insecure-kubelet-readonly-port
```
(https://cloud.google.com/kubernetes-engine/docs/how-to/disable-kubelet-readonly-port) would be the way to disable it explicitly, so that could be worth a try as well if you haven't. Can't find the link, but IIRC, as of a certain version, it will be disabled by default, and at that point, all of this will be less important.

wyardley:
Guessing you saw, @apeabody, but seems like the behavior will continue to be inconsistent / unpredictable due to the [gradual rollout](https://github.com/hashicorp/terraform-provider-google/issues/19792#issuecomment-2405836119), so it's possible you'll see more tests stop working.

Not sure if a partial fix as a breaking release now is better, and letting people pull in the provider fixes as they want / need, or waiting til 6.7.0 and 5.44.2 are out (I'm assuming the latter) -- see [here](https://github.com/hashicorp/terraform-provider-google/issues/19792#issuecomment-2405854279) and adjusting the required version specification once those are out (up to you if you want me to exclude all other 6.x versions up to 6.7, or allow the intermediate ones as well, since they might work for people).

apeabody:
> Guessing you saw, @apeabody, but seems like the behavior will continue to be inconsistent / unpredictable due to the [gradual rollout](https://github.com/hashicorp/terraform-provider-google/issues/19792#issuecomment-2405836119), so it's possible you'll see more tests stop working.
> 
> Not sure if a partial fix as a breaking release now is better, and letting people pull in the provider fixes as they want / need, or waiting til 6.7.0 and 5.44.2 are out (I'm assuming the latter) -- see [here](https://github.com/hashicorp/terraform-provider-google/issues/19792#issuecomment-2405854279) and adjusting the required version specification once those are out (up to you if you want me to exclude all other 6.x versions up to 6.7, or allow the intermediate ones as well, since they might work for people).

Hi @wyardley!

Yeah - Been keeping any eye.  I would agree my preference is the updated provider versions next week as the best fix.

wyardley:
Are you planning on shipping #2126 in v34, or are you going to release one more with 5.44.x supported?

apeabody:
> #2126

Hi @wyardley - At this point we are working on v34: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2133

wyardley:
Yes, I get that, just wondering if you want to get this out in 34 still allowing 5.x, and ship 2126 in a later major; if it’s going to go in v34 as well, we can probably just require 6.7 here (but at the cost of not supporting this feature in 5.x at all)

apeabody:
> Yes, I get that, just wondering if you want to get this out in 34 still allowing 5.x, and ship 2126 in a later major; if it’s going to go in v34 as well, we can probably just require 6.7 here (but at the cost of not supporting this feature in 5.x at all)

Unfortunately the repo's release system doesn't support back port releases.  But we only have a single breaking PR (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2109) merged so far, so we could temporarily revert for a final v34 release next week.

wyardley:
This one is also breaking (tpg version), hence the bang in the PR title etc.

So it also has to go out with a major release, I think, it’s just a question of whether it makes sense to ship this breaking change before that one, which will fully disallow older versions, I think?

apeabody:
> This one is also breaking (tpg version), hence the bang in the PR title etc.
> 
> So it also has to go out with a major release, I think, it’s just a question of whether it makes sense to ship this breaking change before that one, which will fully disallow older versions, I think?

Got it - Yeah, let's see if we can wait on https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2126

wyardley:
@apeabody updated if you want to see how things look with those latest provider updates reflected. Obviously if this is released in the next major, these ugly version specifications can be greatly simplified if #2126 then just updates the min version to `6.7.0`.

autogen/main/versions.tf.tmpl
```diff
@@ -24,11 +24,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 5.40.0, < 7"
+      version = ">= 5.44.2, !=6.0.0, !=6.0.1, !=6.1.0, !=6.2.0, !=6.3.0, !=6.4.0, !=6.5.0, !=6.6.0, < 7"
```
wyardley, line 27: I didn't see any patch releases between 6.1 and 6.6, but let me know if I missed something.
An alternate approach might be to allow some of the other 6.x ones that are not completely broken, but requiring either 5.44.2 or 6.7.0 is probably the safest.

apeabody: Thanks @wyardley - I agree this is the least error prone for now, and in the subsequent breaking release we can drop support for v5 (for https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2126), which will again simplify the version constraint.

autogen/main/cluster.tf.tmpl
apeabody: Hi @wyardley - Looks like `gcfs_config ` was promoted to GA in [v6.5.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v6.5.0), so in the subsequent breaking release (35?) this can be resolved.

apeabody (approved):

wyardley:
@apeabody is the test failure now due to this setting / expectations, or another transient failure?

apeabody:
> @apeabody is the test failure now due to this setting / expectations, or another transient failure?

node pool timed out, retriggered

PR #2095, opened 2024-09-13
fix: correct gcfs_config value

apeabody:


wyardley (approved):
This appears to work for me in a limited test locally, and is probably better than the the version I would have submitted. 🙏 🚀

examples/node_pool/main.tf
```diff
@@ -51,6 +51,7 @@ module "gke" {
       max_count       = 2
       service_account = var.compute_engine_service_account
       auto_upgrade    = true
+      enable_gcfs     = false
```
wyardley, line 54: 🙏  for adding an example so it'll get tested as well.

PR #2099, opened 2024-09-14
docs: clarify templating language in use

wyardley:
Since Google projects often use other templating languages (Golang and Ruby have both been used for some Terraform-adjacent projects), clarify that the templating language in use is Jinja2


apeabody (approved):

apeabody:
Thanks @wyardley!

PR #1756, opened 2023-10-01
feat: support confidential nodepools

abhikaddy:
This fixes open issue #1386.

Issue #1386, opened 2022-09-03
Support confidential nodepools


jawnsy:
### TL;DR

We currently support confidential clusters (a cluster-level setting that enforces that all nodepools are confidential VMs), but do not support the setting at a per-nodepool level. We could add a new node_pool setting (e.g. `enable_confidential_nodes`) to enable this on a per-nodepool basis, similar to the `enable_secure_boot` setting)

### Additional information

[GKE supports mixed clusters](https://cloud.google.com/blog/products/identity-security/announcing-general-availability-of-confidential-gke-nodes) where some nodepools are confidential instances (n2d or c2d) but others run on other instances (n1, n2, c2, t2d), but the submodule does not support this.

bharathkkb:
@jawnsy happy to review a PR exposing this!

maksym-kursin:
It seems impossible to implement it right now because a confidential node is not supported by [container_node_pool](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool) resource.

Here is the related provider issue: https://github.com/hashicorp/terraform-provider-google/issues/13127

DrFaust92:
@apeabody, I think this can be closed as its supported already

apeabody:
Thanks, added in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1815

redoak666:
#1815 does not include support for nodepool level enablement of confidential nodes, only cluster level is supported
#1756 is the correct fix, but NOT merged

redoak666:
as of today, the latest release v32.0.4 lacks of capability of nodepool level enablement of confidential nodes

redoak666:
Since v30 is the release that confidential node goes GA, can this confidential nodepool support backported to v30?

PR #2110, opened 2024-09-23
feat: added confidential-nodes flag for node-pools

t-indumathy:
_**Motivation**_
The current set of confidential node support is available only at the cluster level, but the confidential node support at the node level is required to have new node pools created with confidential nodes.

_**Summary of changes**_
Added `enable_confidential_nodes` variable at the node pool level.

**_P.S_**: This is addressing the issue https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1386 and has same changes as that of the PR https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1756. As the PR author has not responded for the review comments, I'm thus creating a fresh PR as this feature is needed at the earliest.

Fixes: #1386
Closes: #1756

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -597,7 +597,7 @@ resource "google_container_cluster" "primary" {
     content {
       enable_private_endpoint     = private_cluster_config.value.enable_private_endpoint
       enable_private_nodes        = private_cluster_config.value.enable_private_nodes
-      master_ipv4_cidr_block      = private_cluster_config.value.master_ipv4_cidr_block
+      master_ipv4_cidr_block      = var.private_endpoint_subnetwork == null ? private_cluster_config.value.master_ipv4_cidr_block : null
       private_endpoint_subnetwork = private_cluster_config.value.private_endpoint_subnetwork
       dynamic "master_global_access_config" {
         for_each = var.master_global_access_enabled ? [var.master_global_access_enabled] : []
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -720,7 +720,8 @@ locals {
     "enable_confidential_storage",
     "consume_reservation_type",
     "reservation_affinity_key",
-    "reservation_affinity_values"
+    "reservation_affinity_values",
+    "enable_confidential_nodes",
   ]
 }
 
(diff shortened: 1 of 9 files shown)


autogen/main/cluster.tf.tmpl
```diff
@@ -1084,6 +1085,14 @@ resource "google_container_node_pool" "windows_pools" {
       enable_secure_boot          = lookup(each.value, "enable_secure_boot", false)
       enable_integrity_monitoring = lookup(each.value, "enable_integrity_monitoring", true)
     }
+
+    dynamic "confidential_nodes" {
+      for_each = lookup(each.value, "enable_confidential_nodes", false) ? [true] : []
```
apeabody, line 1090: I suggest that if `enable_confidential_nodes` is actively set to `false`, then the block be generated with enabled =  value, rather than omitted.

```suggestion
      for_each = lookup(each.value, "enable_confidential_nodes", null) != null ? [each.value.confidential_nodes] : []
```

apeabody (changes requested):
Thanks for the contribution @t-indumathy!

Can you please document the new variable here:
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/README.md?plain=1#L190

autogen/main/cluster.tf.tmpl
t-indumathy: Done.

t-indumathy:
> Thanks for the contribution @t-indumathy!
> 
> Can you please document the new variable here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/README.md?plain=1#L190

This is already present at https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/autogen/main/README.md?plain=1#L204 .

t-indumathy:
@apeabody Can you pls let me know the details on why the pipeline is unsuccessful.

apeabody:
> @apeabody Can you pls let me know the details on why the pipeline is unsuccessful.

Hi @t-indumathy - I'm going to re-run the tests.  A node pool failed to create, but I don't believe it's related to this change.

apeabody (approved):

t-indumathy:
@apeabody can you pls backport this feature to v30.0.

apeabody:
> @apeabody can you pls backport this feature to v30.0.

Hi @t-indumathy - Unfortunately we don't do backport releases and instead recommend updating to the latest module version, however if required you could fork the v30.0 release for your changes and use a custom module source.  Best Wishes!

PR #2113, opened 2024-09-25
refactor: replace non-default `lookup()` calls with direct access

wyardley:
Resolve `terraform_deprecated_lookup` warning for `lookup()` calls with only 2 arguments (no default).

See:
https://github.com/terraform-linters/tflint-ruleset-terraform/blob/main/docs/rules/terraform_deprecated_lookup.md


apeabody (approved):

Issue #2108, opened 2024-09-20
Tests using terraform test framework


wyardley:
### TL;DR
Is there any movement in the overall ecosystem for these modules to use the Terraform test framework for some of the tests? Converting the existing integration tests might be hard, but using plan only or mock provider tests might be a fast way to get some feedback (and could also probably allow running in publicly visible CI checks in either GHA or cloudbuild, since it wouldn't need real GCP creds at all).

This would make it pretty fast / cheap / easy to test a lot of different scenarios as far as input variables, and shift some of the basic tests "left" a bit.

I'm not sure if it would make sense to template the tests too (probably, though might have to have separate tests or separate tf vars, depending on how varied the scenarios are), or how exactly they'd run in CI, but I'd be happy to try writing a few if there's an example, or if you all are able to setup the skeleton framework.

Feel free to close if this is not planned or not possible.

### Detailed design

The following (very simple) test should pass if you run `terraform init` and `terraform test` in `terraform-google-kubernetes-engine/modules/beta-private-cluster` as `cluster.tftest.hcl`. Hopefully this shows the ways it can be used to do at least basic tests of the module's logic by using different variables.

Note that with a plan only test, some of the computed attributes won't be testable. However, for testing some of the logic, I think it will still prove useful.

I can provide an example with a mock provider as well, if desired.

```hcl
provider "google-beta" {
  project = "foo-testproject"
}

override_data {
  target = data.google_compute_zones.available
  values = {
    id = "projects/foo-testproject/regions/us-central1",
    names = [
      "us-central1-a",
      "us-central1-b",
      "us-central1-c",
      "us-central1-f",
    ],
    project = "foo-testproject"
    region  = "us-central1",
    status  = "null",
  }
}

override_data {
  target = data.google_container_engine_versions.region
  values = {
    latest_master_version = "1.30.4-gke.1348000"
  }
}

override_data {
  target = data.google_container_engine_versions.zone
  values = {
    latest_master_version = "1.30.4-gke.1348000"
  }
}
# Default variables -- override as needed within run blocks for individual
# tests.
variables {
  project_id                 = "foo-testproject"
  name                       = "testcluster"
  regional                   = true
  region                     = "us-central1"
  network                    = "default"
  subnetwork                 = "default"
  ip_range_pods              = "pod-range"
  ip_range_services          = "service-range"
  enable_private_endpoint    = true
  enable_private_nodes       = true
[...]

PR #2114, opened 2024-09-25
chore: update kitchen config

wyardley:
With the test kitchen version (3.5) used by the tests, there are some parameters which have moved to new locations.

Follow the suggestions of `kitchen doctor`.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -20,6 +20,7 @@ provisioner:
   name: "terraform"
 
 transport:
+  name: "terraform"
   command_timeout: 2700
 
 verifier:
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -14,11 +14,13 @@
 ---
 driver:
   name: "terraform"
-  command_timeout: 2700
   verify_version: false
 
 provisioner:
   name: "terraform"
+
+transport:
+  command_timeout: 2700
 
 verifier:
   name: terraform
@@ -30,14 +32,14 @@ platforms:
 
 suites:
   - name: "shared_vpc"
-    driver:
+    transport:
       root_module_directory: test/fixtures/shared_vpc
     verifier:
       systems:
         - name: shared_vpc
           backend: local
   - name: "safer_cluster"
-    driver:
+    transport:
       root_module_directory: test/fixtures/safer_cluster
     verifier:
       systems:
@@ -50,14 +52,14 @@ suites:
           controls:
             - network
   - name: "simple_regional"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_regional
     verifier:
       systems:
         - name: simple_regional
           backend: local
   - name: "simple_regional_with_networking"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_regional_with_networking
     verifier:
       systems:
@@ -74,35 +76,35 @@ suites:
           controls:
             - network
   - name: "simple_regional_private"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_regional_private
     verifier:
       systems:
         - name: simple_regional_private
           backend: local
   - name: "simple_regional_with_gateway_api"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_regional_with_gateway_api
     verifier:
       systems:
         - name: simple_regional_with_gateway_api
           backend: local
   - name: "simple_regional_with_ipv6"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_regional_with_ipv6
     verifier:
       systems:
         - name: simple_regional_with_ipv6
           backend: local
   - name: "simple_regional_with_kubeconfig"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_regional_with_kubeconfig
     verifier:
       systems:
         - name: simple_regional_with_kubeconfig
           backend: local
   - name: "simple_zonal"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_zonal
     verifier:
       systems:
@@ -116,7 +118,7 @@ suites:
           controls:
             - gcp
   - name: "simple_zonal_private"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_zonal_private
     verifier:
       systems:
@@ -125,7 +127,7 @@ suites:
           controls:
             - gcloud
   - name: "stub_domains"
-    driver:
+    transport:
       root_module_directory: test/fixtures/stub_domains
     verifier:
       systems:
@@ -137,27 +139,27 @@ suites:
   # Disabled due to issue #264
   # (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/264)
   #  - name: stub_domains_private
-  #    driver:
+  #    transport:
   #      root_module_directory: test/fixtures/stub_domains_private
   #      systems:
   #        - name: stub_domains_private
   #          backend: local
   - name: "upstream_nameservers"
-    driver:
+    transport:
       root_module_directory: test/fixtures/upstream_nameservers
     verifier:
       systems:
         - name: upstream_nameservers
           backend: local
   - name: "stub_domains_upstream_nameservers"
-    driver:
+    transport:
       root_module_directory: test/fixtures/stub_domains_upstream_nameservers
     verifier:
       systems:
         - name: stub_domains_upstream_nameservers
           backend: local
   - name: "workload_identity"
-    driver:
+    transport:
       root_module_directory: test/fixtures/workload_identity
     verifier:
       systems:
@@ -170,14 +172,14 @@ suites:
           controls:
             - gcp
   - name: "workload_metadata_config"
-    driver:
+    transport:
       root_module_directory: test/fixtures/workload_metadata_config
     verifier:
       systems:
         - name: workload_metadata_config
           backend: local
   - name: "simple_windows_node_pool"
-    driver:
+    transport:
       root_module_directory: test/fixtures/simple_windows_node_pool
     verifier:
       systems:
@@ -190,7 +192,7 @@ suites:
           controls:
             - gcp
   - name: "deploy_service"
-    driver:
+    transport:
       root_module_directory: test/fixtures/deploy_service
     verifier:
       systems:
@@ -200,7 +202,7 @@ suites:
             - gcloud
             - kubectl
   - name: "node_pool"
-    driver:
+    transport:
       root_module_directory: test/fixtures/node_pool
     verifier:
       systems:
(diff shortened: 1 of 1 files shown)


.kitchen.yml
```diff
@@ -30,8 +33,6 @@ platforms:
 
 suites:
   - name: "shared_vpc"
-    driver:
-      root_module_directory: test/fixtures/shared_vpc
```
apeabody, line 34: Curious - Why is this moved under transport, while the other suites directory are still listed below?

wyardley: Oh good call, let me look at that 🤦
probably should be at the same scope but just under the new path.

wyardley: This should be the right way now

apeabody: Thanks!

apeabody:
Hmm, I'll re-trigger just to be sure.

```
-----> Starting Test Kitchen (v3.5.0)
-----> Converging <stub-domains-local>...
$$$$$$ Reading the Terraform client version...
       Terraform v1.9.6
       on linux_amd64
       + provider registry.terraform.io/hashicorp/google v6.4.0
       + provider registry.terraform.io/hashicorp/kubernetes v2.32.0
       + provider registry.terraform.io/hashicorp/random v3.6.3
$$$$$$ Finished reading the Terraform client version.
$$$$$$ Verifying the Terraform client version is in the supported interval of >= 0.11.4, < 2.0.0...
$$$$$$ Finished verifying the Terraform client version.
$$$$$$ Selecting the kitchen-terraform-stub-domains-local Terraform workspace...
       Switched to workspace "kitchen-terraform-stub-domains-local".
$$$$$$ Finished selecting the kitchen-terraform-stub-domains-local Terraform workspace.
$$$$$$ Downloading the modules needed for the Terraform configuration...
$$$$$$ Finished downloading the modules needed for the Terraform configuration.
$$$$$$ Validating the Terraform configuration files...
       Success! The configuration is valid.
       
$$$$$$ Finished validating the Terraform configuration files.
$$$$$$ Building the infrastructure based on the Terraform configuration...
       
       Error: No value for required variable
       
         on variables.tf line 19:
         19: variable "project_id" {
       
       The root module input variable "project_id" is not set, and has no default
       value. Use a -var or -var-file command line argument to provide a value for
       this variable.
       
       Error: No value for required variable
       
         on variables.tf line 24:
         24: variable "name" {
       
       The root module input variable "name" is not set, and has no default value.
       Use a -var or -var-file command line argument to provide a value for this
       variable.
       
       Error: No value for required variable
       
         on variables.tf line 53:
         53: variable "network" {
       
       The root module input variable "network" is not set, and has no default
       value. Use a -var or -var-file command line argument to provide a value for
       this variable.
       
       Error: No value for required variable
       
         on variables.tf line 64:
         64: variable "subnetwork" {
       
       The root module input variable "subnetwork" is not set, and has no default
       value. Use a -var or -var-file command line argument to provide a value for
       this variable.
       
       Error: No value for required variable
       
         on variables.tf line 135:
        135: variable "ip_range_pods" {
       
       The root module input variable "ip_range_pods" is not set, and has no default
       value. Use a -var or -var-file command line argument to provide a value for
       this variable.
       
       Error: No value for required variable
       
         on variables.tf line 146:
        146: variable "ip_range_services" {
       
       The root module input variable "ip_range_services" is not set, and has no
       default value. Use a -var or -var-file command line argument to provide a
       value for this variable.
>>>>>> Expected process to exit with [0], but received '1'
---- Begin output of terraform apply -auto-approve -lock=true -lock-timeout=0s -input=false -no-color -parallelism=10 -refresh=true   ----
STDOUT: 
STDERR: Error: No value for required variable

  on variables.tf line 19:
  19: variable "project_id" {

The root module input variable "project_id" is not set, and has no default
value. Use a -var or -var-file command line argument to provide a value for
this variable.

Error: No value for required variable

  on variables.tf line 24:
  24: variable "name" {

The root module input variable "name" is not set, and has no default value.
Use a -var or -var-file command line argument to provide a value for this
variable.

Error: No value for required variable

  on variables.tf line 53:
  53: variable "network" {

The root module input variable "network" is not set, and has no default
value. Use a -var or -var-file command line argument to provide a value for
this variable.

Error: No value for required variable

  on variables.tf line 64:
  64: variable "subnetwork" {

The root module input variable "subnetwork" is not set, and has no default
value. Use a -var or -var-file command line argument to provide a value for
this variable.

Error: No value for required variable

  on variables.tf line 135:
 135: variable "ip_range_pods" {

The root module input variable "ip_range_pods" is not set, and has no default
value. Use a -var or -var-file command line argument to provide a value for
this variable.

Error: No value for required variable

  on variables.tf line 146:
 146: variable "ip_range_services" {

The root module input variable "ip_range_services" is not set, and has no
default value. Use a -var or -var-file command line argument to provide a
value for this variable.
---- End output of terraform apply -auto-approve -lock=true -lock-timeout=0s -input=false -no-color -parallelism=10 -refresh=true   ----
Ran terraform apply -auto-approve -lock=true -lock-timeout=0s -input=false -no-color -parallelism=10 -refresh=true   returned 1
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Converge failed on instance <stub-domains-local>.  Please see .kitchen/logs/stub-domains-local.log for more details
>>>>>> ----------------------
>>>>>> Please see .kitchen/logs/kitchen.log for more details
>>>>>> Also try running `kitchen diagnose --all` for configuration
```

wyardley:
I can repro that error locally, even without creds or a valid project setup... I'll take a look 🤔

wyardley:
kitchen-terraform is the latest version in the container too, so it's not an old version thing.
What's odd is that if I revert the changes either fully or in part, I still get this if I try to run `kitchen converge shared-vpc-local` or `source /usr/local/bin/task_helper_functions.sh && kitchen_do converge shared-vpc-local`

```
│ Error: No value for required variable
│ 
│   on variables.tf line 17:
│   17: variable "project_ids" {
│ 
│ The root module input variable "project_ids" is not set, and has no default
│ value. Use a -var or -var-file command line argument to provide a value for
│ this variable.
╵
```

Anyway, if you have any pointers as to where this may be going wrong or where these vars are supposed to come from (a var file somewhere?), I can investigate more, or we can close this if it's taking your time for no good reason.

apeabody:
> kitchen-terraform is the latest version in the container too, so it's not an old version thing. What's odd is that if I revert the changes either fully or in part, I still get this if I try to run `kitchen converge shared-vpc-local` or `source /usr/local/bin/task_helper_functions.sh && kitchen_do converge shared-vpc-local`
> 
> ```
> │ Error: No value for required variable
> │ 
> │   on variables.tf line 17:
> │   17: variable "project_ids" {
> │ 
> │ The root module input variable "project_ids" is not set, and has no default
> │ value. Use a -var or -var-file command line argument to provide a value for
> │ this variable.
> ╵
> ```
> 
> Anyway, if you have any pointers as to where this may be going wrong or where these vars are supposed to come from (a var file somewhere?), I can investigate more, or we can close this if it's taking your time for no good reason.

Thanks @wyardley!

We strongly encourage that all new tests be written using CFT, rather than Kitchen, so it's been a while since I've used it significantly.  However this inspec file might be useful for the variables: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/integration/stub_domains/inspec.yml

Ideally these Kitchen tests would be re-written using CFT, so keep that in mind before you invest a lot of time in this space. :)

wyardley:
> We strongly encourage that all new tests be written using CFT, rather than Kitchen

Right, understood... was really just trying to fix what seemed to be a very simple deprecation warning, but if it ends up being too involved, probably not worth it.

That said, might be good to update the [help text](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/7ea17522659a6fb911d72af87f219ccfd8a73765/CONTRIBUTING.md?plain=1#L44-L50) to make it more clear, as well as to provide links to whatever docs exist about the new framework.

From a quick look, it had seemed like maybe kitchen was still used for provisioning, and that the go tests were just called instead of inspec for verification, but maybe I misunderstood.

Either way, since kitchen-terraform is deprecated, totally understand the move to something else (also see #2108 -- while I'm mostly talking about unit testing with mocked provider or plan-only there, the native tests could be used for provisioning directly as well)

wyardley:
Also, I think the inspec config is for verification vs. planning? In this case, since there are no defaults set in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/test/fixtures/shared_vpc/variables.tf, I think kitchen is barfing because these required vars have no values, and I'm still not seeing where they come from.

FWIW, I ran the container and then ran
`kitchen converge stub-domains-local --test-base-path=test/integration` off of main (without these changes), after cleaning up various lockfiles and kitchen yaml configs from previous failed runs, and still get the same error. So still having trouble understanding how this works / worked before.

apeabody:
> > We strongly encourage that all new tests be written using CFT, rather than Kitchen
> 
> Right, understood... was really just trying to fix what seemed to be a very simple deprecation warning, but if it ends up being too involved, probably not worth it. I've worked with test kitchen pretty extensively in the past, but haven't touched it for a couple of years.
> 
> That said, might be good to update the [help text](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/7ea17522659a6fb911d72af87f219ccfd8a73765/CONTRIBUTING.md?plain=1#L44-L50) to make it more clear, as well as to provide links to whatever docs exist about the new framework.
> 
> From a quick look, it had seemed like maybe kitchen was still used for provisioning, and that the go tests were just called instead of inspec for verification, but maybe I misunderstood.
> 
> Either way, since kitchen-terraform is deprecated, totally understand the move to something else (also see #2108 -- while I'm mostly talking about unit testing with mocked provider or plan-only there, the native tests could be used for provisioning directly as well)

This repo might be the only that has both Kitchen and CFT (go) tests, so it gets a bit muddled.  Here is a simple CFT only example: https://github.com/terraform-google-modules/terraform-google-cloud-storage/blob/master/build/int.cloudbuild.yaml

The `prepare` stage is used by both, but the `init-all` is CFT specific, while the `create-all` is for Kitchen.  Here is the general documentation for the new style tests: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/tree/master/infra/blueprint-test

.kitchen.yml
```diff
@@ -14,12 +14,14 @@
 ---
 driver:
   name: "terraform"
-  command_timeout: 2700
   verify_version: false
 
 provisioner:
   name: "terraform"
 
+transport:
```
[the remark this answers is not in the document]
wyardley: Good call. And I see that [here](https://github.com/newcontext-oss/kitchen-terraform/blob/e23570368e3f004edbca8149840df3f23154fcd6/kitchen.yml#L7). (side note: I don't think any of these need to be quoted?)

Updated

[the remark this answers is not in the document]
apeabody: Yeah, the quotes are almost certainly optional as the type "string" can be inferred for these values. :)

apeabody (approved):

PR #2116, opened 2024-09-25
chore: update example module sources and versions

wyardley:
- Update module source. This should prevent the source from getting rewritten when tests are run locally
- Add missing version spec to match other modules

Follow-up to https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2115#issuecomment-2375056239

For example this:
```hcl
module "permissions" {
  source = "../../modules/fleet-app-operator-permissions"
  # also no version specification here
}
```

gets rewritten like this when the docker test / lint make target is run.
```diff
--- a/examples/autopilot_private_firewalls/main.tf
+++ b/examples/autopilot_private_firewalls/main.tf
@@ -33,7 +33,7 @@ provider "kubernetes" {
 }
 
 module "gke" {
-  source                            = "../../modules/beta-autopilot-private-cluster/"
+  source                            = "wyardley/kubernetes-engine/google//modules/beta-autopilot-private-cluster"
   project_id                        = var.project_id
   name                              = "${local.cluster_type}-cluster"
   regional                          = true
```

### DIFF
--- a/examples/autopilot_private_firewalls/main.tf
+++ b/examples/autopilot_private_firewalls/main.tf
@@ -33,7 +33,9 @@ provider "kubernetes" {
 }
 
 module "gke" {
-  source                            = "../../modules/beta-autopilot-private-cluster/"
+  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster/"
+  version = "~> 33.0"
+
   project_id                        = var.project_id
   name                              = "${local.cluster_type}-cluster"
   regional                          = true
--- a/examples/autopilot_private_firewalls/main.tf
+++ b/examples/autopilot_private_firewalls/main.tf
@@ -33,7 +33,7 @@ provider "kubernetes" {
 }
 
 module "gke" {
-  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster/"
+  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster"
   version = "~> 33.0"
 
   project_id                        = var.project_id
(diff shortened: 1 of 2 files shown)


examples/autopilot_private_firewalls/main.tf
```diff
@@ -33,7 +33,9 @@ provider "kubernetes" {
 }
 
 module "gke" {
-  source                            = "../../modules/beta-autopilot-private-cluster/"
+  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster/"
+  version = "~> 33.0"
```
wyardley, line 37: not sure why lint is complaining about this one, unless it was from the previous push?

examples/autopilot_private_firewalls/main.tf
```diff
@@ -33,7 +33,9 @@ provider "kubernetes" {
 }
 
 module "gke" {
-  source                            = "../../modules/beta-autopilot-private-cluster/"
+  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster/"
```
apeabody, line 36: This might be the reason?

```suggestion
  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-autopilot-private-cluster"
```

wyardley: Thanks, good eye

FWIW, vanilla tflint 0.53.0 with the regular ruleset doesn't complain locally.

apeabody: @wyardley Here is the tflint config used for /example if you are curious: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/home/.tflint.example.hcl

apeabody (approved):

wyardley:
@apeabody ack, another flake maybe?

apeabody:
> @apeabody ack, another flake maybe?

Yup, slow API activation

PR #2117, opened 2024-09-26
chore: ignore top level `.terraform.lock`

wyardley:
If there's a template repo, maybe this should get added there as well?


apeabody (approved):

Here is the template repo: https://github.com/terraform-google-modules/terraform-google-module-template/blob/master/.gitignore

Issue #24, opened 2018-10-30
Relax kubernetes version


adrienthebo:
Discovered when verifying https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/20.

Running `kitchen converge deploy-service-local`:

```
     * google_container_cluster.primary: googleapi: Error 400: Master version "1.10.6-gke.2" is unsupported., badRequest
```

Per https://cloud.google.com/kubernetes-engine/versioning-and-upgrades#specifying_cluster_version we can use a looser version constraint. Starting with `1.10.6` should give us a bit more flexibility and will require fewer changes to this module.

PR #2118, opened 2024-09-26
fix: add target tags to `node_pool_auto_config` for standard clusters

### LINKED ISSUE #2104: GKE Standard Clusters with AutoProvisioned NodePools not adding the Firewall `target_tags`
### TL;DR

The [firewall rules](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/private-cluster-update-variant/firewall.tf) created with `target_tags = [local.cluster_network_tag]` have an expectation that the nodepools will have this tag (`"gke-${var.name}"`) applied.

This tag should be added to `network_tags` to ensure it is set in [`node_pool_auto_config`](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/modules/private-cluster-update-variant/cluster.tf#L205-L212) for AutoProvisioned NodePools also.

### Expected behavior

The generated tags used by the firewall rules should be added to the `network_tags` for autoprovisioned nodepools the same as manual nodepools

### Observed behavior

Firewall rules for allowing admission webhook etc aren't applying to the autoprovisioned nodepools as targets

### Terraform Configuration

```hcl
module "gke" {
  source     = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster-update-variant"
  version    = "~> 32.0"
  project_id = local.project_id
  name       = var.gke_cluster_name

  add_cluster_firewall_rules = true
  firewall_inbound_ports     = ["8080", "8443", "9443", "15017"]

  #
  # `"egress-internet"` is a network-tag based on https://github.com/terraform-google-modules/terraform-example-foundation/blob/master/3-networks-dual-svpc/modules/base_shared_vpc/main.tf#L43
  #
  # manually adding the `"gke-${var.gke_cluster_name}"` is necessary to match the target_tag in the module firewall rules
  #
  network_tags = ["egress-internet", "gke-${var.gke_cluster_name}"]

  remove_default_node_pool = true
  cluster_autoscaling = {
    enabled             = true
    autoscaling_profile = "OPTIMIZE_UTILIZATION"
    min_cpu_cores       = 4
    max_cpu_cores       = 86
    min_memory_gb       = 16
    max_memory_gb       = 256
    disk_size           = 100
    disk_type           = "pd-standard"
    image_type          = "COS_CONTAINERD"
    gpu_resources       = []
    auto_repair         = true
    auto_upgrade        = false
    strategy            = "SURGE"
    max_surge           = 1
    max_unavailable     = 0
  }
 }
}

```
```


### Terraform Version

```sh
Terraform v1.9.3
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v5.42.0
+ provider registry.terraform.io/hashicorp/google-beta v5.42.0
+ provider registry.terraform.io/hashicorp/http v3.4.4
+ provider registry.terraform.io/hashicorp/null v3.2.2
+ provider registry.terraform.io/hashicorp/random v3.6.2
```
```


### Additional information

The autoprovisioned nodepool also doesnt honour the Resource Manager Tags provided as `node_pools_resource_manager_tags` .

Issue https://github.com/hashicorp/terraform-provider-google/issues/16614 mentioned it was added but only for AutoPilot clusters. The linked documentation (https://cloud.google.com/kubernetes-engine/docs/how-to/tags-firewall-policies#attach-tags-standard) has a cli mechanism to do the same for Standard Clusters

### PR DESCRIPTION
**Pull request by wyardley**
While #1817 added autopilot support for adding tags to `node_pool_auto_config` when `add_cluster_firewall_rules` is set to `true`, the same change did not apply for standard (non-autopilot) clusters with cluster level autoscaling (nodepool autoprovisioning) in place,

Fixes #2104

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -282,10 +282,10 @@ resource "google_container_cluster" "primary" {
 
 {% if autopilot_cluster != true %}
   dynamic "node_pool_auto_config" {
-    for_each = var.cluster_autoscaling.enabled && length(var.network_tags) > 0 ? [1] : []
+    for_each = var.cluster_autoscaling.enabled && (length(var.network_tags) > 0 || var.add_cluster_firewall_rules) ? [1] : []
     content {
       network_tags {
-        tags = var.network_tags
+        tags = var.add_cluster_firewall_rules ? (concat(var.network_tags, [local.cluster_network_tag])) : var.network_tags
       }
     }
   }
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -137,6 +137,7 @@ steps:
 - id: apply simple-regional-cluster-autoscaling
   waitFor:
     - create-all
+    - destroy simple-autopilot-private-local
   name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
   args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage apply --verbose']
 - id: verify simple-regional-cluster-autoscaling
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -137,7 +137,7 @@ steps:
 - id: apply simple-regional-cluster-autoscaling
   waitFor:
     - create-all
-    - destroy simple-autopilot-private-local
+    - destroy simple-regional-private-local
   name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
   args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage apply --verbose']
 - id: verify simple-regional-cluster-autoscaling
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -134,6 +134,21 @@ steps:
     - verify simple-regional-private-local
   name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
   args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalPrivate --stage teardown --verbose']
+- id: apply simple-regional-cluster-autoscaling
+  waitFor:
+    - create-all
+  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
+  args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage apply --verbose']
+- id: verify simple-regional-cluster-autoscaling
+  waitFor:
+    - apply simple-regional-cluster-autoscaling
+  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
+  args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage verify --verbose']
+- id: destroy simple-regional-cluster-autoscaling
+  waitFor:
+    - verify simple-regional-cluster-autoscaling
+  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
+  args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage teardown --verbose']
 - id: init simple-regional-with-kubeconfig-local
   waitFor:
     - create-all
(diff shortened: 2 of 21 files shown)

### CODE REVIEW

**Review by wyardley — COMMENTED** [review]
@apeabody I'm not sure this example is one you'd want in `examples` -- I think I saw something about that; in that case, do I just fold the whole example into `fixtures`? Or should I just add more comments or adjust the sample code slightly?

**Comment by apeabody** [comment]
> @apeabody I'm not sure this example is one you'd want in `examples` -- I think I saw something about that; in that case, do I just fold the whole example into `fixtures`? Or should I just add more comments or adjust the sample code slightly?

Hi @wyardley! I think it's fine to add a relevant example, although we have to be careful with the concurrent test resource usage (might need to also waitFor a subsequent step).

While this repo has a `cft test run all --stage init`, it doesn't for the subsequent states, so you will need to add something like this:

build/int.cloudbuild.yaml
```
- id: apply simple-regional-cluster-autoscaling
  waitFor:
    - init-all
    - create-all
  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
  args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage apply --verbose']
- id: verify simple-regional-cluster-autoscaling
  waitFor:
    - apply simple-regional-cluster-autoscaling
  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
  args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage verify --verbose']
- id: destroy simple-regional-cluster-autoscaling
  waitFor:
    - verify simple-regional-cluster-autoscaling
  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
  args: ['/bin/bash', '-c', 'cft test run TestSimpleRegionalClusterAutoscaling --stage teardown --verbose']
```

**Comment by wyardley** [comment]
> Hi @wyardley! I think it's fine to add a relevant example,

Yeah, I guess seemed a little niche of a use case; obviously someone copying that one without thinking about what it's doing could end up opening up network holes unnecessarily? And not sure if we should adjust the wording or comments anywhere, or make a more descriptive directory name?

I also picked simple regional as the base for this more or less arbitrarily.

I'll add in the snippet you suggested for the Cloudbuild config (should maybe remove the lint one at some point?) and we can see how it goes.

If the fixture with the expected content is too far off base to adjust, I will probably need your help to figure out how that should be updated / adjusted as well.

**Comment by apeabody** [comment]
> I'll add in the snippet you suggested for the Cloudbuild config (should maybe remove the lint one at some point?) and we can see how it goes.

Sure: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2120

**Comment by apeabody** [comment]
Looks like this will need to also wait on a longer destroy for resource availability.

```
Error: Error waiting to create Network: Error waiting for Creating Network: Quota 'NETWORKS' exceeded.  Limit: 15.0 globally.
	metric name = compute.googleapis.com/networks
	limit name = NETWORKS-per-project
	limit = 15
	dimensions = map[global:global]


  with google_compute_network.main,
  on network.tf line 27, in resource "google_compute_network" "main":
  27: resource "google_compute_network" "main" {
```

wyardley:
> Looks like this will need to also wait on a longer destroy for resource availability.

Are you suggesting to bump the retry interval in `tft.WithRetryableTerraformErrors()`, or to override the timeout for the cloudrun job's step?

apeabody:
> > Looks like this will need to also wait on a longer destroy for resource availability.
> 
> Are you suggesting to bump the retry interval in `tft.WithRetryableTerraformErrors()`, or to override the timeout for the cloudrun job's step?

Sorry - Typo, that should be "later", not "longer.  Like this example: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/build/int.cloudbuild.yaml#L458

wyardley:
> Sorry - Typo, that should be "later", not "longer. Like this example

Since there's no explicit init step for this one, add it to the apply, right? Is 812e4d0773cd9dbd8b076f65b4d28bc9771c9330 what you had in mind?

apeabody:
> > Sorry - Typo, that should be "later", not "longer. Like this example
> 
> Since there's no explicit init step for this one, add it to the apply, right? Is [336cbff](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/commit/336cbff7c00b09eb32f97efa866adaf627e467ff) what you had in mind?

Yeah - As there is now an init all, we shouldn't need the per test inits, and ideally can be removed at some point.

Looks good, let's see how the tests go:

/gcbrun

apeabody:
FYI: Any step you depend on has to be "earlier" in the cloudbuild file.

`invalid build: invalid .steps field: build step https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/24 - "apply simple-regional-cluster-autoscaling" depends on "destroy simple-autopilot-private-local", which has not been defined`

wyardley:
Are we getting closer to a real failure yet? I can try to do the local setup if needed.

apeabody:
> Are we getting closer to a real failure yet? I can try to do the local setup if needed.

I don't think so, there is an API change which is breaking the test data as of this morning.  Will re-run this once that is updated.

wyardley:
@apeabody this should almost definitely still be failing at least partially because of the roughed in / bad fixture, right?
Is there an easy way to collect / create the values that should be in there without setting up the whole integration test setup locally?

wyardley:
Sorry, can you check if I missed something and / or anything else needs to be adjusted?

apeabody:
> Sorry, can you check if I missed something and / or anything else needs to be adjusted?

Looks good!

wyardley:
Ah, ok guess that last failure was another spurious one 🤷

thanks, glad this one is passing

apeabody (approved):

PR #2122, opened 2024-09-27
fix: ignore control plane network when private endpoint subnet is set

### LINKED ISSUE #2119: `private_endpoint_subnetwork` not working
### TL;DR

In #2009 support for `private_endpoint_subnetwork` was added. The problem is that `master_ipv4_cidr_block` has a default value which leads to both being set which again leads to the following error:

```
Error: googleapi: Error 400: When masterIpv4Cidr is set, privateEndpointSubnetwork must be unset.
```

### Expected behavior

If `private_endpoint_subnetwork` is set, `master_ipv4_cidr_block` should not be set on the underlying resource.

### Observed behavior

_No response_

### Terraform Configuration

```hcl
module "infra_gke_cluster" {
  source                              = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster-update-variant"
  version                             = "33.0.4"
  project_id                          = data.google_project.this.project_id
  name                                = "some-cluster"
  regional                            = false
  region                              = local.region
  zones                               = data.google_compute_zones.all_zones_in_region.names
  network                             = module.vpc.network_name
  subnetwork                          = local.subnets.gke.subnet_name
  ip_range_pods                       = "ip-range-pods"
  ip_range_services                   = "ip-range-svc"
  node_metadata                       = "GKE_METADATA"
  cluster_dns_provider                = "CLOUD_DNS"
  cluster_dns_domain                  = "infra.europe-west4"
  cluster_dns_scope                   = "VPC_SCOPE"
  maintenance_start_time              = "02:00"
  dns_cache                           = true
  remove_default_node_pool            = true
  gcs_fuse_csi_driver                 = true
  workload_config_audit_mode          = "BASIC"
  security_posture_mode               = "BASIC"
  security_posture_vulnerability_mode = "VULNERABILITY_BASIC"
  enable_gcfs                         = false # Image Streaming
  enable_private_nodes                = true
  enable_private_endpoint             = true
  create_service_account              = false
  master_ipv4_cidr_block              = null
  service_account                     = "****"
  private_endpoint_subnetwork         = local.subnets.gke-master.subnet_name
  enable_l4_ilb_subsetting            = true
  authenticator_security_group        = "***"
  gateway_api_channel                 = "CHANNEL_STANDARD"

  master_authorized_networks = [***]

  kubernetes_version = "1.30"

  node_pools_oauth_scopes = {
    "all" = [
      "https://www.googleapis.com/auth/cloud-platform",
      "https://www.googleapis.com/auth/devstorage.read_only"
    ]
  }

  cluster_autoscaling = {
    enabled             = true
    autoscaling_profile = "OPTIMIZE_UTILIZATION"
    max_cpu_cores       = 10000
    min_cpu_cores       = 0
    max_memory_gb       = 32000
    min_memory_gb       = 0
    disk_size           = 100
    disk_type           = "pd-balanced"
    auto_upgrade        = true
    auto_repair         = true
    gpu_resources       = []
  }

  timeouts = {
    create = "45m"
    update = "45m"
    delete = "45m"
  }

  cluster_resource_labels = {
    managed-by-terraform = true
  }
}
```

### Terraform Version

```sh
Terraform v1.9.5
```

### Additional information

_No response_

wyardley:
Set `master_ipv4_cidr_block` to `null` when `private_endpoint_subnetwork` is set, as the two conflict with an API level error being thrown:
`Error: googleapi: Error 400: When masterIpv4Cidr is set, privateEndpointSubnetwork must be unset.`

Fixes #2119

This would have the effect of silently ignoring the setting even when `master_ipv4_cidr_block` is explicitly set. A better fix might be to throw an error, but given that there's a baked in default and that terraform module validation can't look at the values of other variables, I don't think that would be trivial.

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -597,7 +597,7 @@ resource "google_container_cluster" "primary" {
     content {
       enable_private_endpoint     = private_cluster_config.value.enable_private_endpoint
       enable_private_nodes        = private_cluster_config.value.enable_private_nodes
-      master_ipv4_cidr_block      = private_cluster_config.value.master_ipv4_cidr_block
+      master_ipv4_cidr_block      = var.private_endpoint_subnetwork == null ? private_cluster_config.value.master_ipv4_cidr_block : null
       private_endpoint_subnetwork = private_cluster_config.value.private_endpoint_subnetwork
       dynamic "master_global_access_config" {
         for_each = var.master_global_access_enabled ? [var.master_global_access_enabled] : []
--- a/modules/beta-autopilot-private-cluster/cluster.tf
+++ b/modules/beta-autopilot-private-cluster/cluster.tf
@@ -295,7 +295,7 @@ resource "google_container_cluster" "primary" {
     content {
       enable_private_endpoint     = private_cluster_config.value.enable_private_endpoint
       enable_private_nodes        = private_cluster_config.value.enable_private_nodes
-      master_ipv4_cidr_block      = private_cluster_config.value.master_ipv4_cidr_block
+      master_ipv4_cidr_block      = var.private_endpoint_subnetwork == null ? private_cluster_config.value.master_ipv4_cidr_block : null
       private_endpoint_subnetwork = private_cluster_config.value.private_endpoint_subnetwork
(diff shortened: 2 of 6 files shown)


apeabody (approved):

PR #2132, opened 2024-10-08
fix: allow changing `enable_identity_service` value in place

### LINKED ISSUE #2129: enable_identity_service ignores false once true
### TL;DR

The input _enable_identity_service_ will enable the GKE (anthos) identity service if set to true.   Once enabled, the identity service option will not trigger once set to false 

### Expected behavior

The boolean input option enable_identity_service should enable when **true** and disable when **false**.  

### Observed behavior

During cluster create, the _enable_identity_service_ option checks for boolean value and proceeds as expected to either:

1. if true, enable the GKE identity service
2. if false, not enable the service or
3. if not set, not enable the service as the default is also false

Once enabled, modifying _enable_identity_service_ to false creates **no action or change** when planning or applying and the Identity Service remains active on the GKE cluster.

### Terraform Configuration

```hcl
module "gke" {
  source   = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster-update-variant"
  version  = "31.1.0"
  ...

  enable_identity_service = false
  ...
}
```

### Terraform Version

```sh
$ terraform version
Terraform v1.5.7
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v5.44.1
+ provider registry.terraform.io/hashicorp/google-beta v5.44.1
+ provider registry.terraform.io/hashicorp/kubernetes v2.32.0
+ provider registry.terraform.io/hashicorp/random v3.6.3
+ provider registry.terraform.io/hashicorp/time v0.12.1
```

### Additional information

_No response_

wyardley:
Make `enable_identity_service` optional, and handle the case where the value changes by only suppressing the dynamic block if `var.enable_identity_service` is null vs. false.

Fixes #2129

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ Then perform the following commands on the root folder:
 | enable\_confidential\_nodes | An optional flag to enable confidential node config. | `bool` | `false` | no |
 | enable\_cost\_allocation | Enables Cost Allocation Feature and the cluster name and namespace of your GKE workloads appear in the labels field of the billing export to BigQuery | `bool` | `false` | no |
 | enable\_default\_node\_pools\_metadata | Whether to enable the default node pools metadata key-value pairs such as `cluster_name` and `node_pool` | `bool` | `true` | no |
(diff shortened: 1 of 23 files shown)


apeabody (approved):

PR #2140, opened 2024-10-16
Fix: correct confidential node field name

redoak666:
The field name confidential_node in dynamic block confidential_nodes at nodepool level is incorrect, it should be enabled_confidential_nodes  
dynamic "confidential_nodes" {
      for_each = lookup(each.value, "enable_confidential_nodes", null) != null ? [each.value.confidential_nodes] : []
      content {
        enabled = confidential_nodes.value
}

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -1094,7 +1094,7 @@ resource "google_container_node_pool" "windows_pools" {
     }
 
     dynamic "confidential_nodes" {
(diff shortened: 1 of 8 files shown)


apeabody (changes requested):
Thanks for the contribution @redoak666!

Can you please update the template as noted here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

redoak666:
@apeabody
updated per instruction

apeabody (approved):

apeabody (approved):
Thanks for the contribution @redoak666!

PR #2146, opened 2024-10-17
chore: split poco into example

apeabody:
Split Policy Controller into it's own example from Config Sync.

### DIFF
--- a/examples/simple_zonal_with_poco/network.tf
+++ b/examples/simple_zonal_with_poco/network.tf
@@ -1,0 +1,45 @@
+/**
+ * Copyright 2021 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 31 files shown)


examples/simple_zonal_with_poco/versions.tf
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+terraform {
+  required_providers {
+    google = {
+      source = "hashicorp/google"
```
g-awmalik, line 20: can we add a version constraint?

examples/simple_zonal_with_poco/network.tf
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright 2021 Google LLC
```
g-awmalik, line 2:
```suggestion
 * Copyright 2024 Google LLC
```

g-awmalik (approved):

examples/simple_zonal_with_poco/versions.tf
apeabody: We dropped them from most example to depend on the actual module's version constraint, however we can revisit that decision for the entire repo if there is a need.

PR #2147, opened 2024-10-17
fix: add kubelet_config to default-pool

apeabody:


wyardley (approved):

autogen/main/cluster.tf.tmpl
```diff
@@ -530,6 +530,21 @@ resource "google_container_cluster" "primary" {
         }
       }
 
+      dynamic "kubelet_config" {
```
wyardley, line 533: Note: this adds to the pool _named_ `default-pool` that's used if `remove_default_node_pool` is set.

However, you may already know this, but I think to also handle the case where `remove_default_node_pool` is `true`, you'll need a top level `node_config` block that I don't think is here currently.

autogen/main/cluster.tf.tmpl
```diff
@@ -530,6 +530,21 @@ resource "google_container_cluster" "primary" {
...
+        for_each = length(setintersection(
+          keys(each.value),
+          ["cpu_manager_policy", "cpu_cfs_quota", "cpu_cfs_quota_period", "insecure_kubelet_readonly_port_enabled", "pod_pids_limit"]
+        )) != 0 ? [1] : []
+
+        content {
+          cpu_manager_policy                     = lookup(each.value, "cpu_manager_policy", "static")
```
wyardley, line 540: the lint warning caught it, but iterator has to be `kubelet_config` (or set to something else explicitly) here vs `each`

apeabody: Yeah, this actually needs to be off `var.node_pools[0]`.

autogen/main/cluster.tf.tmpl
apeabody: Yeah, for now I want to focus on our immediate `kubelet_config` issue.  Longer term revisit everything to do with  `default_node_pool`

apeabody: Actually I see our current CI error is with `default-node-pool`, not `default-pool`.

wyardley: Right. IIRC `default-pool` is the one that does get created when `remove_default_node_pool` is unset.

Currently, I don't think there's an error with that one. But also, there's not currently a way to set _any_ `node_config` settings in `default-pool` (i.e., what's set through the top level `node_config` settings) if I'm reading the code right.

wyardley: And what's also confusing is that I believe the module is explicitly defining a nodepool called `default-pool`
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/cccabcb0eca2a7755908e8d17b40ffba87f4839e/autogen/main/cluster.tf.tmpl#L505-L506 (unless this is somehow just explicitly defining settings for the `default-pool` that is created when `remove_default_node_pool` is unset, in which case, what I said is wrong?

Then, the map of node pools to create defaults to a single one called `default-node-pool`
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/cccabcb0eca2a7755908e8d17b40ffba87f4839e/autogen/main/variables.tf.tmpl#L176-L179.

The hard-coded `var.node_pools[0]` also is confusing to me, since presumably it should behave more or less the same as the other explicitly defined node pools? Or is this section just defining the behavior for the first node pool only?

apeabody: Yeah, I'll need to experiment further to fully understand the intention.

wyardley: The provider docs do kind of explain `node_pool {}` (and clarify some of the limitations around it / urge people to define node pools separately), but I'm not totally clear how it interacts with the actual default pool.

a quick test was pretty informative:
```hcl
resource "google_container_cluster" "test" {
  name               = "test"
  location           = "us-central1-f"
  initial_node_count = 1

  node_config {
    kubelet_config {
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }
  }

  node_pool_defaults {
    node_config_defaults {
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }
  }

  deletion_protection = false
  network             = "default"
  subnetwork          = "default"
  node_pool {
    name = "default-pool"    
  }
}

```
```
│ Error: googleapi: Error 400: It's invalid to specify both cluster.node_config and a node pool. Please only provide a node pool.
```

So, I guess, ignore what I said before: the module will use `node_pool {}` instead when it's defined, and therefore, the changes we're making here will solve the use case I was thinking about, just in a different way.

wyardley: And then I assume there's some kind of defaults merge that defines all the default values for each item in `var.node_pools`. But not sure why this uses `var.node_pools[0]` if it's the first node_pool, since I think the items in `var.node_pools` at the module level go directly to a `google_container_node_pool` resource, and I don't see the first item being stripped out in `local.node_pools`.

apeabody: Yeah - I also did a quick experiment with the default GKE module and that  `default-pool` is created by default, but also has a default of 0 nodes.  So good to fix (this PR), but not the issue we observed in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2145.

The module then created the `default-node-pool` using the recommended `google_container_node_pool` resource from this default:

```
variable "node_pools" {
  type        = list(map(any))
  description = "List of maps containing node pools"

  default = [
    {
      name = "default-node-pool"
    },
  ]
}
```

So I'm currently unclear why that `default-node-pool` didn't follow the `node_pool_defaults`?  Will need to investigate further.

wyardley: Yeah - could try testing directly without using the module, but it may be theat `node_pool_defaults` only affects nodepools created after the cluster is already up and running?

apeabody: > And then I assume there's some kind of defaults merge that defines all the default values for each item in `var.node_pools`. But not sure why this uses `var.node_pools[0]` if it's the first node_pool, since I think the items in `var.node_pools` at the module level go directly to a `google_container_node_pool` resource, and I don't see the first item being stripped out in `local.node_pools`.

Yeah - Looks like the usage of `var.node_pools[0]` for `default-pool` goes WAY back, so I'm wondering if the intended usage got merged/changed over time - now it appears to be used both by the first independently defined nodepool AND if `remove_default_node_pool = false` the `default-pool`...  I'll need to follow-up internally on that history, although a good path forward is probably to more strongly encourage/default `remove_default_node_pool = true`.

apeabody: > Yeah - could try testing directly without using the module, but it may be theat `node_pool_defaults` only affects nodepools created after the cluster is already up and running?

Agreed, I think the next step is to manually (with and without Terraform) replicate, although the `default_node_pool` (unlike `default-pool`) should/is being creating after the cluster:
```
module.gke.google_container_cluster.primary: Still creating... [4m20s elapsed]
module.gke.google_container_cluster.primary: Creation complete after 4m29s [id=projects/{}/locations/us-east4-a/clusters/gke-on-vpc-cluster]
module.gke.google_container_node_pool.pools["default-node-pool"]: Creating...
module.gke.google_container_node_pool.pools["default-node-pool"]: Still creating... [10s elapsed]
```

PR #2149, opened 2024-10-18
fix: fix typo in kube-dns-upstream-nameservers configmap

MaximF:
Fix typo for a configmap resource name from `kube-dns-upstream-namservers` to `kube-dns-upstream-nameservers`.

### DIFF
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -69,7 +69,7 @@ EOF
 }
 
 # Fix the name typo in the previous ConfigMap creation call
-moved = {
+moved {
   from = kubernetes_config_map_v1_data.kube-dns-upstream-namservers
   to   = kubernetes_config_map_v1_data.kube-dns-upstream-nameservers
 }
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -44,7 +44,7 @@ EOF
   ]
 }
 
-resource "kubernetes_config_map_v1_data" "kube-dns-upstream-namservers" {
+resource "kubernetes_config_map_v1_data" "kube-dns-upstream-nameservers" {
   count = !local.custom_kube_dns_config && local.upstream_nameservers_config ? 1 : 0
 
   metadata {
--- a/autogen/main/dns.tf.tmpl
+++ b/autogen/main/dns.tf.tmpl
@@ -68,6 +68,12 @@ EOF
   ]
 }
 
+# Fix the name typo in the previous ConfigMap creation call
+moved = {
+  from = kubernetes_config_map_v1_data.kube-dns-upstream-namservers
+  to   = kubernetes_config_map_v1_data.kube-dns-upstream-nameservers
+}
+
 resource "kubernetes_config_map_v1_data" "kube-dns-upstream-nameservers-and-stub-domains" {
   count = local.custom_kube_dns_config && local.upstream_nameservers_config ? 1 : 0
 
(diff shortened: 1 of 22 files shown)


apeabody (changes requested):
Thanks for the contribution @MaximF!

Please update the template so this is also changed in the sub-modules: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

MaximF:
@apeabody  Thanks for referencing. I made the change to the template and regenerated all the dependencies.
Please take a look.

apeabody:
Hi @MaximF - What will be the impact to current users?  Would it make sense to include: https://developer.hashicorp.com/terraform/language/v1.3.x/moved

MaximF:
Thanks @apeabody. I just added a `moved` block there.

autogen/main/dns.tf.tmpl
```diff
@@ -68,6 +68,12 @@ EOF
   ]
 }
 
+# Fix the name typo in the previous ConfigMap creation call
+moved = {
```
apeabody, line 72: I think their might be a typo in the terraform docs, pretty sure this is a block:

```suggestion
moved {
```

autogen/main/dns.tf.tmpl
```diff
@@ -44,7 +44,7 @@ EOF
   ]
 }
 
-resource "kubernetes_config_map_v1_data" "kube-dns-upstream-namservers" {
+resource "kubernetes_config_map_v1_data" "kube-dns-upstream-nameservers" {
```
wyardley, line 47: @apeabody I know a couple other resources in this file are using kebab case as well, but is it worth switching it to snake case at the same time, since it's being adjusted?
Most of the resources in this module use snake case, and it is recommended most of the time (including [here](https://cloud.google.com/docs/terraform/best-practices/general-style-structure#naming-convention).

apeabody: Hey @wyardley! As this is a correction to an existing name I'm fine with either, but yes, for new resources we should try to use snake case.

autogen/main/dns.tf.tmpl
MaximF: Updated on our end.
Funny enough, their source code has the right reference: https://github.com/hashicorp/terraform/blob/a384d2daa7fa7dfcf891c634c26de4a57ee4c963/website/docs/language/moved.mdx#complete-configuration.
But the published website has `moved = {`.

wyardley: https://developer.hashicorp.com/terraform/language/moved seems to have the right syntax.

autogen/main/dns.tf.tmpl
MaximF: @wyardley, happy to see you! :)

While I generally agree that we should all adhere to the same snake_case convention, I believe it would be better to address this consistently - either within this file or across the entire module. This might even warrant a dedicated `moved.tf` file of its own.
Updating a single variable from _kebab-case_ to _snake_case_ likely isn't worth the effort.

What do you guys think?

autogen/main/dns.tf.tmpl
MaximF: Ah, we were looking at _v1.3_ page, while the latest is _v1.9_!

apeabody: Good catch! - Guess it was fixed later, but they forgot to update across all the doc versions.

apeabody: FYI: This module currently requires TF >=1.3, so we typically reference the v1.3 docs to ensure we aren't including a feature which requires a later version.

autogen/main/dns.tf.tmpl
apeabody: Let's leave as-is given the current mix in the module for a future PR to update everything.

apeabody (approved):
Thanks for the contribution @MaximF!

PR #2151, opened 2024-10-23
fix: update resource names to snake_case

wyardley:
- Update resource names to snake_case
- Add `moved` blocks for resources and consolidate into `moved.tf`

I considered keeping `-` as a separator between the two parts for the `google_project_iam_member` resources, but ended up landing on just keeping everything snake case.

Followup to #2149

### DIFF
--- a/autogen/main/masq.tf.tmpl
+++ b/autogen/main/masq.tf.tmpl
@@ -19,7 +19,7 @@
 /******************************************
   Create ip-masq-agent confimap
  *****************************************/
-resource "kubernetes_config_map" "ip-masq-agent" {
+resource "kubernetes_config_map" "ip_masq_agent" {
   count = var.configure_ip_masq ? 1 : 0
 
   metadata {
--- a/autogen/main/moved.tf.tmpl
+++ b/autogen/main/moved.tf.tmpl
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 # Fix the name typo in the previous ConfigMap creation call
 moved {
   from = kubernetes_config_map_v1_data.kube-dns-upstream-namservers
--- a/autogen/main/moved.tf.tmpl
+++ b/autogen/main/moved.tf.tmpl
@@ -32,6 +32,11 @@ moved {
 }
 
 moved {
+  from = kubernetes_config_map.ip-masq-agent
+  to   = kubernetes_config_map.ip_masq_agent
+}
+
+moved {
   from = google_project_iam_member.cluster_service_account-nodeService_account
   to   = google_project_iam_member.cluster_service_account_node_service_account
 }
--- a/autogen/main/moved.tf.tmpl
+++ b/autogen/main/moved.tf.tmpl
@@ -1,0 +1,46 @@
+# Fix the name typo in the previous ConfigMap creation call
+moved {
+  from = kubernetes_config_map_v1_data.kube-dns-upstream-namservers
+  to   = kubernetes_config_map_v1_data.kube-dns-upstream-nameservers
+}
+
+# Updates for kebab to snake case, to match best practices and Google style.
+moved {
+  from = kubernetes_config_map_v1_data.kube-dns
+  to   = kubernetes_config_map_v1_data.kube_dns
+}
+
+moved {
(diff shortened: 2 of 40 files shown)


wyardley:
@apeabody I checked, but may want to triple-check the `moved {}` blocks to make sure there are no typos.

I didn't do any functional testing, like bringing this up with the old version of the module, then updating to the new version.

autogen/main/moved.tf.tmpl
```diff
@@ -0,0 +1,62 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+# Fix the name typo in the previous ConfigMap creation call
+moved {
+  from = kubernetes_config_map_v1_data.kube-dns-upstream-namservers
+  to   = kubernetes_config_map_v1_data.kube-dns-upstream-nameservers
```
apeabody, line 20: I wonder if this should be direct to `kube_dns_upstream_nameservers`?  No sure if they support "chaining"?

Also very likely we won't have a release with the `kube-dns-upstream-nameservers` anyway.

apeabody:
> @apeabody I checked, but may want to triple-check the `moved {}` blocks to make sure there are no typos.
> 
> I didn't do any functional testing, like bringing this up with the old version of the module, then updating to the new version.

Yeah - Prior to merging we should do a quick migration test (create an example with tf apply using the previous version, then tf apply with this version)

autogen/main/moved.tf.tmpl
wyardley: Fair. I'd assumed it would work since the "from" is different, so it would just relocate either that was found?

If you're sure they're going out together, I'll update it now to just have the one. 👍 
cc: @MaximF

wyardley:
Are you setup to run that easily? I'm not sure if I'll be setup to test it quickly, though I can try if you need me to.

question: is there a way to turn on the `tflint` rule for resource naming specific to this module somehow, to prevent new resources getting created with upper case letters or hyphens? I think this is mostly using Google's internal and somewhat minimal tflint config, right?

autogen/main/moved.tf.tmpl
apeabody: What I don't know is if the move rules are "chained", if they are not and this rule fires, it will just try to move to `kube-dns-upstream-nameservers`, which of course no longer exists.  As this PR also adds `kube_dns_upstream_nameservers`, regardless it should be perfectly safe to go directly to that resource.

apeabody:
> Are you setup to run that easily? I'm not sure if I'll be setup to test it quickly, though I can try if you need me to.
> 
> question: is there a way to turn on the `tflint` rule for resource naming specific to this module somehow, to prevent new resources getting created with upper case letters or hyphens? I think this is mostly using Google's internal and somewhat minimal tflint config, right?

:) I was also thinking if we can use `tflint`, as I believe this rule can be configured to work: https://github.com/terraform-linters/tflint-ruleset-terraform/blob/main/docs/rules/terraform_naming_convention.md

This default tflint config for modules can be found here: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/home/.tflint.module.hcl

However I also recently added the option for local repo tflint configs, especially given the large number of repos that would likely also need to be updated first: https://github.com/GoogleCloudPlatform/terraform-google-enterprise-application/blob/main/.github/.tflint.repo.hcl



As far as testing, I was planning to run a few examples either tomorrow or Friday.

autogen/main/masq.tf.tmpl
```diff
@@ -19,7 +19,7 @@
 /******************************************
   Create ip-masq-agent confimap
  *****************************************/
-resource "kubernetes_config_map" "ip-masq-agent" {
+resource "kubernetes_config_map" "ip_masq_agent" {
```
apeabody, line 22: Hey @wyardley - I don't see a moved for `ip_masq_agent`?

wyardley: Good call, I probably missed that. Thanks for checking.

wyardley: Updated

MaximF:
@wyardley Thanks for the pushing `moved.tf` changes so promptly! 👍 
@apeabody I like the idea of using `tflint` for enforcing the _snake_style_ naming usage!

apeabody:
```
  # module.gke.kubernetes_config_map_v1_data.kube-dns-upstream-namservers[0] has moved to module.gke.kubernetes_config_map_v1_data.kube_dns_upstream_nameservers[0]
    resource "kubernetes_config_map_v1_data" "kube_dns_upstream_nameservers" {
        id            = "kube-system/kube-dns"
        # (3 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

wyardley:
BTW, we can update the commit prefix to `refactor` if you'd rather it not show up in the changelog - it in theory shouldn't affect end-users, but I guess may be good to call it out in the changelog just in case?

apeabody:
> BTW, we can update the commit prefix to `refactor` if you'd rather it not show up in the changelog - it in theory shouldn't affect end-users, but I guess may be good to call it out in the changelog just in case?

When in doubt I prefer to include in the release notes.  If end users do have resources that will be migrated they will see the "moved" in their first subsequent apply, although no changes to the actual resources.

apeabody (approved):

PR #2156, opened 2024-10-25
chore(CI): add new verify

apeabody:


test/integration/utils/utils.go
```diff
@@ -32,3 +35,19 @@ func GetTestProjectFromSetup(t *testing.T, idx int) string {
 	}
 	return projectIDs[idx]
 }
+
+var (
+	// TGKEVerify Exempt Resources. e.g. google_container_cluster.primary
+	verifyExemptResources = []string{}
+)
+
+func TGKEVerify(t *testing.T, b *tft.TFBlueprintTest, assert *assert.Assertions) {
```
bharathkkb, line 44: Would it make sense to allow the call site to provide the slice of address to ignore? We can export `verifyExemptResources` as a starting point.

test/integration/utils/utils.go
```diff
@@ -32,3 +35,19 @@ func GetTestProjectFromSetup(t *testing.T, idx int) string {
...
+	_, ps := b.PlanAndShow()
+	for _, r := range ps.ResourceChangesMap {
+		if slices.Contains(verifyExemptResources, r.Address) {
```
bharathkkb, line 47: Should we consider matching based on suffix (so we can provide a snippet like `google_container_cluster.primary`) otherwise we will to write the full address like `module.example.module.gke.google_container_cluster.primary`

bharathkkb (approved):
Thanks @apeabody!

test/integration/utils/utils.go
apeabody: Added `TGKEVerifyExemptResources()`, LMKWYT

test/integration/utils/utils.go
apeabody: Sure, I think the combo of `slices.ContainsFunc() and `strings.HasSuffix()` should do it.

bharathkkb (approved):

PR #2161, opened 2024-10-29
chore: update vpc module in example

wyardley:
Update the version spec of the VPC module in `safer_cluster_iap_bastion`.

It is likely already getting the newer version already since it's `>=`

~Note: I'm not sure why Renovate isn't PRing updating it, unless it's the `google/` in the module's path? Based on the registry, this seems like the correct way to define that specific top level module, though.~ Oh, maybe just because it's `>=` vs `~=`. Let me know if you want me to make the version spec more specific.


apeabody:
> Update the version spec of the VPC module in `safer_cluster_iap_bastion`.
> 
> It is likely already getting the newer version already since it's `>=`
> 
> ~Note: I'm not sure why Renovate isn't PRing updating it, unless it's the `google/` in the module's path? Based on the registry, this seems like the correct way to define that specific top level module, though.~ Oh, maybe just because it's `>=` vs `~=`. Let me know if you want me to make the version spec more specific.

Yes - With `>=` it will automatically grab the newest version during init, so Renovate won't try to update.  If we are going to manually bump the version, let's also switch to `~>`.

apeabody (approved):

PR #2142, opened 2024-10-17
feat: Add create_ignore_already_exists to workload_identity

flozzone:
Adds [create_ignore_already_exists](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account#create_ignore_already_exists) to workload-identity module to ignore an already existing google-service-account when hitting the [Provider produced inconsistent result after apply](https://github.com/hashicorp/terraform-provider-google/issues/18087) issue.

Fixes #2141

Issue #2141, opened 2024-10-17
Service account: Provider produced inconsistent result after apply


flozzone:
### TL;DR

Sometimes creating GCP service accounts can fail with Terraform due to GCPs eventual consistency model (See [terraform-provider-google#18087](https://github.com/hashicorp/terraform-provider-google/issues/18087)). 
After running apply again, it fails because the service account already exists. 

A possible workaround has been added to [Terraform/google_service_account](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account#create_ignore_already_exists) module to ignore an already existing service account by passing `create_ignore_already_exists=true` when creating the `google_service_account`

### Expected behavior

Service account gets created or at least it would take an already existing one if running apply again

### Observed behavior

Terraform complains about an already existing GCP service account

### Terraform Configuration

```hcl
module "app_workload_identity" {
  source              = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
  name                = "some-sa-name"
  namespace           = "some-ns"
  project_id          = "some-project"
  roles               = []
}
```

### Terraform Version

```sh
Terraform v1.8.4
on linux_amd64
```

PR #2170, opened 2024-11-01
fix(TPG>=5.12.0): Bump for #2142 (#2141)

flozzone:
#2142 introduced `gcp_sa_create_ignore_already_exists` variable for workload-identity module without bumping to TPG<=5.12.0 and fails on installations with older TPG.


apeabody (approved):
Thanks for the contribution @flozzone!

PR #2155, opened 2024-10-25
feat: add logging_config and monitoring_config to autopilot modules

pocesar:
Matches the other submodules inputs

PR #2188, opened 2024-11-18
feat!: Change the default value of "monitoring_enable_managed_prometheus" var to null

legal90:
Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1706

~This PR is a continuation of https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/1715 by removing the condition for `monitoring_config` and turning it into a "static" block.~

This PR is a continuation of https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2155. 
Here we change the default value of the variable `monitoring_enable_managed_prometheus` to `null`.

That is supposed to fix the calculation of `local.logmon_config_is_set`. Now it will return `true` no only when `monitoring_enable_managed_prometheus` is set to `true`, but also when it is set to `false`.


legal90:
@apeabody Thank you! I will take a look at it soon.

legal90:
@apeabody I rolled back my previous changes and implemented the fix using a different approach - by changing the default value of `"monitoring_enable_managed_prometheus"` var to `null`.

Could you please re-trigger the integration test?

apeabody (approved):

PR #2195, opened 2024-11-26
feat(update-variant)!: allow updating disk_type, disk_size_gb, machine_type, enable_gcfs in place

DrFaust92:
this a breaking change as this will cause a recreate of the node pools once more but subsequent updates will be in place

This is a huge QOL change to allow rolling out machine type changes without node pool recreate and allowing apps to migrate gracefully

upstream ref: https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/node_config.go#L1937


apeabody:
Verified this is in [v6.8.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v6.8.0)+

apeabody (changes requested):
Thanks for the contriobution @DrFaust92!

Can you please add something to `docs/upgrading_to_v35.0.md` similar to: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/43d1c566c30cc8a799540a5be412362e6dea0aa9/docs/upgrading_to_v29.0.md?plain=1#L40

DrFaust92:
Thanks [apeabody](https://github.com/apeabody)! added the section in the upgrade guide

apeabody (approved):

PR #2202, opened 2024-12-06
fix!: standardize to >= TF v1.3

apeabody:


bharathkkb (commented):
@apeabody should we release this as breaking?

apeabody:
> @apeabody should we release this as breaking?

Sure (it's going into a breaking release anyway). - The main and most sub-module are already v1.3, just a few sub-modules were not.

bharathkkb (approved):

PR #2203, opened 2024-12-07
feat: support parallelstore csi driver

DrFaust92:


autogen/main/variables.tf.tmpl
```diff
@@ -838,6 +838,12 @@ variable "stateful_ha" {
   default     = false
 }
 
+variable "parallelstore_csi_driver" {
+  type        = bool
+  description = "Whether the Parallelstore CSI driver Addon is enabled for this cluster."
+  default     = false
```
apeabody, line 844: As there is a dynamic block using a `for_each`, could the default be made `null` so this is optional?

autogen/main/cluster.tf.tmpl
```diff
@@ -388,6 +388,14 @@ resource "google_container_cluster" "primary" {
       }
     }
 
+    dynamic "parallelstore_csi_driver_config" {
```
apeabody, line 391: I don't think we need to add for Autopilot clusters as for them it's enabled by default?

apeabody (changes requested):

autogen/main/cluster.tf.tmpl
DrFaust92: will check

autogen/main/variables.tf.tmpl
DrFaust92: will check

autogen/main/cluster.tf.tmpl
DrFaust92: Screenshot 2024-12-16 at 18 18 28

Only found this as ref, will go with not setting this on autopilot for now

DrFaust92:
> From the INT test:
> 
> ```
> Error: Null condition
> 
>   on ../../../main.tf line 94, in locals:
>   94:   parallelstore_csi_driver_config = var.parallelstore_csi_driver ? [{ enabled = true }] : []
>     ├────────────────
>     │ var.parallelstore_csi_driver is null
> 
> The condition value is null. Conditions must either be true or false.}
> ```

I think i fixed it now, i addressed a previous comment but didnt address the nullability

apeabody (approved):
Thanks @DrFaust92!

PR #2208, opened 2024-12-10
fix: min_cpu_platform ref in cluster autoscaling

DrFaust92:


apeabody (commented):
Thanks @DrFaust92 - Could we had a test to confirm this is now working as expected and prevent regression?

apeabody (approved):

PR #2214, opened 2024-12-13
feat: add cgroup & sysctls to safe-cluster & safe-cluster-update-variant

j:


apeabody (changes requested):
Thanks for the contribution @j!

Can you please update the template(s) and regenerate the resulting docs.  Thanks!

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

j:
@apeabody pushed!

apeabody:
Hi @j - Thanks again for the contribution!

I've discussed this proposed change with the larger team, and the decision is to instead recommend using the `beta-private-cluster` or `beta-private-cluster-update-variant` modules.  The purpose of the Safer Cluster modules are to only present a limited set of parameters in-line with the [GKE hardening guide](https://cloud.google.com/kubernetes-engine/docs/how-to/hardening-your-cluster), the CIS framework, and other best practices.  The `beta-private-cluster` or `beta-private-cluster-update-variant` modules provide the full range of parameters, and can also be configured in the same fashion as the Safer Cluster modules as needed.

Best Wishes!

PR #2217, opened 2024-12-16
fix!: allow dns endpoints only

### LINKED ISSUE #2216: Unable to use the new DNS endpoint exclusively.
### TL;DR

My goal is to use only the DNS endpoint on my GKE clusters:
image

This is not possible at the moment because I need to set:

```
  enable_private_endpoint       = true
  deploy_using_private_endpoint = true
```

for the DNS endpoint to be enabled. However, `enable_private_endpoint` will create an empty `master_authorized_networks_config {}` block in [this line](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/9a86b4a894dd93364550fd4194e6f2848d30e8f9/modules/beta-private-cluster-update-variant/cluster.tf#L217), and applying this will re-enable access via IPv4.
image

I created [a PR to fix this](#2217), but I'm not 100% sure if it breaks anything unrelated to DNS endpoints.

Another thing I observed while experimenting with this setting is that removing `master_authorized_networks_config {}` won't actually change my cluster to disable IPv4 access. The code seems to work for both cases: with `master_authorized_networks_config {}` and IPv4 enabled, and omitted with IPv4 disabled. Maybe this is just a limitation in the API, where changing `master_authorized_networks_config {}` to `(empty)` does not trigger an update. 🤔

### Expected behavior

_No response_

### Observed behavior

_No response_

### Terraform Configuration

```hcl
n.a.
```

### Terraform Version

```sh
tofu version                                                                                          app-publisher-dev-7ed19f10be
OpenTofu v1.8.7
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v4.84.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.23.0
+ provider registry.terraform.io/hashicorp/random v3.5.1
```

### Additional information

_No response_

TheKangaroo:
fixes #2216


autogen/main/cluster.tf.tmpl
```diff
@@ -266,7 +266,7 @@ resource "google_container_cluster" "primary" {
   enable_autopilot = true
 {% endif %}
   dynamic "master_authorized_networks_config" {
-    for_each = {% if private_cluster %}var.enable_private_endpoint || {% endif %}var.gcp_public_cidrs_access_enabled != null || length(var.master_authorized_networks) > 0 ? [true] : []
+    for_each = {% if private_cluster %}(var.enable_private_endpoint && !var.deploy_using_private_endpoint) || {% endif %}var.gcp_public_cidrs_access_enabled != null || length(var.master_authorized_networks) > 0 ? [true] : []
```
apeabody, line 269: Based on https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/resource_container_cluster.go#L5035, I'm thinking we should probably gate on `master_authorized_networks` being not `null`.  That should more properly differentiate between `null` (no `master_authorized_networks_config` block, hence disable) and an empty list `[]` (with `master_authorized_networks_config` block enabled, but no `cidr_blocks`):

```suggestion
    for_each = var.master_authorized_networks != null ? [true] : []
```

apeabody (changes requested):
Thanks for the contribution @TheKangaroo!

autogen/main/cluster.tf.tmpl
TheKangaroo: I updated the PR. This works for me; however, I have to explicitly unset `master_authorized_networks` because the default is `[]`, but that's okay.

```
enable_private_endpoint       = true
deploy_using_private_endpoint = true
master_authorized_networks    = null
```

apeabody: Yeah - Ideally we would change the `master_authorized_networks` default to `null` and update the description, but that would be a breaking change.  That can still be done in the future, but we should be able to ship this as just a fix for now.

apeabody: Actually @TheKangaroo this will also result in a breaking change of behavior for clusters which didn't have one of the three previously set `var.enable_private_endpoint || var.gcp_public_cidrs_access_enabled != null || length(var.master_authorized_networks) > 0`, as `master_authorized_networks_config {}` would newly be created for them.

I think we have a couple options:
- Update the `master_authorized_networks` default to `null`. (breaking change, but following the TPG behavior) <- Probably the "best" option?
- Add a new `master_authorized_networks_enabled` which defaults to `true` and would instead be used to gate this block. (more closely follow the API behavior) <- Prefer not to add new variables if possible.
- Something else?

TheKangaroo: I don't think I need to add an explanation in the variable description on why the default is null, but we should include an explanation in the release notes/update docs. However, I do not know where the release-bot will pull this information from. Is it from the PR description or from the commit message?

TheKangaroo: @apeabody I added an upgrade note to the PR, basing it on the default change notice from the last major release.

apeabody:
Hi @TheKangaroo - We could likely resolve this error with `for_each = {% if private_cluster %}var.enable_private_endpoint || {% endif %} var.master_authorized_networks != null ? [true] : []`, however my impression from https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2216 is your use case is specifically `private_endpoint = true` with no `master_authorized_networks_config` block?  If so, it appears that combination might not be possible?

```
TestSimpleAutopilotPrivate 2024-12-18T00:49:44Z command.go:185: module.example.module.gke.google_container_cluster.primary: Creating...
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: 
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: Error: googleapi: Error 400: 
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: 	- invalid value for "cluster.master_authorized_networks_config": "cluster.master_authorized_networks_config.enable_master_authorized_networks" should be enabled if private endpoint is enabled
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: 	- master authorized networks must be enabled if private endpoint is enabled.
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: Details:
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: [
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185:   {
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185:     "@type": "type.googleapis.com/google.rpc.RequestInfo",
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185:     "requestId": "0x31e37f9a8468168d"
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185:   }
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: ]
TestSimpleAutopilotPrivate 2024-12-18T00:49:47Z command.go:185: , badRequest
```

TheKangaroo:
@apeabody You are right that that's exactly what I am trying to achieve: a private endpoint cluster (with a DNS endpoint) and no `master_authorized_networks_config`. I can plan/apply my cluster with this feature branch, so it seems to be fine with the beta-private-cluster-update-variant with:

```
enable_private_endpoint       = true
deploy_using_private_endpoint = true
```

to enable the DNS endpoint. To be honest, I don't fully understand which combinations of variables are possible and supported by the modules and the API. The only things I can imagine are:

* This works differently on autopilot, and I need to exclude autopilot in the config.
* Omitting `master_authorized_networks_config` is only allowed if the DNS endpoint is enabled.

apeabody:
> I can plan/apply my cluster with this feature branch, so it seems to be fine with the beta-private-cluster-update-variant with:
> 
> ```
> enable_private_endpoint       = true
> deploy_using_private_endpoint = true
> ```
> 
> to enable the DNS endpoint.

Thanks @TheKangaroo - I [suspect](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2216#issuecomment-2555270528) this resulted in the IP Endpoints enabled?  Unfortunately, I suspect we need https://github.com/hashicorp/terraform-provider-google/issues/20369 to "cleanly" implement your goal.  This is because currently setting `enable_private_endpoint = true` [results](https://github.com/hashicorp/terraform-provider-google/blame/v6.14.1/google/services/container/resource_container_cluster.go#L5154) in `IPEndpointsConfig. EnablePublicEndpoint  = false` (but with hardcoded `IPEndpointsConfig.Enabled = true`).

TheKangaroo:
Yes, this resulted in IP endpoints being enabled, but with my PR, without `master_authorized_networks_config` configured, I can disable IP endpoints in the UI, and Terraform won't complain.

I read through your explanation in #2216, and I'll continue working on this PR when the provider supports disabling IPEndpointsConfig.

apeabody:
> Yes, this resulted in IP endpoints being enabled, but with my PR, without `master_authorized_networks_config` configured, I can disable IP endpoints in the UI, and Terraform won't complain.
> 
> I read through your explanation in #2216, and I'll continue working on this PR when the provider supports disabling IPEndpointsConfig.

Hi @TheKangaroo - This may now be possible with https://github.com/hashicorp/terraform-provider-google/releases/tag/v6.27.0 adding `ip_endpoints_config`.

apeabody:
Hi @TheKangaroo - Just checking if you are still working on this PR.  Thanks!

TheKangaroo:
Hi @apeabody,

I'm a bit confused about the current state of this module's inputs regarding DNS endpoints, authorised networks, IP endpoints, etc.
I'm not sure if this PR is still necessary, since merging #2380 should have resolved the associated issue.
I haven't had time to test the new inputs recently, but I'll do so and open a new issue/pull request if necessary.

apeabody:
> Hi @apeabody,
> 
> I'm a bit confused about the current state of this module's inputs regarding DNS endpoints, authorised networks, IP endpoints, etc. I'm not sure if this PR is still necessary, since merging #2380 should have resolved the associated issue. I haven't had time to test the new inputs recently, but I'll do so and open a new issue/pull request if necessary.

Thanks @TheKangaroo! - You are correct this thread pre-dates #2380.  I'll close this PR, and please open a new one if needed.

PR #2218, opened 2024-12-17
fix: remove fuse csi driver from autopilot

DrFaust92:
Fuse CSI Driver is not supported for autopilot


apeabody:
This can be included in v35.0.1

apeabody:
> Fuse CSI Driver is not supported for autopilot

For clarify: ["The Cloud Storage FUSE CSI driver is enabled by default for Autopilot clusters."](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-storage-fuse-csi-driver-setup)

apeabody (commented):

Given the removal of the variable for Autopilot clusters, I think we'll want to classify this as a breaking change.  Thoughts?  Otherwise this LGTM.

DrFaust92:
[apeabody](https://github.com/apeabody) Im fine with as i prefer err on the side of caution. but I dont think its breaking as it not supported toggle anyway

apeabody:
> [apeabody](https://github.com/apeabody) Im fine with as i prefer err on the side of caution. but I dont think its breaking as it not supported toggle anyway

Agreed - It's an usual edge case as the cluster "behavior" doesn't change, but if present they would need to remove `gcs_fuse_csi_driver = true`.

apeabody (approved):
Thanks @DrFaust92! For caution we'll wait to include a major version, but otherwise LGTM.

PR #2228, opened 2024-12-30
feat: add enable_secret_manager_addon to safer-cluster & safer-cluster-update-variant

ecrousseau:


apeabody (changes requested):
Thanks for the contribution @ecrousseau!

Please also update the module templates: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/master/CONTRIBUTING.md#templating

PR #2234, opened 2025-01-08
chore(CI): migrate tests to cft

apeabody:
- stub_domains
- stub_domains_private (still disabled for now)
- upstream_nameservers


test/integration/stub_domains/stub_domains_test.go
```diff
@@ -0,0 +1,103 @@
...
+			golden.WithSanitizer(golden.StringSanitizer(randomString, "RANDOM_STRING")),
+			golden.WithSanitizer(golden.StringSanitizer(kubernetesEndpoint, "KUBERNETES_ENDPOINT")),
+		)
+
+		// Cluster Assertions
+		testutils.TGKEAssertGolden(assert, g, &cluster, []string{}, []string{"monitoringConfig.componentConfig.enableComponents"}) // TODO: enableComponents is UL
+
+		// K8s Assertions
+		// CAI does not include k8s.io/ConfigMap
+		gcloud.Runf(t, "container clusters get-credentials %s --region %s --project %s", clusterName, location, projectId)
```
bharathkkb, line 70: Wondering if running this can cause current context to change incase of parallel test executions which also does get-creds? if its an issue, maybe we can pass explicit context in KubectlOptions?

bharathkkb (approved):

test/integration/stub_domains/stub_domains_test.go
apeabody: It likely possible, although I don't believe I've ever seen it happen.  Regardless let's try and generate/pass-in the cluster's context.

apeabody: Looks like that worked, I'll back port to existing tests in my next PR.

PR #2236, opened 2025-01-10
feat!: remove ACM sub-module

apeabody:


docs/upgrading_to_v36.0.md
```diff
@@ -25,6 +25,91 @@ The variable `logging_variant` has been removed from the Autopilot sub-modules.
...
+-  enable_referential_rules = true
+-  policy_bundles = ["https://github.com/GoogleCloudPlatform/acm-policy-controller-library.git/bundles/pss-baseline-v2022"]
+-}
+
+# Config Sync
++resource "google_gke_hub_feature" "config_feature" {
++  project  = var.project_id
++  location = "global"
++  name     = "servicemesh"
++}
```
janetkuo, line 59: Should it be something like this:
https://github.com/terraform-google-modules/terraform-docs-samples/blob/main/gke/autopilot/config_sync/git/main.tf#L35-L38

apeabody: Thanks @janetkuo! Yes, I based this off a migration guide I previously wrote for ASM (prior create the sample you linked) and clearly missed updating this name field.

docs/upgrading_to_v36.0.md
```diff
@@ -25,6 +25,91 @@ The variable `logging_variant` has been removed from the Autopilot sub-modules.
...
++  feature             = google_gke_hub_feature.config_feature.name
++  membership          = module.gke.fleet_membership
++  membership_location = module.gke.region
+
++  configmanagement {
++    config_sync {
++      enabled = true
++      git {
++        sync_repo = "git@github.com:GoogleCloudPlatform/anthos-config-management-samples.git"
++        sync_branch = "1.0.0"
```
janetkuo, line 74: nit: https://github.com/GoogleCloudPlatform/anthos-config-management-samples/ doesn't have 1.0.0 branch anymore. Should we use a branch that actually exists, such as `main`?

janetkuo (approved):

docs/upgrading_to_v36.0.md
apeabody: Thanks @janetkuo! Sure, in theory this shouldn't be used as-is, but we can certainly switch to `main` which I see if the default branch.

g-awmalik (approved):

PR #2245, opened 2025-01-15
fix: node pool creation should depend on necessary firewall rules

### LINKED ISSUE #2243: `google_container_node_pool.pools` should depend on `google_compute_firewall.intra_egress`
### TL;DR

We have a default-deny egress rule enforced. Without the dependency declaration, a node pool will be created first and if a initial node count bigger than zero is specified, it will never get ready because the created node cannot talk to the control plane master.

We met this issue after upgrading the google provider to `v6.15.0`. My guess is it's caused by https://github.com/hashicorp/terraform-provider-google/pull/20738, which limits the concurrency, making the dependency issue prominent.

### Expected behavior

[google_compute_firewall.intra_egress](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/2d5999164ed27aad252ae8b24ac70396b34a194f/modules/beta-private-cluster/firewall.tf#L27) should be created before any [google_container_node_pool.pools](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/2d5999164ed27aad252ae8b24ac70396b34a194f/modules/beta-private-cluster/cluster.tf#L627) is created.

### Observed behavior

[google_compute_firewall.intra_egress](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/2d5999164ed27aad252ae8b24ac70396b34a194f/modules/beta-private-cluster/firewall.tf#L27) is not created when a [google_container_node_pool.pools](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/2d5999164ed27aad252ae8b24ac70396b34a194f/modules/beta-private-cluster/cluster.tf#L627) is being created, causing the creation being stuck.

The node registration failed with the following logs:

```
------------------------------
Service    DNS      Reachable
------------------------------
LOGGING    true     true
GCR        true     true
GCS        true     true
Master     N/A      false
------------------------------
```

I'm able to solve this issue if I create a firewall rule allowing egress to master CIDR for GKE nodes and then apply

### Terraform Configuration

```hcl
// By default deny all egress traffic -- applied to all instances
resource "google_compute_firewall" "deny_all_egress" {
  name      = "rule"
  network   = module.vpc.network_self_link
  project   = var.gcp_project_id
  direction = "EGRESS"
  priority  = 65534

  log_config {
    metadata = "INCLUDE_ALL_METADATA"
  }

  deny {
    protocol = "all"
  }

  destination_ranges = ["0.0.0.0/0"]
}

// Call this module with at least one node pool with initial count > 0
module "gke" {
  source  = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster"
  version = "35.0.1"

...

  node_pools = [
     {
      ...

      min_count          = 0
      max_count          = 100
      initial_node_count = 1

      ...
    }
  ]
}

}
```

### Terraform Version

```sh
1.5.7
```

### Additional information

_No response_

wjf3121:
fix #2243

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -1153,9 +1153,12 @@ resource "google_container_node_pool" "windows_pools" {
     delete = lookup(var.timeouts, "delete", "45m")
   }
 
-  {% if i == 1 %}
-  depends_on = [google_container_node_pool.pools[0]]
-  {% endif %}
+  depends_on = [
+    google_compute_firewall.intra_egress,
+    {% if i == 1 %}
+    google_container_node_pool.pools[0],
+    {% endif %}
+  ]
 }
 {% endfor %}
 {% endif %}
--- a/cluster.tf
+++ b/cluster.tf
@@ -845,6 +845,9 @@ resource "google_container_node_pool" "pools" {
     delete = lookup(var.timeouts, "delete", "45m")
   }
 
+  depends_on = [
+    google_compute_firewall.intra_egress,
+  ]
 }
 resource "google_container_node_pool" "windows_pools" {
   provider = google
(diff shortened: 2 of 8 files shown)


apeabody (approved):
Thanks for the contribution @wjf3121!

PR #2256, opened 2025-01-28
fix!: re-align node_config_defaults to cluster types

apeabody:
- add `logging_variant` to Standard, remove from Autopilot
- remove `enable_gcfs` from Autopilot (always enabled depending on version)
- do not create `node_pool_defaults.node_config_defaults` block on Autopilot

### DIFF
--- a/modules/beta-autopilot-private-cluster/variables.tf
+++ b/modules/beta-autopilot-private-cluster/variables.tf
@@ -394,12 +394,6 @@ variable "enable_confidential_nodes" {
   default     = false
 }
 
-variable "enable_gcfs" {
-  type        = bool
-  description = "Enable image streaming on cluster level."
-  default     = true
-}
-
 variable "enable_secret_manager_addon" {
   description = "Enable the Secret Manager add-on for this cluster"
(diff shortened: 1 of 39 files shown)


modules/beta-autopilot-private-cluster/variables.tf
```diff
@@ -394,12 +394,6 @@ variable "enable_confidential_nodes" {
   default     = false
 }
 
-variable "enable_gcfs" {
```
bharathkkb, line 397: Could we add a note in the upgrade guide about the context of these being removed?

bharathkkb (approved):

modules/beta-autopilot-private-cluster/variables.tf
apeabody: Yes - This auto-merged, but I'll open a docs only PR to do that.

apeabody: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2261

PR #2283, opened 2025-02-13
chore(CI): E2_STANDARD_2 machine type

apeabody:
With [blueprint-test v0.17.5](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/releases/tag/infra%2Fblueprint-test%2Fv0.17.5) we can move to a smaller runner

### DIFF
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -488,4 +488,4 @@ substitutions:
   _DOCKER_IMAGE_DEVELOPER_TOOLS: 'cft/developer-tools'
   _DOCKER_TAG_VERSION_DEVELOPER_TOOLS: '1.23'
 options:
-  machineType: 'E2-STANDARD-2'
+  machineType: 'E2_STANDARD_2'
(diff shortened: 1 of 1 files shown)


build/int.cloudbuild.yaml
```diff
@@ -487,5 +487,3 @@ tags:
 substitutions:
   _DOCKER_IMAGE_DEVELOPER_TOOLS: 'cft/developer-tools'
   _DOCKER_TAG_VERSION_DEVELOPER_TOOLS: '1.23'
-options:
-  machineType: 'E2_HIGHCPU_32'
```
bharathkkb, line 491: Should we revert to the previous type `N1_HIGHCPU_8` or the e2 variant? Some of these tests run in parallel stages on same host so I think it would still benefit from higher cpu

bharathkkb (approved):

build/int.cloudbuild.yaml
apeabody: The `N1_HIGHCPU_8` is deprecated, but I can test against the replacement `E2-HIGHCPU-8`.  This (`E2-STANDARD-2`) was already within 10% of the runtime with the previous `E2_HIGHCPU_32`.

apeabody: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2284

PR #2284, opened 2025-02-14
chore(CI): use E2_HIGHCPU_8 machine type

apeabody:


apeabody:
@bharathkkb - Looks like this is about the same run time as the HIGHCPU_32, and a little faster than the STANDARD_2.

apeabody:
> @bharathkkb - Looks like this is about the same run time as the `HIGHCPU_32`, and a little faster (10-15 minutes) than the default `STANDARD_2`.

bharathkkb (approved):

PR #2313, opened 2025-03-19
feat: add dns_allow_external_traffic

apeabody:
* Add new `dns_allow_external_traffic` bool to set `control_plane_endpoints_config. dns_endpoint_config.allow_external_traffic`


bharathkkb (approved):
LGTM
Regarding breaking change, I think this should be since private cluster users will now need to set `dns_allow_external_traffic` explicitly instead of the previous gating condition. Could another option be to temporarily allow both old gating condition (var.enable_private_endpoint && var.deploy_using_private_endpoint) or new var and then in the next major release we can clean up the old gating condition?

apeabody:
> LGTM Regarding breaking change, I think this should be since private cluster users will now need to set `dns_allow_external_traffic` explicitly instead of the previous gating condition. Could another option be to temporarily allow both old gating condition (var.enable_private_endpoint && var.deploy_using_private_endpoint) or new var and then in the next major release we can clean up the old gating condition?

That a great idea - I'll rework to gate on either condition for now.  Thanks!

bharathkkb (approved):

PR #2298, opened 2025-02-28
feat: add enterprise config

TheKangaroo:
fixes #2293

PR #2314, opened 2025-03-19
fix: Update files with correct values for enterprise config

dheitepriem:
The newly introduced flag for `enterprise_config` (added in #2298) can only take in the values `STANDARD` or `ENTERPRISE` as by the Google API. The mentioned value of `DEFAULT` isn't valid

- Corrected values from `DEFAULT` to `STANDARD`
- Added validation for the variables


apeabody (approved):
Thanks for the contribution @dheitepriem!

Issue #2346, opened 2025-05-09
Creating host project fleet


YpNo:
### TL;DR

To be able creating an empty fleet on a project without creating a cluster

### Terraform Resources

```markdown
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/gke_hub_fleet
```

### Detailed design

```markdown
Creating a fleet without anthos feature and without cluster creation should be possible. 
The terraform resource let configure Anthos parameters as well.
```

PR #2347, opened 2025-05-09
Fleet: Creating fleet without cluster creation

YpNo:
Creating a fleet without anthos feature and without cluster creation should be possible. 
The terraform resource let configure Anthos parameters as well.

Ref : #2346


YpNo:
Up !

PR #2354, opened 2025-05-20
fix: Invalid index for `google_container_cluster.primary.private_cluster_config[0]`

### LINKED ISSUE #2353: private-cluster module: Invalid index for `google_container_cluster.primary.private_cluster_config[0]`
### TL;DR

Am using the `private-cluster` module in the following way:

```
module "gke" {
  source  = "terraform-google-modules/kubernetes-engine/google//modules/private-cluster"
  version = "36.3.0"

  project_id = var.project_id
  name = local.cluster_name

  network = data.google_compute_network.vpc.name
  subnetwork = data.google_compute_subnetwork.subnet.name
  ip_range_pods = local.valid_pods_range_name
  ip_range_services = local.valid_services_range_name

  enable_private_nodes  = false
}
```

Running a `terraform plan` with `enable_private_nodes = false` returns the following error:

```
╷
│ Error: Invalid index
│ 
│   on .terraform/modules/gke/modules/private-cluster/main.tf line 115, in locals:
│  115:   cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
│     ├────────────────
│     │ google_container_cluster.primary.private_cluster_config is empty list of object
│ 
│ The given key does not identify an element in this collection value: the collection has no elements.
╵
```

### Expected behavior

When `enable_private_nodes = false` the `cluster_endpoint_for_nodes` local in `main.tf` should resolve to null, preventing the lookup of `google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block` and thus avoiding the error. 

When  `enable_private_nodes = false` this looked up field does not exist as can be seen in this snippet:

```
  dynamic "private_cluster_config" {
    for_each = var.enable_private_nodes ? [{
      enable_private_nodes        = var.enable_private_nodes,
      enable_private_endpoint     = var.enable_private_endpoint
      master_ipv4_cidr_block      = var.master_ipv4_cidr_block
      private_endpoint_subnetwork = var.private_endpoint_subnetwork
    }] : []

    content {
      enable_private_endpoint     = private_cluster_config.value.enable_private_endpoint
      enable_private_nodes        = private_cluster_config.value.enable_private_nodes
      master_ipv4_cidr_block      = var.private_endpoint_subnetwork == null ? private_cluster_config.value.master_ipv4_cidr_block : null
      private_endpoint_subnetwork = private_cluster_config.value.private_endpoint_subnetwork
      dynamic "master_global_access_config" {
        for_each = var.master_global_access_enabled ? [var.master_global_access_enabled] : []
        content {
          enabled = master_global_access_config.value
        }
      }
    }
  }
```

### Observed behavior

Even if  `enable_private_nodes = false` and `google_container_cluster.primary.private_cluster_config` is an empty list of object the value `google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block` is still set as the value of `local.cluster_endpoint_for_nodes` in `main.tf`.

### Terraform Configuration

```hcl
module "gke" {
  source  = "terraform-google-modules/kubernetes-engine/google//modules/private-cluster"
  version = "~> 36.3.0"

  deletion_protection     = true
  project_id              = var.project_id
  region                  = var.region
  name                    = local.cluster_name
  cluster_resource_labels = local.cluster_resource_labels

  regional = true
  zones    = var.zones

  kubernetes_version = var.kubernetes_version
  release_channel    = var.release_channel

  network                    = data.google_compute_network.vpc.name
  subnetwork                 = data.google_compute_subnetwork.subnet.name
  master_authorized_networks = var.master_authorized_networks
  ip_range_pods              = local.valid_pods_range_name
  ip_range_services          = local.valid_services_range_name

  maintenance_start_time = var.maintenance_start_time
  maintenance_end_time   = var.maintenance_end_time
  maintenance_recurrence = var.maintenance_recurrence
  maintenance_exclusions = var.maintenance_exclusions

  security_posture_mode               = var.security_posture_mode
  security_posture_vulnerability_mode = var.security_posture_vulnerability_mode

  default_max_pods_per_node = var.default_max_pods_per_node

  enable_private_nodes  = false
  enable_shielded_nodes = false
  # insecure_kubelet_readonly_port_enabled = false
  enable_l4_ilb_subsetting = true
  dns_cache                = true
  issue_client_certificate = true
  node_metadata            = "UNSPECIFIED"
  identity_namespace       = "null"
  datapath_provider        = "LEGACY_DATAPATH"

  # ~~~ NODE POOLS ~~~

  remove_default_node_pool = true

  node_pools              = var.node_pools
  node_pools_labels       = var.node_pools_labels
  node_pools_oauth_scopes = var.node_pools_oauth_scopes
  node_pools_taints       = var.node_pools_taints
  node_pools_tags         = local.node_pools_tags
  node_pools_cgroup_mode  = local.node_pools_cgroup_mode
}
```

### Terraform Version

```sh
Terraform v1.12.0
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v6.33.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.36.0
+ provider registry.terraform.io/hashicorp/random v3.7.2
```

### Terraform Provider Versions

```sh
Providers required by configuration:
.
├── provider[registry.terraform.io/hashicorp/google] ~> 6.33.0
└── module.gke
    ├── provider[registry.terraform.io/hashicorp/google] >= 6.14.0, < 7.0.0
    ├── provider[registry.terraform.io/hashicorp/kubernetes] ~> 2.10
    └── provider[registry.terraform.io/hashicorp/random] >= 2.1.0

Providers required by state:

    provider[registry.terraform.io/hashicorp/google]
```

### Additional information

When setting `local.cluster_endpoint_for_nodes` the value of `var.enable_private_nodes` or the length of `google_container_cluster.primary.private_cluster_config` is not taken into consideration.

A simple fix could look like:

```
locals {
  # before
  cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block

  # after
  cluster_endpoint_for_nodes = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block : null
}
```

yuval2313:
Fixes #2353 :)

### DIFF
--- a/autogen/main/main.tf.tmpl
+++ b/autogen/main/main.tf.tmpl
@@ -147,7 +147,7 @@ locals {
 {% if private_cluster %}
   cluster_endpoint           = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? (var.enable_private_endpoint || var.deploy_using_private_endpoint ? google_container_cluster.primary.private_cluster_config[0].private_endpoint : google_container_cluster.primary.private_cluster_config[0].public_endpoint) : google_container_cluster.primary.endpoint
   cluster_peering_name       = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].peering_name : null
-  cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
+  cluster_endpoint_for_nodes = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block : null
 {% else %}
   cluster_endpoint           = google_container_cluster.primary.endpoint
   cluster_endpoint_for_nodes = "${google_container_cluster.primary.endpoint}/32"
--- a/modules/beta-autopilot-private-cluster/main.tf
+++ b/modules/beta-autopilot-private-cluster/main.tf
@@ -78,7 +78,7 @@ locals {
 
   cluster_endpoint           = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? (var.enable_private_endpoint || var.deploy_using_private_endpoint ? google_container_cluster.primary.private_cluster_config[0].private_endpoint : google_container_cluster.primary.private_cluster_config[0].public_endpoint) : google_container_cluster.primary.endpoint
   cluster_peering_name       = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].peering_name : null
-  cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
+  cluster_endpoint_for_nodes = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block : null
 
   cluster_output_master_auth                        = concat(google_container_cluster.primary[*].master_auth, [])
   cluster_output_master_version                     = google_container_cluster.primary.master_version
--- a/modules/beta-private-cluster-update-variant/main.tf
+++ b/modules/beta-private-cluster-update-variant/main.tf
@@ -124,7 +124,7 @@ locals {
 
   cluster_endpoint           = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? (var.enable_private_endpoint || var.deploy_using_private_endpoint ? google_container_cluster.primary.private_cluster_config[0].private_endpoint : google_container_cluster.primary.private_cluster_config[0].public_endpoint) : google_container_cluster.primary.endpoint
   cluster_peering_name       = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].peering_name : null
-  cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
+  cluster_endpoint_for_nodes = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block : null
 
   cluster_output_master_auth                        = concat(google_container_cluster.primary[*].master_auth, [])
   cluster_output_master_version                     = google_container_cluster.primary.master_version
--- a/modules/beta-private-cluster/main.tf
+++ b/modules/beta-private-cluster/main.tf
@@ -124,7 +124,7 @@ locals {
 
   cluster_endpoint           = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? (var.enable_private_endpoint || var.deploy_using_private_endpoint ? google_container_cluster.primary.private_cluster_config[0].private_endpoint : google_container_cluster.primary.private_cluster_config[0].public_endpoint) : google_container_cluster.primary.endpoint
   cluster_peering_name       = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].peering_name : null
-  cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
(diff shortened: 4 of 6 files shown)


yuval2313:
In order to account for the creation of firewall which make use of the `local.cluster_endpoint_for_nodes` i have modified the logic for determining the value of this local variable according to this pseudo code:

```
IF var.enable_private_nodes == true:
  IF var.private_endpoint_subnetwork != null: RETURN data.google_compute_subnetwork.private_endpoint_subnetwork[0].ip_cidr_range
  ELSE IF var.master_ipv4_cidr_block != null: RETURN google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
  ELSE: RETURN local.cluster_subnet_cidr
ELSE: RETURN local.cluster_subnet_cidr
```

It accounts for the following scenarios:
1. Create a GKE cluster with no private nodes
2. Create a GKE cluster with private nodes and `master_ipv4_cidr_block` set
3. Create a GKE cluster with private nodes and `private_endpoint_subnetwork` set
4. Create a GKE cluster with private nodes and no options

This is based on the documentation which states the order of precedence in order to define the control plane network - https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#create_a_cluster_and_select_the_control_plane_ip_address_range.

Let me know please if this is overkill or if perhaps a better solution exists.

autogen/main/main.tf.tmpl
```diff
@@ -147,7 +147,13 @@ locals {
 {% if private_cluster %}
   cluster_endpoint           = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? (var.enable_private_endpoint || var.deploy_using_private_endpoint ? google_container_cluster.primary.private_cluster_config[0].private_endpoint : google_container_cluster.primary.private_cluster_config[0].public_endpoint) : google_container_cluster.primary.endpoint
   cluster_peering_name       = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? google_container_cluster.primary.private_cluster_config[0].peering_name : null
-  cluster_endpoint_for_nodes = google_container_cluster.primary.private_cluster_config[0].master_ipv4_cidr_block
+  cluster_endpoint_for_nodes = (var.enable_private_nodes && length(google_container_cluster.primary.private_cluster_config) > 0) ? (
```
apeabody, line 150: `cluster_endpoint_for_nodes ` is only used in `firewall.tf.tmpl` resources which are already gated `var.add_cluster_firewall_rules`, etc.  So it might be simpler to remove this local (which is *always* evaluated, hence the issue), and instead evaluate directly in those (gated) resources?

apeabody (commented):
Thanks @yuval2313!

An initial thought which might make this change simpler.

autogen/main/main.tf.tmpl
yuval2313: Thanks for the feedback :)

Are cluster firewall rules only relevant when enabling private nodes? Because the module allows their creation even without the flag which would also necessitate the same logic.

Before `35.0.0` I noticed that the `master_ipv4_cidr_block` had a default string value which I guess prevented this since private clusters always had this option specified.

Now it seems that we may need this since there are many possibilities.

yuval2313: Firewalls rules can still be created with `var.add_cluster_firewall_rules`, `var.add_master_webhook_firewall_rules`, and `var.add_shadow_firewall_rules`, regardless of the different scenarios which i described in my first comment:

> 1. Create a GKE cluster with no private nodes
> 2. Create a GKE cluster with private nodes and `master_ipv4_cidr_block` set
> 3. Create a GKE cluster with private nodes and `private_endpoint_subnetwork` set
> 4. Create a GKE cluster with private nodes and no options
> 
> This is based on the documentation which states the order of precedence in order to define the control plane network - https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#create_a_cluster_and_select_the_control_plane_ip_address_range.

Removing `local.cluster_endpoint_for_nodes` and evaluating directly inside the firewall resources wouldn't simplify the code as it still requires the same logic in order to determine the correct value for the master ip cidr range when firewall rules are created.

yuval2313:
@apeabody 
Thanks for the code review! Though I believe the current conditional logic is still necessary despite the `local.cluster_endpoint_for_nodes` being used only in `firewall.tf.tmpl` gated resources. Left a few replies to your review.

apeabody:
> @apeabody Thanks for the code review! Though I believe the current conditional logic is still necessary despite the `local.cluster_endpoint_for_nodes` being used only in `firewall.tf.tmpl` gated resources. Left a few replies to your review.

Thanks @yuval2313 - Right now the CI is failing on an un-related issue, once that is resolved I'll get this tested.

yuval2313:
@apeabody 
I reverted a small change I made to the `data.google_compute_subnetwork.gke_subnetwork` block, would love it if you could test it again :)

yuval2313:
@apeabody 
The tests seem to be passing :)

apeabody (approved):
Thanks for the contribution @yuval2313!

Issue #2348, opened 2025-05-12
IP masquerading not working in autopilot cluster


nbrys:
### TL;DR

As reported in: 

https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1895
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1578

The examples for configuring IP Masquerading (for example https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/a3ae61ff624aed87ae96a287a3619032a6bc0399/modules/beta-autopilot-public-cluster/masq.tf#L22) are not working for autopilot clusters.

When trying to configure the config map, the following error is thrown:

configmaps is forbidden: User "xxxx@xxxx.com" cannot create resource "configmaps" in API group "" in the namespace "kube-system": GKE Warden authz [denied by managed-namespaces-limitation]: the namespace "kube-system" is managed and the request's verb "create" is denied

### Expected behavior

The configmap to be created in the autopilot cluster

### Observed behavior

configmaps is forbidden: User "xxxx@xxxx.com" cannot create resource "configmaps" in API group "" in the namespace "kube-system": GKE Warden authz [denied by managed-namespaces-limitation]: the namespace "kube-system" is managed and the request's verb "create" is denied

### Terraform Configuration

```hcl
resource "kubernetes_config_map" "ip_masq_agent" {
  count = var.configure_ip_masq ? 1 : 0

  metadata {
    name      = "ip-masq-agent"
    namespace = "kube-system"

    labels = {
      maintained_by = "terraform"
    }
  }

  data = {
    config = <<EOF
nonMasqueradeCIDRs:
  - ${join("\n  - ", var.non_masquerade_cidrs)}
resyncInterval: ${var.ip_masq_resync_interval}
masqLinkLocal: ${var.ip_masq_link_local}
EOF
  }

  depends_on = [
    google_container_cluster.primary,
  ]
}
```

### Terraform Version

```sh
1.6.6
```

### Terraform Provider Versions

```sh
google = {
      source  = "hashicorp/google"
      version = ">= 6.27.0, < 7"
    }
    google-beta = {
      source  = "hashicorp/google-beta"
      version = ">= 6.27.0, < 7"
    }
    kubernetes = {
      source  = "hashicorp/kubernetes"
      version = "~> 2.10"
    }
```

Issue #1895, opened 2024-03-05
configure_ip_masq config and non_masquerade_cidrs are invalid for Autopilot Clusters and should be replaced with EgressNATPolicy


fernandoiury:
### TL;DR

Autopilot disables access to kube-system namespace. Thus the current approach to apply the ip-masq configmap does not work.

### Expected behavior

I'd expect to see the EgressNATPolicy being configured as per: https://cloud.google.com/kubernetes-engine/docs/concepts/ip-masquerade-agent#expandable-2

### Observed behavior

![image](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/assets/10088173/7bd1b418-5da1-416b-abfc-3e53fdeac513)

### Terraform Configuration

```hcl
configure_ip_masq = true
```

### Terraform Version

```sh
any
```

Issue #1578, opened 2023-03-02
ip-masq setting in AutoPilot clusters doesn't actually work.


grsiepka:
### TL;DR

Attempting to configure IP Masquerading in an AutoPilot cluster doesn't actually work. 

Terraform apply errors with: 
Error configmaps is forbidden: User "some-super-admin-serviceaccount" cannot create resource "configmaps" in API Group "" in the namespace "kube-system": GKEAutopilot authz: the namespace "kube-system" is managed and the request's verb "create" is denied.

### Expected behavior

I"m not actually sure because of how ip-masquerading is implemented in AutoPilot: From the Docs:

```
GKE enforces the IP masquerading behavior for Autopilot clusters through the following process:

GKE deploys the Egress NAT controller and the ip-masq-agent.
You create the Egress NAT policy.
The GKE controller translates the policy into the ip-masq-agent ConfigMap.
The ip-masq-agent DaemonSet reads the ConfigMap and then GKE enforces the IP masquerading behavior.

```

### Observed behavior

Failure as listed above, 

### Terraform Configuration

```hcl
pretty much same kindaas the example autopilot cluster code is with the following values set:

configure_ip_masq    = true
non_masquerade_cidrs = [data.google_compute_subnet.subnetwork.ip_cidr_range]  #This is our SharedVPC Node Subnet
#tested the below with/without and same results
disable_default_snat = true
```

### Terraform Version

```sh
Terraform 1.3.7
```

### Additional information

Our workaround at the moment is to remove the 10.0.0.0/8 CIDR from the default egressnatpolicy, as our GKE Node subnet is a 10.x address. (actually done via ACM). But was hoping this could be handled better in the module.  It's 100% possible i'm just doing this wrong too and if so would appreciate the guidance.

AliakseiKaniaveha:
We have exactly the same situation, the issue is perfectly relevant,  is there any way to re-open it?

aminealaouiecherif:
I've same issue !

hroost:
Same issue here for me as well. Also found [similar one](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/1895)

PR #2364, opened 2025-06-06
fix(autopilot): remove ip-masq configmap

apeabody:
- Autopilot does not use `ip-masq-agent` configMap for IP masquerade, removed.
- The Autopilot `EgressNATPolicy` method can't practically be implemented with the current Kubernetes provider method. (Can be opened as a Enhancement request)

Ref: https://cloud.google.com/kubernetes-engine/docs/concepts/ip-masquerade-agent

Closes: #2348 
Closes: #1895
Closes: #1578

BEGIN_COMMIT_OVERRIDE
fix(autopilot)!: remove ip-masq configmap (#2364)
END_COMMIT_OVERRIDE

### DIFF
--- a/modules/beta-autopilot-public-cluster/variables.tf
+++ b/modules/beta-autopilot-public-cluster/variables.tf
@@ -207,29 +207,6 @@ variable "network_tags" {
 }
 
 
-variable "non_masquerade_cidrs" {
-  type        = list(string)
-  description = "List of strings in CIDR notation that specify the IP address ranges that do not use IP masquerading."
-  default     = ["10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16"]
-}
-
-variable "ip_masq_resync_interval" {
-  type        = string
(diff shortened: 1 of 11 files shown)


modules/beta-autopilot-public-cluster/variables.tf
```diff
@@ -207,29 +207,6 @@ variable "network_tags" {
 }
 
 
-variable "non_masquerade_cidrs" {
```
bharathkkb, line 210: Should we consider marking as a breaking change due to var removal?

apeabody: Thanks @bharathkkb! - We can, but Autopilot already the configmap, so this configuration can't be successfully applied?

apeabody: This is going into a breaking release, so I'll go ahead and mark as a breaking release.

PR #2370, opened 2025-06-25
feat: add support for in_transit_encryption_config

martin-kulich-whalebone:
This property is set by default for clusters created in Google console. Unsetting it is not allowed, thus it prevents importing existing clusters.

This property was added in Google provider 6.36.0: https://github.com/hashicorp/terraform-provider-google/blob/main/CHANGELOG.md#6360-may-20-2025


apeabody (approved):
Thanks for the contribution @martin-kulich-whalebone!

PR #2374, opened 2025-06-28
feat: add support for max_run_duration

DrFaust92:
allow for node pools with nodes with max lifetime before they are recycled


apeabody (approved):

PR #2375, opened 2025-06-28
feat: add support for windows config

DrFaust92:
open to suggestion making this more robust, im not 100% happy with the implementation


apeabody (approved):

Yeah, ideally `windows_node_pools` would be a list of objects, which would enable validation, etc.  However this should work.

PR #2384, opened 2025-07-15
feat(safer-cluster): add support for gcs fuse csi driver

foosinn:
This feature is currently only available for other clusters, adjusted the template so safer-cluster variants can use it as well.


foosinn:
Awesome, thanks for taking care :)

PR #2387, opened 2025-07-17
feat: add enable_k8s_beta_apis support

maikelpoot:
This PR will expose the [enable_k8s_beta_apis:enabled_apis](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#enable_k8s_beta_apis-1) block of the `google_container_cluster` resource as a list of strings variable.

Example:
```
module "gke_cluster" {
  source     = "terraform-google-modules/kubernetes-engine/google//modules/beta-private-cluster"
  version    = "38.0.0"
  ...
  enable_k8s_beta_apis = [
    "networking.k8s.io/v1beta1/ipaddresses",
    "networking.k8s.io/v1beta1/servicecidrs"
  ]
  ...
 }
```

Tested as a local checkout against a cluster that had api's enabled with gcloud cli

### DIFF
--- a/test/integration/go.mod
+++ b/test/integration/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/GoogleCloudPlatform/cloud-foundation-toolkit/infra/blueprint-test v0.17.8
 	github.com/gruntwork-io/terratest v0.50.0
 	github.com/hashicorp/terraform-json v0.26.0
-	github.com/stretchr/testify v1.11.0
+	github.com/stretchr/testify v1.11.1
 	github.com/tidwall/gjson v1.18.0
 )
 
--- a/test/integration/go.sum
+++ b/test/integration/go.sum
@@ -237,8 +237,8 @@ github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/
 github.com/stretchr/testify v1.7.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.8.0/go.mod h1:yNjHg4UonilssWZ8iaSj1OCr/vHnekPRkoO+kdMU+MU=
 github.com/stretchr/testify v1.8.1/go.mod h1:w2LPCIKwWwSfY2zedu0+kehJoqGctiVI29o6fzry7u4=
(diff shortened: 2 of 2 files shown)


cluster.tf
```diff
@@ -33,6 +33,13 @@ resource "google_container_cluster" "primary" {
   network             = "projects/${local.network_project_id}/global/networks/${var.network}"
   deletion_protection = var.deletion_protection
 
+  dynamic "enable_k8s_beta_apis" {
```
apeabody, line 36: Please add to one of the existing examples (/examples) so this is covered by our tests.

maikelpoot: Damn, Just noticed i read your comment wrong, I added an additional test (by copying `simple_regional_private` to `simple_regional_private_with_beta_apis`) instead of adding it to an existing example.
Is that ok, or should i add it to the `simple_regional_private` example.

apeabody: Hi @maikelpoot! - It's fine to add as a new example, but more complex as that will need to be onboard as a test.  In particular:
- A new ./test/fixtures to apply
- A new ./test/integration to validate
- Add the required apply, validate, teardown for the next example to ./build/int.cloudbuild.yaml
- Etc

Either direction you choose is fine.

maikelpoot: hi @apeabody, 
As tempting as that sounds, i removed the extra example and applied the changes to the `simple_regional_private` example.

apeabody: Sounds good, you will need to pass a value for `enable_k8s_beta_apis` into the example fixture from here: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/main/test/fixtures/simple_regional_private/example.tf

maikelpoot: Added a beta api to the default value of the fixture variable. As an empty default value doesn't test anything.

But this does mean that if this api is promoted within later k8s/gke version it could fail on this value.

apeabody:
From the CI Test:

`Error: googleapi: Error 400: Beta API "networking.k8s.io/v1beta1/ipaddresses" may only be enabled if Beta API "networking.k8s.io/v1beta1/servicecidrs" is also enabled.`

maikelpoot:
> From the CI Test:
> 
> `Error: googleapi: Error 400: Beta API "networking.k8s.io/v1beta1/ipaddresses" may only be enabled if Beta API "networking.k8s.io/v1beta1/servicecidrs" is also enabled.`

Added the requested api to the test fixture

apeabody (approved):
Thanks for the contribution @maikelpoot!

Issue #2382, opened 2025-07-09
nodepool's node_pools_cgroup_mode is forced to be defined when setting any linux_node_config


prune998:
### TL;DR

I added values for `node_pools_linux_node_configs_sysctls` and the module was crashing with an error:

```
Call to function "coalesce" failed: no non-null, non-empty-string arguments.
```

at https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/v37.0.0/cluster.tf#L915

This is mainly due to the `coalesce` terraform function to expect at least one output not `null` or `""` out of the choices.

### Expected behavior

it should be possible to use any of the `linux_node_config` without being forced to create all (or some) of them.

### Observed behavior

```
Call to function "coalesce" failed: no non-null, non-empty-string arguments.
```

when `node_pools_cgroup_mode` is unset, the expression at https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/v37.0.0/cluster.tf#L915 will all evaluate to `null`

```
 cgroup_mode = coalesce(local.node_pools_cgroup_mode[each.value["name"]], local.node_pools_cgroup_mode["all"], null)
```

### Terraform Configuration

```hcl
"node_pools_linux_node_configs_sysctls": {
  "all": {},
  "default-node-pool": {},
  "gke-shared-spot-c3st44": {}
}
"node_pools": [
  {
    "add_fallback": "true",
    "auto_repair": "true",
    "auto_upgrade": "true",
    "autoscaling": "true",
    "disk_size_gb": "100",
    "disk_type": "pd-balanced",
    "enable_gcfs": "true",
    "enable_secure_boot": "true",
    "image_type": "COS_CONTAINERD",
    "initial_node_count": "0",
    "instance_class": "n2d-standard;c3-standard",
    "is_dedicated": "true",
    "local_ssd_ephemeral_storage_count": "0",
    "location_policy": "ANY",
    "machine_type": "n2d-standard-32",
    "max_count": "2000",
    "max_pods_per_node": "110",
    "max_surge": "5",
    "min_count": "0",
    "name": "gke-shared-spot-n2dst32",
    "node_locations": "",
    "number": "32",
    "orig_name": "gke-shared-spot",
    "sizes": "32,48;44",
    "skip_on_demand": "true",
    "spot": "true"
  }
]
```

### Terraform Version

```sh
Terraform v1.5.7
on darwin_arm64
[...]

PR #2396, opened 2025-08-03
fix: wrap cgroup_mode coalesce with try() to handle empty values

Dutta78:
When node_pools_linux_node_configs_sysctls is configured without node_pools_cgroup_mode, the coalesce() function fails because all arguments are empty strings or null.

Fixes: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2382


apeabody (approved):
Thanks for the contribution @Dutta78!

PR #2399, opened 2025-08-08
chore(CI): add conventional commitlint

apeabody:


.github/workflows/lint.yaml
```diff
@@ -47,8 +50,25 @@ jobs:
...
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
```
bharathkkb, line 66: Would configuring the npm cache help?

.github/workflows/lint.yaml
```diff
@@ -47,8 +50,25 @@ jobs:
...
+      - name: Install commitlint
+        run: |
+          npm install -D @commitlint/cli @commitlint/config-conventional
```
bharathkkb: nit: we should eventually pin the package version

bharathkkb (approved):

.github/workflows/lint.yaml
apeabody: I wanted to, but that appears to require a lock file be checked into the repo (or some other method).

.github/workflows/lint.yaml
apeabody: We can do that now.

PR #2404, opened 2025-08-10
feat:! add data cache count

DrFaust92:
this also adds missing keeper for local_ssd_ephemeral_storage_count


apeabody (approved):

PR #2414, opened 2025-08-20
fix(metadata): update ADC roles and services per module

daniel-cit:
This PR update the list of roles and services required by the ADC modules


apeabody:
We want to trigger a new release, so I've changed to a `fix`.

PR #2434, opened 2025-09-04
feat(TPG>=7.0)!: adding default_compute_class_enabled

rs1986x:
Hi

i have added the default_compute_class_enabled parameter. it is supported since version 7 of the provider so it's all good. the test fail in some examples because other dependency modules (e.g. project factory) have not yet updated to version 7 and for that reason it uses still provider 6.x


apeabody:
> i have added the default_compute_class_enabled parameter. it is supported since version 7 of the provider so it's all good. the test fail in some examples because other dependency modules (e.g. project factory) have not yet updated to version 7 and for that reason it uses still provider 6.x

Thanks @rs1986x - Can you also please bump the minimum TPG constraint as appropriate, that will make the version constraint failure clear rather than unknown `default_compute_class_enabled` field errors: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/main/autogen/main/versions.tf.tmpl#L27

Also, https://github.com/terraform-google-modules/terraform-google-project-factory/releases/tag/v18.1.0, now includes support for TPG v7. :)

rs1986x:
> > i have added the default_compute_class_enabled parameter. it is supported since version 7 of the provider so it's all good. the test fail in some examples because other dependency modules (e.g. project factory) have not yet updated to version 7 and for that reason it uses still provider 6.x
> 
> Thanks @rs1986x - Can you also please bump the minimum TPG constraint as appropriate, that will make the version constraint failure clear rather than unknown `default_compute_class_enabled` field errors: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/main/autogen/main/versions.tf.tmpl#L27
> 
> Also, https://github.com/terraform-google-modules/terraform-google-project-factory/releases/tag/v18.1.0, now includes support for TPG v7. :)

Hi @apeabody thanks for the comment, is there anything else you would require before merging the PR? thanks

apeabody:
Probably wait for v40 for TPG>=7

rs1986x:
> Probably wait for v40 for TPG>=7

No problem. For now i am using my fork, i imagine that eventually somebody would have made the same change i have made but if it can help speed up the release why not. Let me know if you require any other change, in the meantime thanks again and i look forward to see the default class parameter in the main modules

apeabody:
At a minimum this PR is needed first: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2446

apeabody:
Currently waiting on https://github.com/terraform-google-modules/terraform-google-bastion-host/pull/231

autogen/main/cluster.tf.tmpl
```diff
@@ -156,6 +156,7 @@ resource "google_container_cluster" "primary" {
 
   cluster_autoscaling {
     enabled = var.cluster_autoscaling.enabled
+    default_compute_class_enabled = var.default_compute_class_enabled
```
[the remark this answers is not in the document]
rs1986x: compute class should not require cluster autoscaling to work, you can have a default compute class that does use declared node pools as the base, no?

apeabody (approved):
Thanks for the contribution @rs1986x!

autogen/main/variables.tf.tmpl
```diff
@@ -858,6 +858,13 @@ variable "enable_shielded_nodes" {
 }
 
 {% endif %}
+
+variable "default_compute_class_enabled" {
+  type        = bool
+  description = "Enable Spot VMs as the default compute class for Node Auto-Provisioning"
+  default     = null
+}
+
```
RuiSMagalhaes, line 867: description is misleading, no? I believe this has nothing to do with SPOT vms?

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#default_compute_class_enabled-1

@rs1986x @apeabody

rs1986x: that's a good point - yeah... that is the way I use it, it is not the correct description.

PR #2447, opened 2025-09-16
feat: add support for NAP/autopilot resource manager tags

### LINKED ISSUE #2330: dynamic block "node_pool_auto_config" in cluster.tf missing "resource_manager_tags" field
Hi,

i build a standard GKE private cluster in a shared VPC in which the netsec team implements secure tags for firewall policies. 

I would like to tag my worker nodes with the key/value provided at the node pool level.

It works fine for all nodes being part of nodes pools defined through the "node_pools" variable, by setting the "node_pools_resource_manager_tags" parameter. 

But as soon as a node pool is provisioned through the auto-provisioning option by a user request for example, the secure tag is not set on the node. In the "private-cluster" submodule, in cluster.tf, i see only the "network_tags" field of the native terraform resource implemented, not the "resource_manager_tags" in the "node_pool_auto_config" dynamic block:

`  dynamic "node_pool_auto_config" {
    for_each = var.cluster_autoscaling.enabled && (length(var.network_tags) > 0 || var.add_cluster_firewall_rules) ? [1] : []
    content {
      network_tags {
        tags = var.add_cluster_firewall_rules ? (concat(var.network_tags, [local.cluster_network_tag])) : var.network_tags
      }
    }
  }`

Am i missing something? How could i set secure tags on auto provisioned node pools ?

pib:
Fixes #2330


pib:
@apeabody would you like me to add a new test or two for this change? Or I could add the new field to one or more existing tests if you'd like.

I got the test projects set up and ran the tests in a sandbox org. All the tests in build/int.cloudbuild.yaml passed (I put them in a shell script with `set -e` and ran them all that way).

apeabody:
> @apeabody would you like me to add a new test or two for this change? Or I could add the new field to one or more existing tests if you'd like.
> 
> I got the test projects set up and ran the tests in a sandbox org. All the tests in build/int.cloudbuild.yaml passed (I put them in a shell script with `set -e` and ran them all that way).

Thanks @pib! - The current test failure does not appear to be related to your change and can be re-run.  That said we should add test coverage to an existing example, likely `examples/node_pool`.

pib:
@apeabody I have added it to the node_pool test. Rather than putting the new field in the golden file, I had to check for it explicitly in the test because testutils.TGKEAssertGolden doesn't handle having values that need to be sanitized in map keys.

PR #2492, opened 2025-11-23
fix: Adding output types to metadata.yaml for gke-standard-cluster

Saranya3:
Added output type to metadata.yaml inorder to make connections to the gke-standard-cluster and deploy the same with connections.
Ref: https://g3doc.corp.google.com/cloud/config/gcpblueprints/g3doc/design_center.md#1-generate-output-type-in-metadata


modules/gke-standard-cluster/metadata.yaml
```diff
@@ -973,44 +973,151 @@ spec:
...
       - name: logging_service
         description: Logging service used
+        type: string
       - name: master_authorized_networks_config
         description: Networks from which access to master is permitted
+        type:
+          - list
+          - - object
+            - cidr_blocks:
+                - set
```
ayushmjain: would `list` make more sense here instead of `set`? Same for other places where `set` is used.

Saranya3: Acknowledged.

Daisyprakash:
LGTM

Saranya3:
@apeabody unable to start the test for the PR: [terraform-google-kubernetes-engine-int-trigger (cloud-foundation-cicd)](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2492/checks?check_run_id=57515371967)
Also, please let me know if something is pending in the PR that needs to be checked.

apeabody:
> @apeabody unable to start the test for the PR: [terraform-google-kubernetes-engine-int-trigger (cloud-foundation-cicd)](https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2492/checks?check_run_id=57515371967) Also, please let me know if something is pending in the PR that needs to be checked.

Hi @Saranya3 - The Int Test isn't required as this change is only to metadata which is covered by the Lint check.

apeabody (approved):
Thanks @Saranya3!

PR #2495, opened 2025-12-03
fix!: making location field as required for standard cluster

Daisyprakash:


apeabody (approved):
Thanks for the contribution @Daisyprakash

PR #2496, opened 2025-12-03
fix(gke-node-pool)!: module to allow List configurations for fields

opopov-google:
Update `gke-node-pool` module to allow List configurations for fields `additional_node_network_configs` and `additional_pod_network_configs`. This aligns with GKE TF Provider capabilities


modules/gke-node-pool/main.tf
```diff
@@ -276,15 +276,15 @@ resource "google_container_node_pool" "main" {
...
+        for_each = network_config.value.additional_node_network_configs != null ? network_config.value.additional_node_network_configs : []
         content {
           network    = additional_node_network_configs.value.network
           subnetwork = additional_node_network_configs.value.subnetwork
         }
       }
 
       dynamic "additional_pod_network_configs" {
-        for_each = network_config.value.additional_pod_network_configs != null ? [network_config.value.additional_pod_network_configs] : []
+        for_each = network_config.value.additional_pod_network_configs != null ? network_config.value.additional_pod_network_configs : []
```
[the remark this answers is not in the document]
opopov-google: Fixed

modules/gke-node-pool/main.tf
```diff
@@ -276,15 +276,15 @@ resource "google_container_node_pool" "main" {
       pod_range            = network_config.value.pod_range
 
       dynamic "additional_node_network_configs" {
-        for_each = network_config.value.additional_node_network_configs != null ? [network_config.value.additional_node_network_configs] : []
+        for_each = network_config.value.additional_node_network_configs != null ? network_config.value.additional_node_network_configs : []
```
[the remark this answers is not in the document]
opopov-google: fixed

apeabody:
updated subject for commitlint

apeabody:
Marked as a breaking change

modules/gke-node-pool/main.tf
```diff
@@ -276,15 +276,15 @@ resource "google_container_node_pool" "main" {
       pod_range            = network_config.value.pod_range
 
       dynamic "additional_node_network_configs" {
-        for_each = network_config.value.additional_node_network_configs != null ? [network_config.value.additional_node_network_configs] : []
+        for_each = coalesce(network_config.value.additional_node_network_configs, [])
```
apeabody, line 279: Thinking about this further, both `coalesce()` and `coalescelist()` return the first non-empty, so neither will work this further propose.  I think instead `concat()` can be used?

```suggestion
        for_each = concat(network_config.value.additional_node_network_configs, [])
```

apeabody: @gemini-code-assist - Can you please confirm?

apeabody: @gemini-code-assist - You are correct that `concat()` will fail as written if `network_config.value.additional_node_network_configs` is `null`.  However Terraform's `coalesce()` does not work directly with lists, they must be expanded with `...` and it will not return an empty list.

PR #2501, opened 2025-12-10
feat: allow new hashicorp/kubernetes version 3.0.0

regressEdo:
New version 3.0.0 got released: https://github.com/hashicorp/terraform-provider-kubernetes/releases/tag/v3.0.0 but this current module version constraint doesn't allow the upgrade

```
  Error: Failed to query available provider packages
  
  Could not retrieve the list of available versions for provider
  hashicorp/kubernetes: locked provider
  registry.terraform.io/hashicorp/kubernetes 3.0.0 does not match configured
  version constraint ~> 2.10, ~> 2.13, 3.0.0; must use terraform init -upgrade
  to allow selection of new versions
  
  To see which modules are currently depending on hashicorp/kubernetes and what
  versions are specified, run the following command:
      terraform providers
  
  exit status 1
```

PS: I had some problems with my previous MR https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2499, I ended up opening a new one

PR #2521, opened 2026-01-13
fix: move deprecated to versioned resources

HarisDotParis:
- Terraform's Kubernetes provider's v3 update has deprecated several unversioned resource names
- This PR updates all affected resources
  - kubernetes_service
  - kubernetes_service_account
  - kubernetes_config_map
  - kubernetes_pod
- If a resource is part of regular code, then there is always a corresponding moved block to move the old resource to the new one
- If a resource is part of directory `examples/`, then its resource type is only renamed, but there is no corresponding moved block as that is only exemplary code
- Additionally, the file `examples/workload_identity/main.tf` has been updated to be valid and syntactically correct

### DIFF
--- a/examples/workload_identity/main.tf
+++ b/examples/workload_identity/main.tf
@@ -60,10 +60,11 @@ module "workload_identity" {
   source  = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
   version = "~> 43.0"
 
-  project_id          = var.project_id
-  name                = "iden-${module.gke.name}"
-  namespace           = "default"
-  use_existing_k8s_sa = false
+  project_id    = var.project_id
+  name          = "iden-${module.gke.name}"
+  network       = var.network
+  subnetwork    = var.subnetwork
+  ip_range_pods = var.ip_range_pods
 }
 
 # example with existing KSA
@@ -80,13 +81,11 @@ module "workload_identity_existing_ksa" {
   source  = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
   version = "~> 43.0"
 
-  project_id          = var.project_id
-  name                = "existing-${module.gke.name}"
-  cluster_name        = module.gke.name
-  location            = module.gke.location
-  namespace           = "default"
-  use_existing_k8s_sa = true
-  k8s_sa_name         = kubernetes_service_account_v1.test.metadata[0].name
+  project_id    = var.project_id
+  name          = "existing-${module.gke.name}"
+  network       = var.network
+  subnetwork    = var.subnetwork
+  ip_range_pods = var.ip_range_pods
 }
 
 # example with existing GSA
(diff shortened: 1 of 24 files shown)


apeabody (changes requested):
Thanks for the contribution @HarisDotParis!

Please run `make docker_generate_docs` and add to your change.

apeabody (changes requested):
Thanks for the contribution @HarisDotParis 

From the lint:

```
terraform_validate ./test/fixtures/workload_identity
╷
│ Error: Unsupported argument
│ 
│   on ../../../examples/workload_identity/main.tf line 65, in module "workload_identity":
│   65:   network       = var.network
│ 
│ An argument named "network" is not expected here.
╵
╷
│ Error: Unsupported argument
│ 
│   on ../../../examples/workload_identity/main.tf line 66, in module "workload_identity":
│   66:   subnetwork    = var.subnetwork
│ 
│ An argument named "subnetwork" is not expected here.
╵
╷
│ Error: Unsupported argument
│ 
│   on ../../../examples/workload_identity/main.tf line 67, in module "workload_identity":
│   67:   ip_range_pods = var.ip_range_pods
│ 
│ An argument named "ip_range_pods" is not expected here.
```

examples/workload_identity/main.tf
```diff
@@ -80,13 +81,11 @@ module "workload_identity_existing_ksa" {
...
-  cluster_name        = module.gke.name
-  location            = module.gke.location
-  namespace           = "default"
-  use_existing_k8s_sa = true
-  k8s_sa_name         = kubernetes_service_account.test.metadata[0].name
+  project_id    = var.project_id
+  name          = "existing-${module.gke.name}"
+  network       = var.network
+  subnetwork    = var.subnetwork
+  ip_range_pods = var.ip_range_pods
```
patjlm, line 88: I think those are irrelevant here, but the deleted ones should remain.

examples/workload_identity/main.tf
```diff
@@ -99,9 +98,11 @@ module "workload_identity_existing_gsa" {
...
   version = "~> 43.0"
 
-  project_id          = var.project_id
-  name                = google_service_account.custom.account_id
-  use_existing_gcp_sa = true
+  project_id    = var.project_id
+  name          = google_service_account.custom.account_id
+  network       = var.network
+  subnetwork    = var.subnetwork
+  ip_range_pods = var.ip_range_pods
```
patjlm, line 105: Same here, irrelevant fields, but `use_existing_gcp_sa` should be kept

patjlm:
x-ref kubernetes provider bump: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2501

patjlm:
this PR will handle service accounts: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2560

PR #2529, opened 2026-01-28
update private cluster module to support rotation interval

ggprod:


ggprod:
apologies, I meant to open a PR against my forked repo

PR #2547, opened 2026-02-23
fix: auto_provisioning_defaults not created when enable_default_compute_class=true and NAP disabled

elayaraja-kv:
## Problem

When using a Standard GKE cluster with:
- `cluster_autoscaling.enabled = false` (NAP disabled)
- `cluster_autoscaling.enable_default_compute_class = true` (autopilot ComputeClass enabled)

The `auto_provisioning_defaults` block (which sets `service_account`, `oauth_scopes`, `disk_size`, etc. for NAP-provisioned nodes) is never created because the `for_each` guard only checks `cluster_autoscaling.enabled`:

```hcl
dynamic "auto_provisioning_defaults" {
  for_each = var.cluster_autoscaling.enabled ? [1] : []  # ← bug: ignores enable_default_compute_class
```

`defaultComputeClassConfig` works **independently** of `enableNodeAutoprovisioning` — it provisions autopilot ComputeClass nodes on demand even when NAP is disabled. Without `auto_provisioning_defaults`, these nodes are created without the configured service account, falling back to the default Compute Engine SA.

## Fix

Also evaluate `enable_default_compute_class` in the `for_each` condition:

```hcl
dynamic "auto_provisioning_defaults" {
  for_each = var.cluster_autoscaling.enabled || lookup(var.cluster_autoscaling, "enable_default_compute_class", false) ? [1] : []
```

## Affected modules

All non-autopilot standard cluster modules share the same bug:
- `cluster.tf` (root / public cluster)
- `modules/private-cluster`
- `modules/private-cluster-update-variant`
- `modules/beta-private-cluster`
- `modules/beta-private-cluster-update-variant`
- `modules/beta-public-cluster`
- `modules/beta-public-cluster-update-variant`

Autopilot cluster modules are not affected.

## References

- [Autopilot on Standard clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/about-autopilot-mode-standard-clusters)
- `defaultComputeClassConfig` is independent of `enableNodeAutoprovisioning` per GKE API docs

### DIFF
--- a/autogen/safer-cluster/main.tf.tmpl
+++ b/autogen/safer-cluster/main.tf.tmpl
@@ -140,6 +140,7 @@ module "gke" {
   // We enable private endpoints to limit exposure.
   enable_private_endpoint       = var.enable_private_endpoint
   deploy_using_private_endpoint = true
+  private_endpoint_subnetwork   = var.private_endpoint_subnetwork
 
   // Private nodes better control public exposure, and reduce
   // the ability of nodes to reach to the Internet without
@@ -148,7 +149,7 @@ module "gke" {
 
   master_global_access_enabled = true
 
-  master_ipv4_cidr_block = var.master_ipv4_cidr_block
+  master_ipv4_cidr_block = var.private_endpoint_subnetwork != null ? null : var.master_ipv4_cidr_block
 
   cloudrun = var.cloudrun
 
--- a/autogen/safer-cluster/variables.tf.tmpl
+++ b/autogen/safer-cluster/variables.tf.tmpl
@@ -323,6 +323,12 @@ variable "master_ipv4_cidr_block" {
   default     = "10.0.0.0/28"
 }
 
+variable "private_endpoint_subnetwork" {
+  type        = string
+  description = "The subnetwork to use for the hosted master network. If specified, `master_ipv4_cidr_block` is ignored."
+  default     = null
+}
+
 variable "dns_cache" {
   type        = bool
   description = "(Beta) The status of the NodeLocal DNSCache addon."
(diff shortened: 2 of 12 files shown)


elayaraja-kv:
Both issues raised in the Gemini review have been addressed in the second commit:

1. **Source template updated**: `autogen/main/cluster.tf.tmpl` now includes all three fixes so future code generation will not regress them.

2. **`node_pool_auto_config` fixed**: Applied the same `enable_default_compute_class` guard to both the outer `node_pool_auto_config` and inner `network_tags` `for_each` conditions, ensuring network tags and security settings are applied when ComputeClass provisions nodes with NAP disabled.

elayaraja-kv:
Addressed in the third commit:

- Replaced all three `lookup(var.cluster_autoscaling, "enable_default_compute_class", false)` expressions with direct attribute access `var.cluster_autoscaling.enable_default_compute_class` (it is `optional(bool, false)` in the variable definition)
- Extracted the repeated boolean logic into `local.autoscaling_or_compute_class_enabled` in `main.tf` / `main.tf.tmpl`
- All three `for_each` conditions (`auto_provisioning_defaults`, `node_pool_auto_config` outer and inner) now reference the local
- Applied across source template and all 7 generated modules

elayaraja-kv:
Reverted to `lookup()` in the fourth commit for backward compatibility, as suggested. The `local.autoscaling_or_compute_class_enabled` variable now uses `lookup(var.cluster_autoscaling, "enable_default_compute_class", false)` — applied to both the source template and all 7 generated `main.tf` files.

apeabody (approved):
Thanks for the contribution @elayaraja-kv!

PR #2442, opened 2025-09-12
feat!: support enabling default compute class in cluster autoscaler

romancin:
This PR enables [Default Compute Class](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#default_compute_class_enabled-1) support for Cluster Autoscaler.

PR #2548, opened 2026-02-24
fix: missing default_compute_class_enabled from safer-cluster

ben-pulley:
It looks like #2442 didn't add `default_compute_class_enabled` to the variable `cluster_autoscaling` declarations for the `safer-cluster` and safer-cluster-update-variant` modules, so Terraform silently drops those variables:

```
Warning: Object attribute is ignored
...
enable_default_compute_class = true

The object type for input variable "cluster_autoscaling" does not include an attribute named "enable_default_compute_class", so this definition is unused.
```

This PR adds the missing attribute to the input variable so that the value can be used by consumers.

### DIFF
--- a/examples/confidential_autopilot_private/network.tf
+++ b/examples/confidential_autopilot_private/network.tf
@@ -16,7 +16,7 @@
 
 module "gcp-network" {
   source  = "terraform-google-modules/network/google"
-  version = "~> 13.0"
+  version = "~> 16.0"
 
   project_id   = var.project_id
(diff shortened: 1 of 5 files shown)


ben-pulley:
> Thanks for the contribution @ben-pulley!
> 
> Please run the 'make build' command and commit changes.

@apeabody  Just ran and committed the changes!

ben-pulley:
@apeabody is there anything else needed on this PR to merge?

PR #2550, opened 2026-03-05
feat: add allow_net_admin support to standard cluster modules

elayaraja-kv:
## Description

`allow_net_admin` enables the NET_ADMIN capability on the cluster master and is fully supported by the `google_container_cluster` resource for all cluster types (see [provider docs](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#allow_net_admin-1)). However, the autogen template gated it behind `{% if autopilot_cluster %}`, making it unavailable for standard (non-autopilot) cluster modules.

This PR brings parity with the autopilot modules by exposing `allow_net_admin` across all standard cluster modules.

## Changes

- **`autogen/main/cluster.tf.tmpl`** — removed `{% if autopilot_cluster %}` gate so `allow_net_admin` is included for all cluster types
- **`autogen/main/variables.tf.tmpl`** — same, moved variable declaration outside the autopilot condition, and aligned formatting (`type` / `default` spacing)
- **All generated standard cluster modules** — added `allow_net_admin = var.allow_net_admin` to `cluster.tf` and the variable declaration to `variables.tf`:
  - Root module
  - `modules/private-cluster`
  - `modules/private-cluster-update-variant`
  - `modules/beta-private-cluster`
  - `modules/beta-private-cluster-update-variant`
  - `modules/beta-public-cluster`
  - `modules/beta-public-cluster-update-variant`
- **`modules/gke-standard-cluster`** — changed hardcoded `allow_net_admin = false` to `var.allow_net_admin` and added the variable (default `false` to preserve existing behaviour)

## Backward Compatibility

- All new variables default to `null` (or `false` for `gke-standard-cluster` to match the previous hardcoded value), so existing configurations are unaffected.

## Related

- Autopilot modules already support this: `modules/beta-autopilot-private-cluster`, `modules/beta-autopilot-public-cluster`, `modules/gke-autopilot-cluster`


apeabody:
Hi @elayaraja-kv 

Per the provider documentation, it appears this field should only be enabled for Autopilot clusters.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#allow_net_admin-1
"allow_net_admin - (Optional) Enable NET_ADMIN for the cluster. Defaults to false. This field should only be enabled for Autopilot clusters (enable_autopilot set to true)."

elayaraja-kv:
Thanks for the review, @apeabody!

You are right that the provider docs say "should only be enabled for Autopilot clusters" -- that is the primary and most impactful use case, and I have updated the variable description across all affected files to make that clear.

To clarify the distinction:

- **Autopilot clusters**: `allow_net_admin = true` is the *only* way to grant `NET_ADMIN` to workloads, since Autopilot's security model strips privileged capabilities by default. This is required for service meshes like Linkerd or Istio when running without CNI mode.
- **Standard clusters**: Workloads can grant `NET_ADMIN` at the pod level via `securityContext.capabilities.add`, so the cluster-level flag is not strictly necessary. However, the underlying GKE API and Terraform provider both accept it for standard clusters -- the autogen template was simply being more restrictive than the provider itself.

The intent of this PR is not to encourage use on standard clusters, but to remove an arbitrary autogen gate so that:
1. The module is not more restrictive than the Terraform provider
2. Users migrating between cluster types do not hit unexpected missing variables
3. The variable defaults to `null` (no change to existing behaviour)

The updated description now reads:

> Primarily intended for Autopilot clusters, where Autopilot's security model prevents pods from requesting privileged capabilities and this flag is the only way to grant NET_ADMIN. On Standard clusters, workloads can grant NET_ADMIN at the pod level via securityContext (e.g. for service meshes such as Linkerd or Istio running without CNI mode).

Happy to make further adjustments if you prefer a different approach.

PR #2556, opened 2026-03-17
fix: add enable_k8s_tokens_via_dns to prevent state drift

ThYpHo0n:
GCP sets enable_k8s_tokens_via_dns to true server-side, but the module didn't manage it — causing Terraform to detect true -> null drift on every plan. Expose the attribute so it can be set explicitly.


ThYpHo0n:
Applied the Gemini code review feedback, this is true we tested the changes and found out that the value of enable_k8s_tokens_via_dns was always reset still causing a drift. Instead of making that variable available in the module the coalesce is the cleaner apprach.

Linting should be fixed as well, had some trouble running them locally due to AMR/amd64 emulation issues.

PR #2573, opened 2026-04-17
feat: Add secret_sync_config support to all beta cluster modules

### LINKED ISSUE #2511: Expose the new secret_sync_config setting to enable allow synchronize secrets to Kubernetes Secrets in all modules
### TL;DR

The `google_container_cluster` resource has the option to enable the new (beta) [secret_sync_config](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#secret_sync_config-1) so that we can [Synchronize secrets to Kubernetes Secrets](https://docs.cloud.google.com/secret-manager/docs/sync-k8-secrets).

Currently the UI does not expose it as well, so we had to enable it manually using:

```sh
gcloud --project <project name> \
  beta container clusters update <cluster name> \
  --location=<cluster location> \
  --enable-secret-sync
```

This took about 20-30 minutes to enable, but now we are able to create the `SecretSync` resources.

```
apiVersion: secret-sync.gke.io/v1
kind: SecretSync
metadata:
  name: some-secret
spec:
  serviceAccountName: some-sa
  secretProviderClassName: some-secret
  secretObject:
    type: Opaque
    data:
      - sourcePath: "foo"
        targetKey: "BAR"
```

It seems this feature will be "enabled" somewhere around Q1 26 ([source](https://issuetracker.google.com/issues/187158567#comment6))

### Terraform Resources

```markdown
https://registry.terraform.io/providers/hashicorp/google/7.14.1/docs/resources/container_cluster#secret_sync_config-2
```

### Detailed design

```markdown
Allow us consumers to enable this feature in IaC before GA
```

### Additional information

_No response_

litzius:
Closes #2511

## Summary

- Adds `secret_sync_config` support (enables the [Secret Sync add-on](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#secret_sync_config-1)) across all cluster modules where the block is supported by the provider
- Uses `enable_secret_sync` (bool, default `false`) for autogen-based modules, and `secret_sync_config` object variable for handwritten `gke-*` modules — matching the existing `secret_manager_config` patterns
- Autogen templates (`autogen/main/`, `autogen/safer-cluster/`) updated with `{% if beta_cluster %}` guard so `make generate` stays consistent

## Affected modules

| Module | Variable |
|--------|----------|
| `beta-private-cluster` | `enable_secret_sync` |
| `beta-private-cluster-update-variant` | `enable_secret_sync` |
| `beta-public-cluster` | `enable_secret_sync` |
| `beta-public-cluster-update-variant` | `enable_secret_sync` |
| `beta-autopilot-private-cluster` | `enable_secret_sync` |
| `beta-autopilot-public-cluster` | `enable_secret_sync` |
| `gke-standard-cluster` | `secret_sync_config` |
| `gke-autopilot-cluster` | `secret_sync_config` |
| `safer-cluster` | `enable_secret_sync` (pass-through) |
| `safer-cluster-update-variant` | `enable_secret_sync` (pass-through) |

## Why GA modules are excluded

The GA `google` provider (v7.28.0, the version in the lock file) does not yet expose the `secret_sync_config` block — only `google-beta` does. The root module, `private-cluster`, and `private-cluster-update-variant` use `provider = google` and are therefore excluded until the block lands in a GA release.

## Test plan

- [x] `terraform validate` passes on all 12 modules (10 beta/gke + root + private-cluster)
- [x] `make build` run — modules regenerated from autogen templates and all README Inputs/Outputs tables updated


autogen/main/cluster.tf.tmpl
```diff
@@ -311,6 +311,15 @@ resource "google_container_cluster" "primary" {
...
   }
 
+  {% if beta_cluster %}
+  dynamic "secret_sync_config" {
+    for_each = var.enable_secret_sync ? [true] : []
+    content {
+      enabled = true
+    }
+  }
+  {% endif %}
```
apeabody: nit: Please move the new line return after `{% endif %}` to just above, the goal is to eliminate the diff in the non-beta cluster files.

autogen/main/variables.tf.tmpl
```diff
@@ -713,6 +713,14 @@ variable "enable_secret_manager_addon" {
   type        = bool
   default     = false
 }
+
```
apeabody: nit: Please move the new line return before `{% if beta_cluster %}` to just after, the goal is to eliminate the diff in the non-beta cluster files.

autogen/main/variables.tf.tmpl
```diff
@@ -713,6 +713,14 @@ variable "enable_secret_manager_addon" {
...
+{% if beta_cluster %}
+variable "enable_secret_sync" {
+  description = "Enable the Secret Sync add-on for this cluster."
+  type        = bool
+  default     = false
+}
+{% endif %}
```
apeabody: nit: This `{% endif %}` can be removed along with the immediately following `{% if beta_cluster %}`, the goal is to consolidate the duplicative `if` blocks.

apeabody (changes requested):
Thanks for the contribution @litzius!  A few minor formatting requests below.

Please note the CI is currently blocked on a recent change to the GKE API, once that is resolved we can re-run the Integration tests: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2575  

Confirmed `secret_sync_config` is present in [v7.17.0](https://registry.terraform.io/providers/hashicorp/google/7.17.0/docs/resources/container_cluster) beta TGP.

autogen/main/cluster.tf.tmpl
litzius: Done, moved the blank line inside the block.

autogen/main/variables.tf.tmpl
litzius: Done, moved the blank line to just after the opening tag.

autogen/main/variables.tf.tmpl
litzius: Done, consolidated both variable declarations into a single {% if beta_cluster %} block.

apeabody (approved):
Thanks @litzius - LGTM!

tcolgate:
is there a specific reason this has been left out of the top level module?

litzius:
Hi @tcolgate! At the time PR #2573 was merged, `secret_sync_config` was only supported in the `google-beta` provider (since v7.17.0) and had not yet landed in the GA `google` provider. Because the top-level module and standard private cluster modules use the GA `google` provider, adding it there would have caused schema validation failures.

Once the feature became GA in the Google provider, we prepared the follow-up to bring `secret_sync_config` (and nested `rotation_config`) to the top-level module and all GA modules — this is now open in #2645!

PR #1946, opened 2024-05-15
feat: Add secondary_boot_disks to node_pool configuration.

yorik:
PR adds secondary_boot_disks setting to cluster.tf in the node_config part.

PR #2580, opened 2026-04-28
feat: allow configurable mode for secondary_boot_disk

bbensky:
## Summary

Parameterizes the hardcoded `mode = "CONTAINER_IMAGE_CACHE"` in the `secondary_boot_disks` dynamic block, exposing a new optional `secondary_boot_disk_mode` field on `node_pools` entries.

## Why

GKE [secondary boot disks](https://cloud.google.com/kubernetes-engine/docs/how-to/data-container-image-preloading) support two distinct features that share the same API:

| `mode` value | Use case |
|--------------|----------|
| `CONTAINER_IMAGE_CACHE` | Preloads container images for faster pod start. Requires Image Streaming (`gcfs_config { enabled = true }`). |
| (mode omitted) | Preloads arbitrary data onto a node-local disk for pods to mount via hostPath. |

The module currently only supports the first variant — `mode` is hardcoded since #1946. There is no way for a consumer to use the data-preload variant via this module today.

This change makes `mode` configurable per-pool, while preserving the existing default for backward compatibility.

## Backward compatibility

Existing consumers using `secondary_boot_disk` continue to receive `CONTAINER_IMAGE_CACHE` automatically — the `lookup()` default matches the previous hardcoded value. No tfvars changes required for existing users.

## Usage

For container image preloading (existing behavior, no change):

```hcl
node_pools = [{
  name                = "my-pool"
  secondary_boot_disk = "projects/my-project/global/images/my-cached-image"
  # secondary_boot_disk_mode defaults to "CONTAINER_IMAGE_CACHE"
}]
```

For data preloading (new):

```hcl
node_pools = [{
  name                     = "my-pool"
  secondary_boot_disk      = "projects/my-project/global/images/my-data-image"
  secondary_boot_disk_mode = null   # omits the mode field — data preload
}]
```

## Files changed

- Source of truth: `autogen/main/cluster.tf.tmpl` and `autogen/main/README.md`
- Regenerated via `make build` and `make docker_generate_docs` (per `CONTRIBUTING.md`)

## Tests

This is a small behavioral change with a backward-compatible default. I haven't added an integration test (would require a GCP project I can't easily provision); happy to add one if the maintainers prefer — pointers to the closest existing test welcome.


apeabody (approved):
Thanks for the contribution @bbensky!

PR #2600, opened 2026-06-09
fix!: Change resource type to kubernetes_config_map_v1

jullianow:
Fix deprecation warning when planning against provider v7.15.0


jullianow:
@apeabody Can you review this PR?

apeabody (changes requested):
Thanks for the contribution @jullianow 

Unfortunately it looks like this will be a breaking change.  Can you please add a quick section on upgrading to a new file/section `docs/upgrading_to_v45.0.md`

Probably something like `terraform state mv kubernetes_config_map.ip_masq_agent kubernetes_config_map_v1.ip_masq_agent` (please verify the full/module path, etc)

jullianow:
@apeabody ping

tcarrondo:
This is the right shape for the fix, and worth noting it is the only one of the four open PRs that is: the `moved` block approach taken in #2619 and #2627 cannot work, because the kubernetes provider does not implement `MoveResourceState`.

Tested on Terraform 1.10.5 with `hashicorp/kubernetes` 3.2.1:

```hcl
resource "kubernetes_config_map_v1" "ip_masq_agent" {
  count = var.configure_ip_masq ? 1 : 0
  metadata {
    name      = "ip-masq-agent"
    namespace = "kube-system"
  }
  data = { config = "x" }
}

moved {
  from = kubernetes_config_map.ip_masq_agent
  to   = kubernetes_config_map_v1.ip_masq_agent
}
```

| State | Result |
| --- | --- |
| nothing at the source address (`configure_ip_masq = false`) | no-op, plan succeeds |
| `kubernetes_config_map.ip_masq_agent` present in state (`configure_ip_masq = true`) | `Error: Move Resource State Not Supported` |

So a `moved` block breaks plans for precisely the users it is meant to help, with no way around it other than removing the block. `terraform state mv` is no help either — it rejects the move outright with `Invalid state move request: resource types don't match`, since `_v1` is a distinct type in the provider schema.

That leaves what this PR does — rename plus an upgrade guide — as the only workable option. For the guide, users have two paths, neither of which Terraform can automate:

1. Let the config map be recreated. It is a config map in `kube-system` with no persistent data, and only users who set `configure_ip_masq = true` are affected at all.
2. Rewrite the type in state, if recreation is not acceptable:

```bash
terraform state pull > state.json
jq '.resources |= map(if .type == "kubernetes_config_map" then .type = "kubernetes_config_map_v1" else . end)
    | .serial += 1' state.json > state.new.json
terraform state push state.new.json
```

Also confirming the warning is still reproducible on v44.3.0 — `masq.tf` on `main` still declares the unsuffixed type. A workaround until this lands, for anyone whose only kubernetes provider requirement comes from this module: pin `hashicorp/kubernetes` to `2.38.0`, which does not carry the deprecation marking.

The CI here shows `ACTION_REQUIRED` rather than a failure, so this appears to be waiting on a maintainer to authorise the run.

PR #2619, opened 2026-07-11
fix: bump ip-masq-agent configmap to kubernetes_config_map_v1

kiarashazarnia:
## Summary
- Migrate the `ip-masq-agent` ConfigMap from deprecated `kubernetes_config_map` to `kubernetes_config_map_v1`
- Add `moved` blocks so existing state migrates automatically on upgrade
- Update autogen templates and regenerate affected module files

## Test plan
- [x] `make build` (or `make generate`) to confirm autogen output is in sync
- [x] `terraform validate` on a module/example with `configure_ip_masq = true`
- [x] Confirm `terraform plan` shows no deprecation warnings and no resource recreation for existing deployments

### Sample Deprecation Warning

```sh
$ terraform plan

<SOME-OUTPUT>

Plan: 0 to add, 1 to change, 0 to destroy.

Changes to Outputs:
  ~ gke_module                         = (sensitive value)
╷
│ Warning: Deprecated Resource
│
│   with module.gke.kubernetes_config_map.ip_masq_agent,
│   on .terraform/modules/gke/modules/private-cluster/masq.tf line 22, in resource "kubernetes_config_map" "ip_masq_agent":
│   22: resource "kubernetes_config_map" "ip_masq_agent" {
│
│ Deprecated; use kubernetes_config_map_v1.
╵
```

Since this is my first PR on this repo, I’ve addressed a trivial issue to get familiar with the process. Thanks! 🙌


tcarrondo:
Heads up that the `moved.tf` part of this PR cannot work: the kubernetes provider does not implement `MoveResourceState`, so Terraform refuses cross-resource-type moves for it.

Tested on Terraform 1.10.5 with `hashicorp/kubernetes` 3.2.1, using this module's exact resource shape:

```hcl
resource "kubernetes_config_map_v1" "ip_masq_agent" {
  count = var.configure_ip_masq ? 1 : 0
  metadata {
    name      = "ip-masq-agent"
    namespace = "kube-system"
  }
  data = { config = "x" }
}

moved {
  from = kubernetes_config_map.ip_masq_agent
  to   = kubernetes_config_map_v1.ip_masq_agent
}
```

| State | Result |
| --- | --- |
| nothing at the source address (`configure_ip_masq = false`) | no-op, plan succeeds |
| `kubernetes_config_map.ip_masq_agent` present in state (`configure_ip_masq = true`) | `Error: Move Resource State Not Supported` |

The block is silently fine for the majority who leave `configure_ip_masq` at its default, which is probably why it looks like it works — but it hard-fails the plan for anyone who actually has the config map in state, i.e. exactly the users the move is for. There is no way past the error other than deleting the block.

`terraform state mv` does not help either: `Invalid state move request … resource types don't match`, because `_v1` is a distinct type in the provider schema even though the implementation is shared.

Suggest dropping `moved.tf` and documenting the migration instead, as #2600 does. Users have two options: let the config map be recreated (it holds no persistent data), or rewrite the type in state directly:

```bash
terraform state pull > state.json
jq '.resources |= map(if .type == "kubernetes_config_map" then .type = "kubernetes_config_map_v1" else . end)
    | .serial += 1' state.json > state.new.json
terraform state push state.new.json
```

Issue #2476, opened 2025-10-21
Add `single_process_oom_kill` to `kubelet_config`


upodroid:
### TL;DR

Some keys got added to kubelet_config recently

### Terraform Resources

```markdown
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#nested_kubelet_config
```

### Detailed design

```markdown

```

PR #2631, opened 2026-08-18
feat: add single_process_oom_kill to node pool kubelet_config

robert-ziegon:
## Description

Adds passthrough of a new node-pool key `single_process_oom_kill` to `kubelet_config` (all variants, linux and windows pool blocks, plus the head-node-pool block), as requested in #2476.

The underlying provider field exists in `google` / `google-beta` (GA since 6.x). On cgroupv2 nodes the kubelet defaults to `memory.oom.group=1`, so the kernel OOM-kills the *entire container* when any process exceeds memory. For multi-process containers (e.g. php-fpm pools, nginx) this turns a single over-memory request into the loss of every in-flight request in the container; setting `single_process_oom_kill = true` restores per-process kills (cgroup v1 semantics). Real-world motivation: a production incident where one oversized request pattern serially OOM-killed every pod of a php-fpm deployment, because each kill took out all 30 fpm workers at once.

## Notes

- The key defaults to `null`, which omits the provider field and preserves the kubelet default — no behavior change for existing users.
- Changes are generated: `autogen/main/cluster.tf.tmpl` + `autogen/main/README.md` edited, then `make generate` and `make generate_docs`.
- Known interaction (pre-existing, out of scope here): setting *only* `single_process_oom_kill` on a pool materializes the `kubelet_config` block, whose `cpu_manager_policy` lookup defaults to `"static"` — so users opting into this key also get `cpu_manager_policy = "static"` applied unless they set it explicitly. That default predates this PR; happy to open a separate discussion about defaulting it to `null`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,6 +367,7 @@ The node_pools variable takes the following parameters:
 | image_type | The image type to use for this node. Note that changing the image type will delete and recreate all nodes in the node pool | COS_CONTAINERD | Optional |
 | initial_node_count | The initial number of nodes for the pool. In regional or multi-zonal clusters, this is the number of nodes per zone. Changing this will force recreation of the resource. Defaults to the value of min_count | " " | Optional |
 | insecure_kubelet_readonly_port_enabled | (boolean) Whether or not to enable the insecure Kubelet readonly port. | null | Optional |
+| single_process_oom_kill | (boolean) On cgroupv2 nodes, defines whether processes in the container are OOM killed individually (true) or as a group (false, the kubelet default). Leave null to keep the kubelet default. | null | Optional |
 | key | The key required for the taint | | Required |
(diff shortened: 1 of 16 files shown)


robert-ziegon:
@apeabody can I get a review?

apeabody (approved):
Thanks for the contribution @robert-ziegon!

robert-ziegon:
@apeabody what is the problem with the failed check? I cant see logs for this one. Anything I can do?

apeabody:
> @apeabody what is the problem with the failed check? I cant see logs for this one. Anything I can do?

Hi @robert-ziegon - There is currently a CI issue, pending https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/2633

robert-ziegon:
@apeabody I think the CI fix is merged now? Can you do another /gcbrun ?

PR #2644, opened 2026-09-03
fix: allow explicitly disabling service external IPs

mirobertod:
## Summary

`service_external_ips = false` currently omits the `service_external_ips_config` block, so existing clusters cannot explicitly disable Service external IPs.

This change makes the input nullable:

- `null` leaves the setting unmanaged, preserving existing default behavior.
- `true` and `false` explicitly manage the GKE setting.

Generated module variants, documentation, and blueprint metadata are updated.

## Validation

- `make docker_generate_modules`
- `terraform fmt -check -recursive`
- GitHub Actions `lint` and `commitlint`


mirobertod:
Hello, any update on this? Thanks

Issue #2646, opened 2026-09-07
Not documented breaking change in v45 force recreation of all nodepools


pvillaverde:
### TL;DR

Upgrading to v45 of the `private-cluster-update-variant` adds another breaking change, recreating ALL NODEPOOLS due to the change in the random suffix resource to add the `pod_cidr_overprovision_disabled`
 (https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2594) 

<img width="592" height="264" alt="Image" src="https://github.com/user-attachments/assets/885b4ceb-f20d-4dfb-a851-0b8520a35d41" />

### Expected behavior

Ideally, this should not cause this breaking change and force the recreation of all nodepools with the downtimes that implies. Alternatively, add the warning in the breaking changes notes of the release and provide alternatives to avoid the recreation of the nodepools.

### Observed behavior

When running `terraform plan`, all nodepools are recreated due to the change of the node pool name caused by the recreation of the random id used as suffix.

### Terraform Configuration

```hcl
Any pre-existing configuration with nodepools should have the same behaviour.
```

### Terraform Version

```sh
1.15.8
```

### Terraform Provider Versions

```sh
google = {
      source  = "hashicorp/google"
      version = "~> 7.46"
    }
```

### Additional information

Editing the tfstates and adding to each random_id resource's instance keepers the key:
`"pod_cidr_overprovision_disabled": "",` avoids the recreation of the resource and thus the nodepools.

agarov:
I have the same problem as @pvillaverde 

I think we should have kept the same system as the other keepers.

You can see that now we have this : 
```
# This keepers list is based on the terraform google provider schemaNodeConfig
# resources where "ForceNew" is "true". schemaNodeConfig can be found in resource_container_node_pool.go at
# https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/resource_container_node_pool.go and node_config.go at
# https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/node_config.go
resource "random_id" "name" {
  for_each    = merge(local.node_pools, local.windows_node_pools)
  byte_length = 2
  prefix      = "${each.key}-"
  keepers = merge(
    zipmap(
      local.force_node_pool_recreation_resources,
      [for keeper in local.force_node_pool_recreation_resources : lookup(each.value, keeper, "")]
    ),
    {
      taints = join(",",
        sort(
          flatten(
            concat(
[...]

PR #2653, opened 2026-09-11
fix: Preserve pool names when ForceNew inputs are unset

agarov:
Build the update-variant `random_id` keeper map from ForceNew node-pool attributes that are explicitly configured.

This prevents an optional ForceNew field introduced in a module release from adding an empty keeper to every existing pool and unnecessarily changing generated node-pool names on upgrade, thereby preventing a complete node pool replacement.

Also add pod_cidr_overprovision_disabled to locals.force_node_pool_recreation_resources to make it consistent with the others and remove its specific line on random_id.

Resolving https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2646, or rather for the forthcoming additions to the list, because the damage has already been done here.

## Compatibility

There will be a one-time "false" breaking change for existing update-variant pools: their prior state contains empty-string keepers for unset ForceNew inputs, while this version omits them. Upgrading to this change will therefore replace those node pools once even when their effective configuration has not changed.

After that migration, future optional ForceNew attributes can be added without causing replacements for users who do not configure them. The ForceNew attribute list remains authoritative: configuring, changing, or removing any listed attribute still changes the keeper map and creates a replacement pool through the update variant.

### DIFF
--- a/autogen/main/cluster.tf.tmpl
+++ b/autogen/main/cluster.tf.tmpl
@@ -967,6 +967,7 @@ locals {
     "flex_start",
     "local_ssd_ephemeral_storage_count",
     "ephemeral_storage_local_ssd_data_cache_count",
+    "pod_cidr_overprovision_disabled",
   ]
 }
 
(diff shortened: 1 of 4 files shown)


agarov:
Hello @apeabody,
Do I need to fix the CI failure? Unfortunately, I don't seem to be able to access the logs

